Issue #25, opened 2015-11-19
Reseeding doesn't fail if no certificates loaded


majestrate:
I need to verify this but after nuking ~/.kovri i noticed reseed still worked even though no certificates were loaded.

Issue #22, opened 2015-11-18
Bump I2P_VERSION to 0.9.22


anonimal:
Needs review before bumping. Check for any I2NP/SSU/NTCP implementation discrepancies between Kovri and i2pd-bb and check if anything *needs* to be implemented if it has not already been done.

Note: I2NP/SSU/NTCP specs have not changed since 0.9.20, so this is all about implementation.

Issue #17, opened 2015-11-17
Enable option to skip SSU extended options


anonimal:
Related to #16. We need to implement ability to *skip* the extended option in case a router doesn't understand extended options. This was recently implemented in i2pd-bb.

PR #26, opened 2015-11-19
Fix reseed, start refactoring ssu, bump router version to .22

majestrate:
* added initial ssu extended options / rekey awareness
* when reseed fails, complain about it and die.
* bump RI version to 0.9.22


majestrate:
#25 #22 #17 should be fixed in this PR

majestrate:
Fixes #17 
Fixes #22 
Fixes #25

Issue #69, opened 2016-01-13
Signup for CodeDocs.xyz


anonimal:
We currently have a Doxygen build option but this would be a nice feature for easy online reference.

From the [site](https://codedocs.xyz/):
> CodeDocs will generate and publish documentation for you, using Doxygen. It's free and easy to setup with your public GitHub repositories.

Monero would simply visit the site and signup via GitHub oauth. As for post-signup configuration, I have no idea what would be needed - so we'll see what needs to be done when we get there.

paulnovo:
Hello, I created codedocs and noticed this issue. @anonimal is correct, you just need to sign on to the service and enable the kovri repository. I created a pull request #78 that adds custom doxygen settings that closely match your Doxyfile.in. You can see the results on my fork:

  https://codedocs.xyz/paulnovo/kovri/

The pull request also adds a badge to your README.md that points to:

  https://codedocs.xyz/monero-project/kovri/

This currently won't work until you enable the repository in CodeDocs. Hope this helps, and let me know what you think.

fluffypony:
Ok done - it says "Generating documentation", so I guess now we wait.

anonimal:
JFTR: the conversation branched off to the [PR](https://github.com/monero-project/kovri/pull/78).

PR #78, opened 2016-01-20
Add .codedocs file and CodeDocs badge to README.md

paulnovo:
Use the CodeDocs.xyz GitHub integration to automatically build the
doxygen documentation.

Fixes monero-project/kovri#69

### DIFF
--- a/.codedocs
+++ b/.codedocs
@@ -1,0 +1,46 @@
+# CodeDocs.xyz Configuration File
+
+# Optional project name, if left empty the GitHub repository name will be used.
+PROJECT_NAME = "The Kovri I2P Router Project"
+
+# One or more directories and files that contain example code to be included.
+EXAMPLE_PATH =
+
+# One or more directories and files to exclude from documentation generation.
+# Use relative paths with respect to the repository root directory.
+EXCLUDE = pkg build
+
+# One or more wildcard patterns to exclude files and directories from document
+# generation.
+EXCLUDE_PATTERNS =
+
+# One or more symbols to exclude from document generation. Symbols can be
+# namespaces, classes, or functions.
+EXCLUDE_SYMBOLS =
+
+# Override the default parser (language) used for each file extension.
+EXTENSION_MAPPING =
+
+# Set the wildcard patterns used to filter out the source-files.
+# If left blank the default is:
+# *.c, *.cc, *.cxx, *.cpp, *.c++, *.java, *.ii, *.ixx, *.ipp, *.i++, *.inl,
+# *.idl, *.ddl, *.odl, *.h, *.hh, *.hxx, *.hpp, *.h++, *.cs, *.d, *.php,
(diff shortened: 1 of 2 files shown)


anonimal:
Hi Paul. So, if a .codedocs file is nothing more than a Doxfile.in (as it appears to be), then we need to use our current Doxyfile.in because but what I'm looking at only provides the barest of minimums. I'll merge but then I'll need to push the appropriate configuration.

Thanks for the PR.

paulnovo:
@anonimal The .codedocs config only uses a subset of the doxygen configuration.  Partially just so I can keep things simpler, and partially so I can add them as needed to make sure they work. There currently allowed options are described in the example [codedocs file](https://codedocs.xyz/static/codedocs_example.txt). Let me know which ones you would like added, and I'll take a look.

One note, I see you copied your current doxyfile.in into .codedocs. This is a cmake input file that is used to generate the doxyfile. Ie, there are a couple of variables (ie @DOXYFILE_OUTPUT_DIR@) that cmake will fill in. And doxygen doesn't know what to do with. Although OUTUT_DIRECTORY isn't used by codedocs anyways. So probably should work fine. Just an FYI.

anonimal:
I'm well aware of the CMake variables. It's easier to ```$ cp build/Doxyfile.in ./.codedocs``` for maintainability and to not have two configurations when one will suffice. The link you provided simply points to Doxyfile options but it is somewhat confusing:
```
# Currently, the following Doxygen configuration options are available. Refer
# to http://doxygen.org/manual/config.html for detailed explanation of the
# options. To request support for more options, contact support@codedocs.xyz.
```
So does codedocs only support the options in the example file or the options that Doxygen supports by default? Considering that everything appears to be Doxygen, a ```$ cp``` shouldn't be a problem.

IMHO, if codedocs doesn't support something, it should be smart enough to drop the line and move on but if it can't do that, it should at least report an error so we can adjust.

I'm curious, do you run codedocs.xyz?

paulnovo:
CodeDocs only supports options in the example file, I'll try and clarify the example to make it more clear.

Currently, yes, if an option is not supported it is just ignored.

Yes, I do run codedocs. Sorry, I thought I mentioned that somewhere, I meant to be clear about that. BTW, how did you run across it? I have only posted a couple messages to the doxygen users mailing list to ask for beta testers, is that where you learned about it? I found the issue your created for kovri when I was googling codedocs.xyz, and thought I would help out.

anonimal:
Oh, so that *is* you! :smile: Nice to meet you.

Yes, ironically, I stumbled onto that post when trying to debug a Doxygen issue and I also saw the CodeDocs badge on Doxygen's github so, I thought we'd give it a try.

Re: options, most of what's [here](https://github.com/monero-project/kovri/blob/development/.codedocs#L4-L152) would be of much (online) use to us; in particular the ```INPUT_FILTER``` as it has a nice display for TODOs. I think scrapping everything below L152 (except for any needed html options (not sure what you're doing server-side)) would be safe to do.

I'm flying blind here because I've never actually signed up to the site (that's Monero's side) so I appreciate your help and feedback.

PR #121, opened 2016-02-07
Reseed bug fix + Lib reorganization/CMake updates.

anonimal:


anonimal:
```
2016-02-07 18:58:34     &anonimal       I see what you mean. Since we never http reseed, I didn't consider that. 
2016-02-07 18:59:36     +EinMByte       anonimal: No, but we use http for hosts.txt
2016-02-07 18:59:55     &anonimal       But that b32 is always 80, yes?
2016-02-07 19:00:16     +EinMByte       Sure, but maybe next week someone decides it isn't
2016-02-07 19:00:21     +EinMByte       and then the house collapses
2016-02-07 19:00:21     &anonimal       Yep.
2016-02-07 19:00:44     &anonimal       This is why I tend to go slow and prefer rewrites to patches.
2016-02-07 19:00:51     &anonimal       I'll make the change.
```

anonimal:
JFTR: @EinMByte and I agree that our URI parser is way under par so we are researching cpp-netlib as noted [here](https://github.com/PurpleI2P/i2pd/pull/283) and [here](https://github.com/monero-project/kovri/blob/master/src/core/util/HTTP.h#L69-L80). If not cpp-netlib, then we should consider something safer/smarter than what we have now - even if that means a rewrite.

Issue #65, opened 2016-01-10
Find at least "5" (there are more) raw memory leaks due to missing or mis-handled implementation


anonimal:
Easter-egg hunt in January?
```
21:15:06       zzz | for your todo list: <orignal> there are approcimately 5 more places sending raw memory
21:48:55     str4d | <orignal> they are all marked with TODO: <-- raw memory locations (I assume he means padding)
```

Note: focus on any ```TODO```'s, especially those marked with ```???```

#### Running tally (If there are more, they will be resolved by #58):
9dd3dc03f7f013629a7146da8e6b55a6884281ef
1dcbe27bde2f20757ffb10a5f52d24897dafb8c3
#123 
#125 
#165 

Note: regarding the time-frame to complete d649e8a in #165:

1. I was waiting for #140 to be resolved but, in our meetings, we eventually agreed to not wait any longer and pursue a patch instead
2. I wanted to focus on finishing #149 before flipping branches and committing d649e8a. This could have easily gone the other way but, since we are not in production, and the leak is actually not raw (it *is* initialized, just not with random padding as java i2p does it), I chose to focus on the bigger work involved in #149 and, once completed, resolve this ticket.

KanediasTheMaker:
What prevents from using `valgrind`?

anonimal:
Nothing. Running valgrind at this stage is comical. Please, see for yourself.

This ticket is an extension of a recent IRC discussion and is related to raw memory contents leaking into message space because of improper implementation and not necessarily improperly deallocated "memory leaks" in the traditional sense (though, that is also an agenda item (see #58)).

See https://github.com/monero-project/kovri/commit/9dd3dc03f7f013629a7146da8e6b55a6884281ef as an example.

If you'd like to learn more, visit us on IRC (see [CONTRIBUTING.md](https://github.com/monero-project/kovri/blob/master/CONTRIBUTING.md))

KanediasTheMaker:
>  not necessarily improperly deallocated "memory leaks" in the traditional sense

Ahh, got it

> If you'd like to learn more, visit us on IRC

Sorry, I'm currently involved in other FLOSS project, but I'll keep an eye on your fork that seems to be far less toxic than orig[i]nal one.

For now the only thing I can help with is armv7h GNU/Linux compile and day-to-day use

anonimal:
[...]

PR #125, opened 2016-02-10
I2NP: fix tunnel build request raw memory leak.

anonimal:
* Added spec documentation
* Referencing #65 

Note: peer review would be appreciated before we merge this into master (@EinMByte).


EinMByte:
Late review: I think this should be fine. Good work on the documentation too.

PR #132, opened 2016-02-14
Crypto: create true RandInRange(). Fixes #130.

### LINKED ISSUE #130: Crypto: RandInRange() fails to stay within range when using signed or unsigned integers
For example:
```cpp
int port = i2p::crypto::RandInRange<int>(9111, 30777);
```
When looped, will produce the following:
```
4835
12969
12302
-7772
-9926
-10952
6444
20583
29331
609
22397
...etc.
```
For the time being, using any unsigned int appears to be fine. We currently do not use any signed int's for ```RandInRange()``` but this issue is quite dangerous because this caveat is undocumented/(code is unfinished?) so, should we ever need signed int's with ```RandInRange()``` in critical code...

anonimal:


EinMByte:
Mersenne twister is not a CSRNG. Do *not* merge this.

anonimal:
In its *native form* yes, but do you see where we implement ```RandInRange()```? Our current implementation is not CSRNG but at least this fix keeps things within bounds. @EinMByte How long do you propose we wait for a new engine before merging?

anonimal:
**Edit: s/where/how/**. And I just now saw your [comment](https://github.com/monero-project/kovri/issues/130#issuecomment-183880837) in #130 and have [responded](https://github.com/monero-project/kovri/issues/130#issuecomment-183977642).

(I'm still not getting emails :frowning: are you also having i2pmail problems?)

anonimal:
JFTR: dialogue in #130 has given us a successful resolution. My thanks to @EinMByte.

PR #174, opened 2016-05-08
Misc. enhancements/fixes, update build requirements

anonimal:

- [X] I confirm.

---

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,11 @@ dist: trusty
 language: cpp
 compiler:
 - gcc
-# Uncomment when we can actually build with clang
-#- clang
 addons:
   apt:
     packages:
+    - gcc-5
+    - g++-5
     - libboost-all-dev
     - libcrypto++-dev
     - libssl-dev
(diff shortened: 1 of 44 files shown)


anonimal:
Note: there are now *two* issues with the build

1) Apparently Travis CI's Ubuntu LTS does not allow installing gcc-5/g++-5 even though both packages should be available as I have successfully built this PR on my own, identical, LTS machine. This could be a problem on their end and I will look into the issue soon (late Sunday/this week) before this PR is merged.

2) The OSX and FreeBSD build was broken *before* this PR. I will open new tickets to resolve OSX/BSD linkage issues after this Travis CI issues is fixed and this PR is merged (note: compilation is successful on those respective systems but linking is the only issue).

anonimal:
Referencing #175 and #176.

In [today's meeting](https://getmonero.org/2016/05/08/logs-for-the-kovri-dev-meeting-held-on-2016-05-08.html) we agreed to stash the bsd/osx gcc/clang build and instead focus on linux/windows(cygwin) gcc-only for now.  If our only option is to bring back clang for osx/bsd, then we can do so though, because of limited resources, the build may not be consistently successful (as opposed to linux/windows(cygwin)).

Issue #191, opened 2016-06-05
Valgrind profiling


anonimal:
**By submitting this issue, I confirm the following:**
- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

_Place an X inside the bracket to confirm_
- [X] I confirm.

---

A snippet of [valgrind report](https://paste.fedoraproject.org/374871/14651435/raw/) (built from 2270ee031efd8fc6f7b87093e59434263ac51ad6)

The `Invalid read of size 1` leak is related to a `CryptoPP` bug that has been fixed upstream but, since it won't be distro-ready until their next release, we [had to patch](https://github.com/monero-project/kovri/issues/141) - but that is not the real problem.  The `NTCP/I2NP/Tunnel` related leaks will leak indefinitely - and that is the real problem.

anonimal:
JFTR: I've left important notes for reference in #218 with one being:

> We should look closely at GetFormattedSessionInfo() and possibly refactor out its implementation because commenting out its functionality also produces good results - but that could be coincidental.

anonimal:
New valgrind report against 4a7c6a39a63dbf736e8b83b24bef54d0f725fbe8 for about 15 minutes, killed with `SIGINT`:
```bash
$ valgrind --leak-check=yes --error-limit=no --show-reachable=yes --log-file=$HOME/valgrind_kovri_$(date +%Y.%m.%d) build/kovri --log-to-console 0
```
[kovri_valgrind_2016.09.09.txt](https://github.com/monero-project/kovri/files/464468/kovri_valgrind_2016.09.09.txt)

anonimal:
New report attached, built against 00e0dfe6ce1b458cf28bd453b858b1f57fbe4200
- Uptime: 30+ minutes
- No memory leaks detected until *after* `SIGINT` (referencing #342).
- Ran several shorter duration test-runs before this report; similar results were produced

Command and report:
```bash
$ valgrind --leak-check=yes --error-limit=no --show-reachable=yes --log-file=$HOME/valgrind_kovri_$(date +%Y.%m.%d) build/kovri --log-to-console 0
```
[...]

PR #218, opened 2016-07-07
LogPrint/All transports: alleviate massive memory leak.

anonimal:

- [X] I confirm.

---

* References #191

### DIFF
--- a/src/core/transport/NTCP.cpp
+++ b/src/core/transport/NTCP.cpp
@@ -54,13 +54,9 @@ NTCPServer::NTCPServer(
       m_NTCPEndpoint(boost::asio::ip::tcp::v4(), port),
       m_NTCPEndpointV6(boost::asio::ip::tcp::v6(), port),
       m_NTCPAcceptor(nullptr),
-      m_NTCPV6Acceptor(nullptr) {
-  LogPrint(eLogDebug, "NTCPServer: initializing");
-}
+      m_NTCPV6Acceptor(nullptr) {}
 
-NTCPServer::~NTCPServer() {
-  LogPrint(eLogDebug, "NTCPServer: destroying");
-}
+NTCPServer::~NTCPServer() {}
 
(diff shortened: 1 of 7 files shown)


anonimal:
This appeared to have fixed the leak after many - **many** - tests against fresh instances but apparently, after yet another test, it has not been entirely fixed. At the very least, this fix has reduced the consistency of recurring leaking instances - but it is not a final solution.

This PR should stay open until this **very annoying leak** is resolved once-and-for-all.

anonimal:
Pushed two new commits. 3902502 seems to completely resolve the massive leak.

I have **repeatedly** tested with fresh ```KOVRI_DATA_DIR```'s, stale ```KOVRI_DATA_DIR```'s, multiple reseeds, different ports, and have been able to successfully complete a client tunnel and connect to IRC on nearly every instance: all without the massive leak.

I'll certainly keep an eye out for a performance regressions but I think this PR should remain open for at least another day or two or until someone else can confirm that these solutions are viable with regular testing.

anonimal:
The good news: this has tremendously reduced the occurrence and time duration of leak. The bad news: the bastards still live and we should remove LogPrint from all ctors/dtors and then see where we stand.

Also, a review of tunnels are in order as the problem seems to grow *after* a client tunnel has been in use.

Note: valgrind is to slow to work with for these leaks despite what's been pasted in #191.

anonimal:
This is still very hard to reproduce and I can successfully run a working router for hours without issue. Also, I cannot find any ```LogPrint()``` in critical ctor/dtor areas that weren't already there from the beginning.

We should look closely at ```GetFormattedSessionInfo()``` and possibly refactor out its implementation because commenting out its functionality also produces good results - but that could be coincidental.

I'll merge this and keep the discussion going in #191.

anonimal:
Note: moneromooo had also confirmed that there was no massive memory leak.

Issue #210, opened 2016-06-19
Reseed fails on Debian Sid (unstable)


anonimal:
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

*Place an X inside the bracket to confirm*
- [X] I confirm.

---

First reported by @EinMByte, this appears to be specific to Debian Jessie (all tests pass and Reseed is fully functional on Arch and Ubuntu Wily/Trusty).

Output (omitting passing tests):
> kovri/src/tests/unit_tests/core/Reseed.cpp(254): error in "GoodSU3": check su3.SU3Impl() failed
*** 1 failure detected in test suite "Master Test Suite"

Build info:
> -- The C compiler identification is GNU 4.9.2
-- The CXX compiler identification is GNU 4.9.2
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Performing Test CXX14_SUPPORTED
-- Performing Test CXX14_SUPPORTED - Success
-- Looking for include file pthread.h
-- Looking for include file pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Boost version: 1.55.0
-- Found the following Boost libraries:
--   chrono
--   log
--   program_options
--   date_time
--   thread
--   system
--   filesystem
--   regex
-- Found Crypto++: /usr/include, /usr/lib/libcryptopp.so
-- Found OpenSSL: /usr/lib/x86_64-linux-gnu/libssl.so;/usr/lib/x86_64-linux-gnu/libcrypto.so (found version "1.0.1t") 
[...]

PR #220, opened 2016-07-07
Reseed: do not rely on [a-z] catching all letters

moneromooo-monero:

- [ ] I confirm.

---

*Enter your pull request summary here*

This will fail in some locales, the list of which I am totally
unable to list. At least one Turkish one, I believe.

Fixes tests and main binary on Fedora.


anonimal:
Thanks again @moneromooo-monero. JFTR: we made the decision long ago to move away from macros when possible - so that portion will have to be refactored but, regardless, this was a great catch!

Referencing #210.

Issue #176, opened 2016-05-08
Support FreeBSD 10


anonimal:
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

*Place an X inside the bracket to confirm*
- [X] I confirm.

---

FreeBSD ships boost 1.55 and we need 1.58 minimum. How is [1.60](https://wiki.freebsd.org/BoostPortingProject/1.55-to-1.60) progressing? For now, we can instruct to build boost from source but we'll need machine time to confirm a successful build.

Issue #175, opened 2016-05-08
OSX: build fails on linkage


anonimal:
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

*Place an X inside the bracket to confirm*
- [X] I confirm.

---

**Edit**: original post no longer applies when building with clang against 368bfdc.

All that remains for this ticket is https://github.com/weidai11/cryptopp/issues/232.

anonimal:
```
anonimal | bigreddmachine: FYI regarding the link, I added -stdlib=libc++ back in December when I was more active with clang and it did fix the linkage issue.                                                            │         
anonimal | Now that I've dropped clang support for bsd/osx, that option does not exist in gcc - so we can't use it.                                                                                                       │         
anonimal | This does bring up the question of *if* we must now build respective libraries (boost/crypto++/more?) using g++-5/g++-6 and not clang.                                                                         │         
anonimal | I really hope that's not the case.
```

rakhimov:
I doubt you can build with GCC 5 on OS X due to binary incompatibility of all dependencies,
which are likely to have been compiled with clang-llvm.

rakhimov:
I got a working build on Travis with clang: https://travis-ci.org/rakhimov/kovri/builds/145718362

rakhimov:
The binary incompatibility goes the other way around as well,
so with GCC 5+ on Ubuntu as default (Wily, Xenial), 
clang builds will fail to link.

anonimal:
@rakhimov [proved that, on Travis-CI,](https://travis-ci.org/rakhimov/kovri/builds/145718362) we can still build with clang on OSX but, in our in-house OSX instance, though passing ```c++1y``` fixes the cmake issue, make will still fail for various reasons.

@monero-project @fluffypony or myself needs to review our in-house instance for discrepancies. Once that is resolved (or if we shell-out fiat for the OSX Travis build), I'd like to update the building documentation to reflect OSX clang support.

PR #237, opened 2016-07-15
Fix compilation with Clang 3.8

rakhimov:

- I am not basing this pull-request on branch ```master```
- I am sending this pull-request to branch ```development```

- [x] I confirm.

---

Verified compilation with Clang 3.8 on Ubuntu 16.04;
however, no linking due to binary incompatibility of libstdc++.
The compilation needs further testing on Mac or FreeBSD.

---
This change is [Reviewable](https://reviewable.io/reviews/monero-project/kovri/237)

### DIFF
--- a/src/client/I2PControl/I2PControl.h
+++ b/src/client/I2PControl/I2PControl.h
@@ -321,6 +321,7 @@ class I2PControlSession
                                         m_RouterManagerHandlers,
                                         m_NetworkSettingHandlers;
 
+  /// @todo Unused private field. Why?
   boost::asio::io_service& m_Service;
   boost::asio::deadline_timer m_ShutdownTimer,
                               m_ExpireTokensTimer;
--- a/src/client/I2PService.h
+++ b/src/client/I2PService.h
@@ -95,7 +95,7 @@ class I2PService {
   virtual void Stop() = 0;
 
   // everyone must override this
-  virtual std::string GetName() = 0;
+  virtual std::string GetName() const = 0;
 
  private:
   std::shared_ptr<ClientDestination> m_LocalDestination;
@@ -199,9 +199,7 @@ class TCPIPAcceptor : public I2PService {
   virtual std::shared_ptr<I2PServiceHandler> CreateHandler(
       std::shared_ptr<boost::asio::ip::tcp::socket> socket) = 0;
 
-  virtual std::string GetName() {
-    return "generic TCP/IP accepting daemon";
-  }
+  std::string GetName() const { return "generic TCP/IP accepting daemon"; }
 
  protected:
   std::string m_Address;
--- a/src/client/I2PControl/I2PControl.cpp
+++ b/src/client/I2PControl/I2PControl.cpp
@@ -317,7 +317,7 @@ I2PControlSession::Response I2PControlSession::HandleRequest(
     }
     // Call the appropriate handler
     (this->*(it->second))(params, response);
-  } catch (const boost::property_tree::ptree_error& error) {
+  } catch (const boost::property_tree::ptree_error&) {
     response.SetError(ErrorCode::e_ParseError);
   } catch (...) {
     response.SetError(ErrorCode::e_InternalError);
@@ -340,7 +340,7 @@ bool I2PControlSession::Authenticate(
       response.SetError(ErrorCode::e_ExpiredToken);
       return false;
     }
-  } catch (const boost::property_tree::ptree_error& error) {
+  } catch (const boost::property_tree::ptree_error&) {
     response.SetError(ErrorCode::e_NoToken);
     return false;
   }
--- a/src/client/I2PTunnel/I2PTunnel.cpp
+++ b/src/client/I2PTunnel/I2PTunnel.cpp
@@ -384,9 +384,7 @@ std::unique_ptr<const i2p::data::IdentHash> I2PClientTunnel::GetIdentHash() {
   return std::move(m_DestinationIdentHash);
 }
 
-std::string I2PClientTunnel::GetName() {
-  return m_TunnelName;
-}
+std::string I2PClientTunnel::GetName() const { return m_TunnelName; }
 
 std::shared_ptr<I2PServiceHandler> I2PClientTunnel::CreateHandler(
     std::shared_ptr<boost::asio::ip::tcp::socket> socket) {
@@ -582,9 +580,7 @@ void I2PServerTunnel::CreateI2PConnection(
   conn->Connect();
 }
 
-std::string I2PServerTunnel::GetName() {
-  return m_TunnelName;
-}
+std::string I2PServerTunnel::GetName() const { return m_TunnelName; }
 
 I2PServerTunnelHTTP::I2PServerTunnelHTTP(
     const std::string& name,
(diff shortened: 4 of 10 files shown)


anonimal:
Hi again @rakhimov.

I'm glad to see this PR and I hope to see many more!

A few notes before merging:

1. Though this PR is specific to 3.8 on Ubuntu 16.04, I should be noted for the record that the clang build still fails on other default installs (though Arch builds successfully):
  - Debian stable (clang 3.5): ```I2PControlServer.cpp:139:3: error: no matching function for call to 'buffer' (Boost.Asio buffer)```
  - Ubuntu 14.04 (clang 3.6): ```I2PControlServer.cpp:139:3: error: no matching function for call to 'buffer' (Boost.Asio buffer)```
  - Ubuntu Wily (clang 3.6): libstd++ linkage error
  - OS X 10.9.5 (Apple LLVM 6.0 (based off of 3.5)): ```C++14 not supported```
  - FreeBSD 10 (clang 3.8): 215 warnings, 1 error: ```kovri/src/core/Reseed.cpp:270:14: error: no matching member function for call to 'Seekg'
    m_Stream.Seekg(Offset::version, std::ios::cur);```
2. We write all todo's as ```TODO(unassigned):``` (replacing unassigned with your name if you'd like) so Doxygen can catch them as described in the [TODO's portion of the contributing guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md#todos).
3. Long ago, it was decided to keep the style more vertical than horizontal as it slows down the eye and tends to prevent easy mistakes; that means we've avoided one-liners when possible (e.g., 'write to be read'). This note is mostly an FYI in case this portion of works is refactored in the future (then you'll know why).

I'm also referencing #175  and #176. In all, the OSX and BSD build needs much love - and so does clang. Clang is noted as unsupported and our cmake [will fail in an attempt to compile with clang](https://github.com/monero-project/kovri/blob/master/CMakeLists.txt#L59).

If we can get the clang build completely up-and-running for these systems, we should definitely bring back clang support. If you chose further work in this area, your efforts will be greatly appreciated.

src/client/I2PControl/I2PControl.h
```diff
@@ -321,6 +321,7 @@ class I2PControlSession
                                         m_RouterManagerHandlers,
                                         m_NetworkSettingHandlers;
 
+  /// @todo Unused private field. Why?
```
anonimal: Good catch. Also, I've commented [here](https://github.com/monero-project/kovri/pull/237#issuecomment-233101841) about TODO's

src/client/I2PService.h
```diff
@@ -199,9 +199,7 @@ class TCPIPAcceptor : public I2PService {
   virtual std::shared_ptr<I2PServiceHandler> CreateHandler(
       std::shared_ptr<boost::asio::ip::tcp::socket> socket) = 0;
 
-  virtual std::string GetName() {
-    return "generic TCP/IP accepting daemon";
-  }
+  std::string GetName() const { return "generic TCP/IP accepting daemon"; }
```
anonimal: Comments about one-liners [here](https://github.com/monero-project/kovri/pull/237#issuecomment-233101841).

EinMByte: Just want to note that I agree that we should refrain from one-liners (and inline implementation overall).

rakhimov: Just a side not. I was wondering here about the temporary strings.
Why not simple ```const char*```?

EinMByte: @rakhimov [https://en.wikipedia.org/wiki/Return_value_optimization](There are no temporary strings.)

rakhimov: Probably this code is not hot to consider the optimization,
so it may not matter.
Had it been hot,
temporary strings would be there regardless of the RVO.
RVO with copy elision avoids one more temporary,
but the main one remains regardless.

Consider the user code:
```cpp
std::string name = object->GetName();  // Bad. Temporary string. Only use if ``name`` to be modified later (unlikely).
const std::string& name = object->GetName();  // Bad. The same temporary to const ref binding.
```
The temporary string is unavoidable for this API.
Traditionally, the return value would be expected to be ``const std::string&``.
With ``const char*``, it may avoid string construction at all
because the compile time literal name seems to the common case.

anonimal:
> Other than that, several other reasons can be given if you're interested.

I would also add that if we need a C idiom for design work then we're *probably* doing something wrong. By 'wrong', I mean that it's *possible* that we're not seeing the bigger picture design-wise; regardless of whether the variable is temporary or not. 

STL is grand and flexible. Its overhead ranges from minuscule to none. Maintainability and flexibility are some of the benefits of our chosen language and I think that we should exploit them when possible.

Also, personally, regardless of the known arguments, I never want to see a raw pointer in C++ again - but that's just me (and is **not** a mandate of this project).

> Consider the user code:
```std::string name = object->GetName();  // Bad. Temporary string. Only use if ``name`` to be modified later (unlikely).```
```const std::string& name = object->GetName();  // Bad. The same temporary to const ref binding.```

I would say to use ```auto``` and move on. IMHO, ```auto``` allows for better OOP and easier maintenance. ```GetName()``` should do whatever it needs to do but only on its own terms and, when possible, it should not drag down the rest of the code (IMHO).

We are moving toward C++11/14 compliance so I hope that we can enjoy these perks given during the process.

src/client/I2PService.h
EinMByte: If the string has a name, it's not longer temporary (it's an lvalue).
Returning `const std::string&` used to be better, although its importance has somewhat decreased because of move constructors.

I do agree that there is a performance cost of creating a `std::string` object, though. But we should accept this cost, for many good reasons. Firstly, because the cost is likely to be minimal. Bound checking on a `const char*` is terrible and it has been the cause of many security issues. Also, raw pointers in general complicate memory management (although in this case, that's not applicable). Optimizations are premature and even unnecessary, and yes, they are even a bit evil.
Optimization of C++ code at such a low level is far too complex for ordinary human being such as us. We should improve our logic and algorithms instead, because that's where the real improvements can be made.

That said, I do appreciate that you are looking into all of this old crappy code. But at this stage of the project it would be more interesting if you could search for bugs and optimizations at the logical level. In fact, that'd be awesome for us.

anonimal: >old crappy code

Couldn't have said it better :+1:

rakhimov:
Don't take my PRs too seriously; I am just passing by to find something that I don't already know about C++.
I didn't know that map::erase was safe in a loop. I learned something!
Thanks!

anonimal:
That's good, I'm glad to hear. Tourism is nice but has its drawbacks. You're welcome to take a bigger bite out of clang support here - and I'm sure we would all learn something new along the way.

rakhimov:
Btw, Coverity seems to be not so happy with that map::erase, list::erase inside of loops with iterators
that I thought were perfectly safe from the Kovri code.
For some reason, it is reported as an API misuse. It is very strange, but seems to be a false-positive.

anonimal:
@rakhimov yes, Coverity is still very prone to false-positives; so much as they even have a *Modeling File* section in project settings.

rakhimov:
I usually take Coverity false-positives seriously as a sign that the code is convoluted for a human as well as the static analysis tool. It serves as a good hint to refactor the code, sometimes even the design.

Issue #325, opened 2016-08-28
Do NOT abuse C++11 ``auto``


rakhimov:
**By submitting this issue, I confirm the following:**
- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

_Place an X inside the bracket to confirm_
- [x] I confirm.
---

I am noticing several places in Kovri database
where auto is abused and makes the code hard to read.
Please follow the reasonable suggestions from the GCSG:
https://google.github.io/styleguide/cppguide.html#auto

Some suggestions to improve the readability with auto:
- Do not use `auto` for non-local var types.
- Don't make readers do and remember template type deduction in their heads.
- Do not omit `*` and `&` from the auto type of pointers and references.

EinMByte:
Instead, we should use `auto` whenever convenient.

rakhimov:
Abuse of ``auto`` obfuscates the code,
and make you use your precious 7 registers of your brain.

rakhimov:
Reading through the code I have no idea what those abused ``auto`` mean.
I have to stop and think what the function return types are.
Find a function and remember -> move on.
The code should be written for readers.

rakhimov:
It is best to use ``auto`` when:
1. the type is super-obvious to a reader (repetitions)
2. you-really-shouldn't-care about the type (e.g. iterators)
3. it is impossible to find the type without compiler help (templates, return type deduction).

anonimal:
Or simply look at the header: that's what they're there for (no memorization required). Or use a decent IDE (e.g., netbeans).

>The code should be written for readers.

Exactly.

>```auto``` obfuscates the code

I think it abstracts more than obfuscates.

>Do not use auto for file-scope or namespace-scope variables, or for class members

For this code-base, I disagree with their decision.

I think that once you spend a little time with this base, you'll get into the swing of things and see that the changes you proposed may actually be counterproductive in the long run.

rakhimov:
I strongly disagree with the current Kovri practice for ``auto``.
My editor of choice is Vim,
and I learned from practice that if your codebase needs an IDE, that is a sign of a problem.
Jumping back-and-forth between declaration and usage is rather anti-reader.
[...]

Issue #263, opened 2016-07-23
Fix Coverity Defects


anonimal:
**By submitting this issue, I confirm the following:**
- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

_Place an X inside the bracket to confirm_
- [X] I confirm.

---

Coverity has proven itself to be problematic:
1. #209 
2. Less than 2 weeks ago, they' gave us a passing OK of 0 defects - which was false. For no inexplicable reason, they [now report that we have 39 defects](https://scan.coverity.com/projects/monero-project-kovri?tab=overview) and, from what I've read from a small snippet of email they sent me, are almost all in old inherited code (all being the snippet they sent me). This means that the **same code** that produced 0 defects less than two weeks ago is now defect worthy.
3. The website won't allow me to view the defects despite having correct permissions (see 1.)

Of the 39 existing defects, I have a feeling that a fair amount are either upstream (components for metrics need fine-tuning) and/or false-positives. I cannot confirm this yet until @fluffypony can take screenshots of the defects or if Coverity emails me _all_ the defects or if Coverity magically fixes their website.

When I have a better idea of where we are, I'll relabel this issue as needed.

anonimal:
Coverity's website is still not allowing me to view defects. An email has been sent and CC'd to @fluffypony.

anonimal:
They've fixed permissions and I've skimmed over the defects. The components were able to focus only on our codebase.

```
&anonimal | Well, 37 of the 39 are not my code. With the remaining 2, 1 is a questionable false-positive but the other is something I should've caught.                                                            │
&anonimal | Of those remaining 2, neither are in any critical areas.
```

The remaining 37 range across the entire base, from medium to high impact. Most are easily fixed. Some are rooted in Romako Shelkin design flaws and will require larger refactoring work.

Two of the SSU defects [are known](https://github.com/monero-project/kovri/issues/191) and should be addressed in #140. ping @EinMByte.

Regarding time-frame to complete: all of these would go away quicker if I was paid. I've never been paid to do any work on this project so the defects will eventually be chipped away with time. If anyone wants to speed up the process, send me an email and I'll send you an invite to view them.

[...]

PR #326, opened 2016-08-28
Deal with Datagram msg leak false-positive

rakhimov:

- [x] I confirm.

---
This is a Coverity reported leak,
which seems to be false positive.
The current refactoring is to make the code easier on Coverity.

---
This change is [Reviewable](https://reviewable.io/reviews/monero-project/kovri/326)

### DIFF
--- a/src/client/datagram.cc
+++ b/src/client/datagram.cc
@@ -59,43 +59,43 @@ void DatagramDestination::SendDatagramTo(
     uint16_t fromPort,
     uint16_t toPort) {
   uint8_t buf[MAX_DATAGRAM_SIZE];
-  auto identityLen =
-    m_Owner.GetIdentity().ToBuffer(buf, MAX_DATAGRAM_SIZE);
+  auto identityLen = m_Owner.GetIdentity().ToBuffer(buf, MAX_DATAGRAM_SIZE);
   uint8_t* signature = buf + identityLen;
   auto signatureLen = m_Owner.GetIdentity().GetSignatureLen();
   uint8_t* buf1 = signature + signatureLen;
   size_t headerLen = identityLen + signatureLen;
   memcpy(buf1, payload, len);
-  if (m_Owner.GetIdentity().GetSigningKeyType() ==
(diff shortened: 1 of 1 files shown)


src/client/datagram.cc
```diff
@@ -59,43 +59,43 @@ void DatagramDestination::SendDatagramTo(
...
+  I2NPMessage* msg = CreateDataMessage(buf, len + headerLen, fromPort, toPort);
+  std::shared_ptr<const i2p::data::LeaseSet> remote
+      = m_Owner.FindLeaseSet(ident);
+  if (remote) {
     m_Owner.GetService().post(
-        std::bind(
-          &DatagramDestination::SendMsg,
-          this, msg, remote));
-  else
+        [this, msg, remote] { SendMsg(msg, remote); });
```
rakhimov: Is there any possibility that this callback is never executed?
(if it is not executed, then the message will not get deleted.)

src/client/datagram.cc
```diff
@@ -59,43 +59,43 @@ void DatagramDestination::SendDatagramTo(
...
   } else {
     m_Owner.Sign(buf1, len, signature);
   }
-  auto msg =
-    CreateDataMessage(buf, len + headerLen, fromPort, toPort);
-  auto remote = m_Owner.FindLeaseSet(ident);
-  if (remote)
+  I2NPMessage* msg = CreateDataMessage(buf, len + headerLen, fromPort, toPort);
+  std::shared_ptr<const i2p::data::LeaseSet> remote
+      = m_Owner.FindLeaseSet(ident);
```
anonimal: I don't see a need to refactor-out ```auto``` (to do so is counterproductive for abstraction and should be refactored back).

rakhimov: Please see #325

src/client/datagram.cc
anonimal: No, not that I can see. If lease-set is available, great, if not, it's requested below.

anonimal:
Thanks @rakhimov. I'll push a coverity build after this is merged.

anonimal:
Referencing #263.

PR #327, opened 2016-08-28
Refactor I2np protocol message with unique_ptr

rakhimov:

- [x] I confirm.

---

The message is refactored with a unique_ptr instead of a raw pointer that acts like a handle.
Releasable shared_ptr is implemented to make callback functions copy constructible.
This seems to suggest that messages must be constructed with a shared_ptr by default.

If that is the case,
I will rewrite the messages with the shared_ptr instead of unique_ptr.

This approach should solve some false-positives from Coverity.

---
This change is [Reviewable](https://reviewable.io/reviews/monero-project/kovri/327)

### DIFF
--- a/src/client/datagram.cc
+++ b/src/client/datagram.cc
@@ -59,43 +59,43 @@ void DatagramDestination::SendDatagramTo(
     uint16_t fromPort,
     uint16_t toPort) {
   uint8_t buf[MAX_DATAGRAM_SIZE];
-  auto identityLen =
-    m_Owner.GetIdentity().ToBuffer(buf, MAX_DATAGRAM_SIZE);
+  auto identityLen = m_Owner.GetIdentity().ToBuffer(buf, MAX_DATAGRAM_SIZE);
   uint8_t* signature = buf + identityLen;
   auto signatureLen = m_Owner.GetIdentity().GetSignatureLen();
   uint8_t* buf1 = signature + signatureLen;
   size_t headerLen = identityLen + signatureLen;
   memcpy(buf1, payload, len);
-  if (m_Owner.GetIdentity().GetSigningKeyType() ==
-      i2p::data::SIGNING_KEY_TYPE_DSA_SHA1) {
+  if (m_Owner.GetIdentity().GetSigningKeyType()
+      == i2p::data::SIGNING_KEY_TYPE_DSA_SHA1) {
     uint8_t hash[32];
     i2p::crypto::SHA256().CalculateDigest(hash, buf1, len);
     m_Owner.Sign(hash, 32, signature);
   } else {
     m_Owner.Sign(buf1, len, signature);
   }
-  auto msg =
-    CreateDataMessage(buf, len + headerLen, fromPort, toPort);
-  auto remote = m_Owner.FindLeaseSet(ident);
-  if (remote)
+  I2NPMessage* msg = CreateDataMessage(buf, len + headerLen, fromPort, toPort);
+  std::shared_ptr<const i2p::data::LeaseSet> remote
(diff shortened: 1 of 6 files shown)


rakhimov:
This PR depends on #326.
I will rebase this PR if #326 gets merged.

src/client/datagram.cc
```diff
@@ -52,6 +52,41 @@ DatagramDestination::DatagramDestination(
...
+/// @returns The released payload from the shared ptr.
+template <typename T>
+T* release(ReleasableSharedPtr<T>& smart_ptr) {
+  T* paylaod = *smart_ptr;
+  *smart_ptr = nullptr;
+  return paylaod;
+}
+
+}  // namespace
+
```
anonimal: Great template code but maybe I've missed *why* we need to make the callback copy-constructable (thus needing to hack a releasable ```shared_ptr```)?

rakhimov: I tried to keep using ``unique_ptr`` but boost::asio started complaining that it required copy_constructable function. Moving functions didn't work.

I can refactor to make the msg work with shared pointers by default instead of unique_ptr.
it would remove some code.
However, this is a design choice.
Usually shared_ptr in places where unique_ptr will do is not a good sign.

rakhimov: Also, note that this approach with releasable shared_ptr incurs extra performance overhead per message.

anonimal: Looking at ```datagram``` code, I can tell that I've spent almost no time in this area (its hideous), so I can't give any solid opinions at the moment other than:

1. If you're willing to dive deeper, refactor bigger design to accommodate true ```unique_ptr```
2. If diving deeper, you see that ```shared_ptr``` is the only option (I don't see why though (who else *really* needs ownership of the message?)), then feel free to refactor
3. Or, add a huge ```TODO``` to resolve this issue and we will merge this and get back to it at another time

Or maybe some other ideas?

rakhimov:
I am reluctant to make a design choice in this piece of code.
In this PR, unique_ptr gets the job of preventing resource leaks and getting rid of raw pointer.

I am sure future refactorings and bug-fixes will give clearer picture on the issue of shared_ptr vs unique_ptr for messages.
The strange thing I see is that there's a helper function that converts message ptr into shared_ptr,
so it seems like maldesign from the beginning.
I vote for putting this redesign for later.

I can put ``todo`` in the message class code and assign myself.

anonimal:
>so it seems like maldesign from the beginning.

I wouldn't be surprised :smile:

src/client/datagram.cc
```diff
@@ -116,14 +124,13 @@ void DatagramDestination::SendDatagramTo(
...
     });
   } else {
     m_Owner.RequestDestination(
         ident,
-        [this, temp_msg](
-            const std::shared_ptr<i2p::data::LeaseSet>& remote) mutable {
+        [this, temp_msg](const std::shared_ptr<i2p::data::LeaseSet>& remote) {
           HandleLeaseSetRequestComplete(
               remote, std::unique_ptr<I2NPMessage>(release(temp_msg)));
         });
```
anonimal: >I can put todo in the message class code and assign myself.

Did you still plan to put in TODO's?

src/core/i2np_protocol.h
```diff
@@ -299,6 +299,9 @@ struct I2NPMessageBuffer : public I2NPMessage {
   uint8_t m_Buffer[SZ + 16] = {};
 };
 
+// TODO(rakhimov): Consider shared_ptr instead of unique_ptr
```
anonimal: Thanks @rakhimov

Issue #187, opened 2016-05-29
NTCP: multiple core issues


anonimal:
**By submitting this issue, I confirm the following:**
- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

_Place an X inside the bracket to confirm_
- [X] I confirm.

---

Up until very recently, with full knowledge of the quality of this code-base and the many security issues/issues at-large (many of which have been resolved), I had never ran Kovri over clearnet - not once - since the first Kovri commit. That's over 6+ months of non-clearnet development! Instead, I had been using a Tor transproxy to do the myriad of other other critical/major/misc development that was needed - but did not require clearnet access.

Now that I am regularly developing Kovri over clearnet, core issues have been exposed of which I would have caught earlier had I been able to safely develop over clearnet (I had assumed some of these issues were transproxy-related and I also did not have proper debugging output until fe9654af467ebf8aae16fecf24cda9db39854ada). One of these core issues is an apparently very broken NTCP implementation.

An example, as seen in **every** NTCP session (**Edit:** i.e., every session that successfully connects and completes the 4-phase message sequence):

```
2016:05:29|06:44:40.300303 default:default              NFO             NTCPServer: connected to 184.58.98.218:24234
2016:05:29|06:44:40.300510 default:default              DBG             NTCPSession: [oT4e] 184.58.98.218:24234 *** Phase1, preparing
2016:05:29|06:44:40.300686 default:default              DBG             NTCPSession: [oT4e] 184.58.98.218:24234 <-- Phase1, sending
2016:05:29|06:44:40.300866 default:default              DBG             NTCPSession: [oT4e] 184.58.98.218:24234 *** scheduling termination
2016:05:29|06:44:40.301042 default:default              DBG             NTCPSession: [oT4e] 184.58.98.218:24234 --> Phase1 sent, receiving
2016:05:29|06:44:40.554746 default:default              DBG             NTCPSession: [oT4e] 184.58.98.218:24234 *** Phase2 received, processing
2016:05:29|06:44:40.554968 default:default              DBG             NTCPSession: [oT4e] 184.58.98.218:24234 *** creating shared key
[...]

Issue #184, opened 2016-05-28
Implement options to choose transports


anonimal:
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

*Place an X inside the bracket to confirm*
- [X] I confirm.

---

Implement options to use either NTCP or SSU instead of both at all times. This will require some overhaul but will also help when debugging a specific transport.

PR #335, opened 2016-09-06
Add transport options, fix various NTCP bugs

### LINKED ISSUE #334: Integer overflow due to more than 5 simultaneous key exchanges

- [x] I confirm.

---

`DHKeysPairSupplier::Run()` at `transports.cc:84`:

```
 80 void DHKeysPairSupplier::Run() {
 81   LogPrint(eLogDebug, "DHKeysPairSupplier: running");
 82   while (m_IsRunning) {
 83     std::size_t num;
 84     while ((num = m_QueueSize - m_Queue.size ()) > 0)
 85       CreateDHKeysPairs(num);
 86     std::unique_lock<std::mutex> l(m_AcquiredMutex);
 87     m_Acquired.wait(l);  // wait for element gets acquired
 88   }
 89 }
```

If ` m_QueueSize < m_Queue.size()`, this causes an overflow leadings to the creation of (literally) 10 quintillion key pairs. Note that `m_QueueSize == 5` by magic.
This can happen in the follow scenario:

1. We're routing lots of  traffic, so we have at least one key exchange going on.
2. Another key exchange starts, so `m_QueueSize - m_Queue.size() == 1`. That means we'll create a new key pair. After that, we have five pairs again.
3. The new key pair gets acquired, so we now have 4 pairs again.
4. The first key exchange finishes, and the DH key pair is returned. We now have five key pairs.
5. The second key exchange finishes, and the DH key pair is returned. We now have six key pairs.
6. A new key exchange starts so we compute `m_QueueSize - m_Queue.size()`, which equals... `18446744073709551615`!

This probably went unnoticed because NTCP was pretty broken until recently (see #187), but it actually happens almost always when running with `--enable-ssu=0` (see #184).

Note also how this is a trivial DoS attack vector.

EinMByte:

- [x] I confirm.

---

Adds `--enable-ntcp` and `--enable-ssu` (#184)
Seems to fix the most critical issues listed in #187.
Fixes #334.

### DIFF
--- a/src/app/daemon.cc
+++ b/src/app/daemon.cc
@@ -93,6 +93,11 @@ bool Daemon_Singleton::Init() {
       i2p::util::config::var_map["reseed-from"].as<std::string>());
   i2p::context.ReseedSkipSSLCheck(
       i2p::util::config::var_map["reseed-skip-ssl-check"].as<bool>());
+  i2p::context.SetSupportsNTCP(
+      i2p::util::config::var_map["enable-ntcp"].as<bool>());
+  i2p::context.SetSupportsSSU(
+      i2p::util::config::var_map["enable-ssu"].as<bool>());
+
   // Initialize the ClientContext
   InitClientContext();
   return true;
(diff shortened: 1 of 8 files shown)


anonimal:
Referencing #191 (re: NTCP, the biggest "leak" was simply the rampant generation of keypairs and not a real "leak". Fixed in d6d952c).

Thanks @EinMByte :+1: :fireworks: :tada:

Issue #155, opened 2016-03-05
Review cpp-netlib for full implementation


anonimal:
**By submitting this issue, I confirm the following:**
- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

_Place an X inside the bracket to confirm_
- [X] I confirm.

---

In [today's meeting](https://github.com/monero-project/kovri/issues/47), we agreed to move on cpp-netlib pending more review and provided that we are able to make certain adjustments to fit our needs. Excerpt from meeting:

```
2016-03-05  * anonimal repasting
2016-03-05  &anonimal 2016-03-05 18:45:33 &anonimal Considering the importance of auto-update,
2016-03-05  &anonimal 2016-03-05 18:45:42 &anonimal I don't want to go cheap on any library so,
2016-03-05  &anonimal 2016-03-05 18:45:56 &anonimal maybe the flexibility of cpp-netlib is worth the cost.
2016-03-05  &anonimal 2016-03-05 18:47:06 &anonimal Or we can just keep doing things by hand and do extensive review over time.
2016-03-05  &anonimal 2016-03-05 18:47:23 &anonimal But that takes time away from other things.
2016-03-05  &anonimal 2016-03-05 18:49:02 &anonimal EinMByte: are you able to vote now or do you need more time?
2016-03-05  &anonimal 2016-03-05 18:58:08 &anonimal EinMByte: last message you received?
2016-03-05  +EinMByte I think auto-update would be easier if we bundle a library
2016-03-05  +EinMByte that is, anything static is easy to auto-update because we don't have to require any updated dependencies
2016-03-05  +EinMByte I'm looking at http::client
2016-03-05  +EinMByte It seems like there is a template parameter called Tag that can be changed
2016-03-05  +EinMByte The default is UDP
2016-03-05  +EinMByte anonimal: https://github.com/cpp-netlib/cpp-netlib/blob/master/boost/network/protocol/http/tags.hpp
2016-03-05  &anonimal Yep, so do you vote yay or nay for full implementation of cpp-netlib or do you want to review more?
2016-03-05  +EinMByte I vote yes on the condition that we can create our own tag
[...]

PR #336, opened 2016-09-06
AddressBook + HTTP: major refactoring, enhancements, fixes. Refs #305

### LINKED ISSUE #305: Improve Address Book

_Place an X inside the bracket to confirm_
- [X] I confirm.

---

Bits and pieces from  `#kovri-dev`:
- [X] Design refactor
- [X] Implementation review + [specification](https://geti2p.net/en/docs/naming)  review
  - review and fill in missing implementation
  - refactor as needed
- [x] Harden implementation and subscription handling
  
  ```
  &anonimal tl;dr
  &anonimal spec: "Naming Rules": https://geti2p.net/en/docs/naming
  &anonimal https://github.com/i2p/i2p.i2p/blob/master/apps/addressbook/java/src/net/i2p/addressbook/AddressBook.java#L47-L74
  &anonimal https://github.com/monero-project/kovri/blob/master/src/client/address_book.cc#L260
  &anonimal Resolving the TODO and expanding what we don't do.
  ```
- [x] Documentation
  - Move filesystem/storage code to `address_book_storgage{.cc,.h}`
  - Doxygen documentation
  - Refactor design layout
    - Flow should appear as linear as possible
    - Supply program flow narrative
- ~~[ ] Finish cpp-netlib HTTP/S refactor~~ **Moved to #338**
  
  ```
  &anonimal So, why I'm even in address book was to finish up the cpp-netlib http/s refactor
  &anonimal and then stumbled onto everything else
  &anonimal Another topic point will be to:
  &anonimal 1) either redirect our default subscription address to a monero-owned server
  &anonimal 2) or kindly ask i2p-projekt.de / i2p-projekt.i2p to house our custom hosts.txt
  &anonimal Also, I'm still unsure of the reason why we should *only* bootstrap hosts.txt over i2p. Hopefully zzz has some input.
  &anonimal I imagine for resiliency, because internet dns can be poopy.
  &anonimal But this is bootstrapping, and hosts.txt is a packaged deal. If it somehow goes missing, we can easily point to a number of sites that house hosts.txt
  * anonimal speaking out loud so I can copypasta into new ticket
  &anonimal I'm complaining about this because I can trivially replace the hacked http code in address book with robust cpp-netlib and then move onto the other issues at hand.
  &anonimal If we decide to still bootstrap hosts.txt (still packaged mind you) over i2p, that will require more work/cpp-netlib research.
  ```
- [X] Possibly start work on #168 before the HTTP/S refactor (if not before, than after). See also #48.
  
  ```
  &anonimal Hmmm, what we really need is a resolved #168 so this and #48 becomes less of an issue.
  ```
- [X] Remove useless code
- [ ] Resolve remaining assigned TODO's

I will add more issues, or adjust the issues presented, as work progresses.

Referencing #296 #155

anonimal:

- [X] I confirm.

---

This addresses most of #305 but **the work is not yet complete**. Most notably unfinished (as noted in TODO's) is the [cpp-netlib refactor](https://github.com/monero-project/kovri/issues/155) in HTTP. 

Design-wise, ```AddressBook``` is still a bit hairy because of the update-timer - but the design is functional (enough). I want to get this into master so I can take a step back & gain some perspective and so someone else can easily contribute if they'd like.

Also, though we can have multiple publishers, we currently throw everything into one subscription file (overwritten, not appended) and only use one subscription file instead multiple (e.g., default, custom, private). This area of the spec is up for debate @EinMByte @fluffypony. This feature could be nice (low priority) but I think it could fit well in its own ticket instead of throwing it with #305 (but I'm also antsy to close #305  and move on, so I may be biased).

With that said, **this PR is in a stable state**: clearnet and in-net subscription handling is functional, tests pass, etc. More details are given in commit log messages.


src/client/address_book.cc
```diff
@@ -55,221 +55,292 @@
...
+  // edit publisher's file manually with any effect after router start
+  if (m_Subscribers.empty()) {
+    auto publishers = GetDefaultPublishersFilename();
+    LogPrint(eLogInfo, "AddressBook: loading publisher file ", publishers);
+    std::ifstream file(
+        i2p::util::filesystem::GetFullPath(publishers),
+        std::ofstream::in);
+    if (file.is_open()) {
+      std::string publisher;
+      while (!file.eof()) {
```
EinMByte: @anonimal Can we use `while (!getline(file, publisher))`?

src/client/address_book.cc
```diff
@@ -55,221 +55,292 @@
...
+        getline(file, publisher);
+        if (!publisher.length())
+          continue;  // skip empty line
```
EinMByte: @anonimal Only empty or empty with possible whitespace?

src/client/address_book.cc
```diff
@@ -55,221 +55,292 @@
...
-  auto filename = GetPath() / (address.GetIdentHash().ToBase32() + ".b32");
-  std::ofstream f(filename.string(), std::ofstream::binary | std::ofstream::out);
-  if (f.is_open()) {
-    size_t len = address.GetFullLen();
-    auto buf = std::make_unique<std::uint8_t[]>(len);
-    address.ToBuffer(buf.get(), len);
-    f.write(reinterpret_cast<char *>(buf.get()), len);
+void AddressBook::SubscriberUpdateTimer(
+    const boost::system::error_code& ecode) {
+  if (ecode != boost::asio::error::operation_aborted) {
```
EinMByte: @anonimal Maybe handle the error first, if possible.

src/client/address_book.cc
```diff
@@ -55,221 +55,292 @@
...
+    m_SubscriptionIsLoaded = true;
+    return;
+  }
+  // If available, load default subscription from file
+  auto filename = GetDefaultSubscriptionFilename();
+  std::ifstream file(
+      i2p::util::filesystem::GetFullPath(filename),
+      std::ofstream::in);
+  LogPrint(eLogInfo, "AddressBook: loading subscription ", filename);
+  if (file.is_open()) {  // Open subscription, validate, and save to storage
```
EinMByte: @anonimal Maybe we should use `operator bool` to check whether the file is OK for reading. `is_open()` only checks if there is an associated file.

src/client/address_book.cc
```diff
@@ -55,221 +55,292 @@
...
+      " Last-Modified: ", m_HTTP.GetPreviousLastModified());
+  bool download_result = m_HTTP.Download();
+  if (download_result) {
+    std::stringstream stream(m_HTTP.GetDownloadedContents());
+    if (!m_Book.ValidateSubscriptionThenSaveToStorage(stream)) {
+      // Error during validation or storage, download again later
+      download_result = false;
+    }
+  }
+  m_Book.HostsDownloadComplete(download_result);
```
EinMByte: @anonimal I don't think this is thread safe.

src/client/address_book.cc
```diff
@@ -55,221 +55,292 @@
...
-        "AddressBook: subscription is downloading, waiting for termination");
-    for (int i = 0; i < 30; i++) {
-      if (!m_IsDownloading) {
-        LogPrint(eLogInfo, "AddressBook: subscription download complete");
-        break;
+  // Host per line and total number of host addresses
+  std::string host;
+  std::size_t num_addresses = 0;
+  // Read through stream, add to address book
+  while (!stream.eof()) {
```
EinMByte: @anonimal Can we use `while (getline(stream, host))`?

src/client/address_book.cc
```diff
@@ -55,221 +55,292 @@
...
+    getline(stream, host);
+    if (!host.length())
+      continue;  // skip empty line
+    // File ready, write/overwrite to disk
+    if (m_SubscriptionFileIsReady)
+      file << host << std::endl;
```
EinMByte: @anonimal `std::endl` also flushes. It may be better to write `'\n'` and flush at the end.

src/client/address_book.cc
```diff
@@ -55,221 +55,292 @@
...
+    // Parse host from address, save to address book
+    std::size_t pos = host.find('=');
+    if (pos != std::string::npos) {
+      std::string name = host.substr(0, pos++);
+      std::string addr = host.substr(pos);
```
EinMByte: @anonimal I'd make these `std::string` objects `const` if possible (but that's just me really).

src/client/address_book.cc
```diff
@@ -55,221 +55,292 @@
...
+      i2p::data::IdentityEx ident;
+      if (ident.FromBase64(addr)) {
+        m_Addresses[name] = ident.GetIdentHash();
+        m_Storage->AddAddress(ident);
+        num_addresses++;
```
EinMByte: @anonimal Add TODO to check for overflow?

src/client/address_book.h
```diff
@@ -35,151 +35,239 @@
 
 #include <boost/asio.hpp>
 
-#include <string.h>
-
+#include <atomic>
+#include <cstdint>
 #include <iostream>
```
EinMByte: @anonimal We probably don't need `<iostream>` here. Can we do with `<iosfwd>`?

src/client/address_book_storage.cc
```diff
@@ -0,0 +1,161 @@
...
+          "AddressBookFilesystemStorage: failed to create addressbook directory");
+  }
+}
+
+bool AddressBookFilesystemStorage::GetAddress(
+    const i2p::data::IdentHash& ident,
+    i2p::data::IdentityEx& address) const {
+  auto filename = GetAddressBookPath() / (ident.ToBase32() + ".b32");
+  std::ifstream file(filename.string(), std::ifstream::binary);
+  if (file.is_open()) {
```
EinMByte: @anonimal Prefer `if (file)`?

src/client/address_book_storage.cc
```diff
@@ -0,0 +1,161 @@
...
+    file.seekg(0, std::ios::end);
+    std::size_t len = file.tellg();
```
EinMByte: @anonimal `const`?

src/client/address_book_storage.cc
```diff
@@ -0,0 +1,161 @@
...
+  } else {
+    return false;
+  }
+}
+
+void AddressBookFilesystemStorage::AddAddress(
+    const i2p::data::IdentityEx& address) {
+  auto filename = GetAddressBookPath() / (address.GetIdentHash().ToBase32() + ".b32");
+  std::ofstream file(filename.string(), std::ofstream::binary | std::ofstream::out);
+  if (file.is_open()) {
```
EinMByte: @anonimal `if (file)`

src/client/address_book_storage.cc
```diff
@@ -0,0 +1,161 @@
...
+    return true;
+  } else {
+    return false;
+  }
+}
+
+void AddressBookFilesystemStorage::AddAddress(
+    const i2p::data::IdentityEx& address) {
+  auto filename = GetAddressBookPath() / (address.GetIdentHash().ToBase32() + ".b32");
+  std::ofstream file(filename.string(), std::ofstream::binary | std::ofstream::out);
```
EinMByte: @anonimal I don't think it's necessary to pass `std::ofstream::out` again.

src/client/address_book_storage.cc
```diff
@@ -0,0 +1,161 @@
...
+**/
+
+std::size_t AddressBookFilesystemStorage::Load(
+    std::map<std::string, i2p::data::IdentHash>& addresses) {
+  std::size_t num = 0;
+  auto filename = GetAddressBookPath() / GetDefaultAddressesFilename();
+  std::ifstream file(filename.string(), std::ofstream::in);
+  if (file.is_open()) {
+    addresses.clear();
+    while (!file.eof()) {
```
EinMByte: @anonimal `while (getline(file, host))`

src/client/address_book_storage.cc
```diff
@@ -0,0 +1,161 @@
...
+    boost::filesystem::remove(filename);
+}
+**/
+
+std::size_t AddressBookFilesystemStorage::Load(
+    std::map<std::string, i2p::data::IdentHash>& addresses) {
+  std::size_t num = 0;
+  auto filename = GetAddressBookPath() / GetDefaultAddressesFilename();
+  std::ifstream file(filename.string(), std::ofstream::in);
+  if (file.is_open()) {
```
EinMByte: @anonimal `if (file)`

src/client/address_book_storage.cc
```diff
@@ -0,0 +1,161 @@
...
+  if (boost::filesystem::exists(filename))
+    boost::filesystem::remove(filename);
+}
+**/
+
+std::size_t AddressBookFilesystemStorage::Load(
+    std::map<std::string, i2p::data::IdentHash>& addresses) {
+  std::size_t num = 0;
+  auto filename = GetAddressBookPath() / GetDefaultAddressesFilename();
+  std::ifstream file(filename.string(), std::ofstream::in);
```
EinMByte: @anonimal I don't think we need `std::ofstream::in` (`std::ifstream::in`)?

src/client/address_book_storage.cc
```diff
@@ -0,0 +1,161 @@
...
+        "AddressBookFilesystemStorage: ", filename, " not found");
+  }
+  return num;
+}
+
+std::size_t AddressBookFilesystemStorage::Save(
+    const std::map<std::string, i2p::data::IdentHash>& addresses) {
+  std::size_t num = 0;
+  auto filename = GetAddressBookPath() / GetDefaultAddressesFilename();
+  std::ofstream file(filename.string(), std::ofstream::out);
```
EinMByte: @anonimal Don't need `std::ofstream::out`.

src/client/address_book_storage.cc
```diff
@@ -0,0 +1,161 @@
...
+  if (file.is_open()) {
+    for (auto it : addresses) {
```
EinMByte: @anonimal Use `const auto& it`?

src/client/address_book_storage.cc
```diff
@@ -0,0 +1,161 @@
...
+  }
+  return num;
+}
+
+std::size_t AddressBookFilesystemStorage::Save(
+    const std::map<std::string, i2p::data::IdentHash>& addresses) {
+  std::size_t num = 0;
+  auto filename = GetAddressBookPath() / GetDefaultAddressesFilename();
+  std::ofstream file(filename.string(), std::ofstream::out);
+  if (file.is_open()) {
```
EinMByte: @anonimal `if (file)`

src/client/address_book_storage.h
```diff
@@ -0,0 +1,176 @@
...
+ * Parts of the project are originally copyright (c) 2013-2015 The PurpleI2P Project          //
+ */
+
+#ifndef SRC_CLIENT_ADDRESS_BOOK_STORAGE_H_
+#define SRC_CLIENT_ADDRESS_BOOK_STORAGE_H_
+
+#include <boost/filesystem.hpp>
+
+#include <cstdint>
+#include <iostream>
```
EinMByte: @anonimal Probably not necessary, at most `<iosfwd>`.

src/client/address_book_storage.h
```diff
@@ -0,0 +1,176 @@
...
+#include <map>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "identity.h"
+#include "core/util/filesystem.h"
+#include "router_context.h"
+#include "util/base64.h"
```
EinMByte: @anonimal Don't know if we need this.

src/client/address_book_storage.h
```diff
@@ -0,0 +1,176 @@
...
+  virtual void AddAddress(
+      const i2p::data::IdentityEx& address) = 0;
+
+  /**
+  // TODO(unassigned): currently unused
+  virtual void RemoveAddress(
+      const i2p::data::IdentHash& ident) = 0;
+  **/
+
+  virtual std::size_t Load(
```
EinMByte: @anonimal Do we need the polymorphism, or just the base class? If the latter, don't add `virtual` methods.

src/core/util/http.cc
```diff
@@ -44,63 +42,70 @@
...
 
 #include "router_context.h"
+#include "client/client_context.h"
+#include "client/address_book.h"
 #include "util/filesystem.h"
 
 namespace i2p {
 namespace util {
-namespace http {
+namespace http {  // TODO(anonimal): consider removing this namespace (its not needed)
```
EinMByte: @anonimal What do you mean by "not needed"?

src/core/util/http.cc
```diff
@@ -44,63 +42,70 @@
...
-    const std::string& data) {
-  std::string res(data);
-  for (size_t pos = res.find('%');
-      pos != std::string::npos;
-      pos = res.find('%', pos + 1)) {
-    const char c = strtol(res.substr(pos + 1, 2).c_str(), NULL, 16);
-    res.replace(pos, 3, 1, c);
+bool HTTP::HostIsI2P() {
+  auto uri = GetURI();
+  if (uri.host().substr(uri.host().size() - 4) == ".i2p") {
```
EinMByte: @anonimal Handle error first.

src/core/util/http.cc
```diff
@@ -44,63 +42,70 @@
...
-  if (!uri.Parse(address))
-    return false;
-  namespace http = boost::network::http;
-  namespace network = boost::network;
-  http::client::options options;
+bool HTTP::DownloadViaClearnet() {
+  auto uri = GetURI();
+  // Create and set options
+  Options options;
+  options.timeout(45);  // Java I2P defined
```
EinMByte: @anonimal But should it be a constant?

src/core/util/http.cc
```diff
@@ -109,87 +114,255 @@ bool HTTP::Download(
...
+        request << boost::network::header("If-Modified-Since", GetPreviousLastModified());
+    } else {
+      // Set path to test against for future download (if this is a single instance)
+      SetPath(uri.path());
+    }
+    // Create response object, send request and receive response
+    Response response = client.get(request);
+    // Test HTTP response status code
+    switch (response.status()) {
+      case 200:  // OK  (also, cached version does not match, so we download)
```
EinMByte: @anonimal Do we want constants for HTTP codes?

src/core/util/http.cc
```diff
@@ -109,87 +114,255 @@ bool HTTP::Download(
...
     }
-    std::stringstream ss;
-    if (isChunked)
-      MergeChunkedResponse(response, ss);
-    else
-      ss << response.rdbuf();
-    return ss.str();
+    // Get content after header
+    std::stringstream content;
+    while (!m_Response.eof()) {
```
EinMByte: @anonimal The last `eof()` I'll mention.

src/core/util/http.h
```diff
@@ -33,17 +33,21 @@
...
-#include <openssl/err.h>
-#include <openssl/ssl.h>
+// cpp-netlib
+#define BOOST_NETWORK_ENABLE_HTTPS
+#include <boost/network/include/http/client.hpp>
+#include <boost/network/uri.hpp>
 
+#include <cstdint>
 #include <fstream>
 #include <iostream>
```
EinMByte: @anonimal Remove both of these includes.

anonimal:
JFTR: nearly all the previous comments refer to old code that I chose not to touch because A) it worked (enough) and B) fixing/cleaning up would equate to more work for me :laughing:  but yes, this all should be fixed before this is merged.

Thanks for the insight and for cracking the whip (and full inbox) @EinMByte :wink: 

Also, what I'm going to do is open two new tickets: 1 for spec discussion proposed above and 1 for finishing HTTP/cpp-netlib (its now out of scope for address book). After that, we should mark #305 as resolved.

src/core/util/http.cc
anonimal: We currently only process 2 codes and the codes are not used elsewhere. If we expand our client later, then I'd say yes.

src/core/util/http.cc
anonimal: It could but, unless its reused elsewhere (used more than once) or part of a bigger ideological container, I don't see the point.

src/core/util/http.cc
anonimal: Its not needed. :smile:  We're nowhere near overcrowding the ```util``` namespace, and namespace ```http``` is too small to warrant its own namespace. I've been wanting to open a ticket re: namespace design for a long time now.  When that happens, I'm sure this will be mentioned.

src/client/address_book_storage.h
anonimal: This was pre-existing. AFAICT, as an interface, it may have been going in the direction of being reused elsewhere (its old code, who knows). I don't see why we would do that now, I'll look closer.

src/core/util/http.cc
EinMByte: Ok, add a TODO in that case.

anonimal: cpp-netlib may have this readily available. If not, I'll create constants and then a TODO to create a container if we expand use-case.

src/client/address_book.cc
anonimal: After removing `getline()` from within the block, I think we'd want `while (getline(file, publisher))` or else we'll skip the block entirely.

src/client/address_book.cc
anonimal: Only empty. This should be fixed (though it will be caught during URI sanity test).

src/client/address_book.cc
anonimal: And why do we limit ourselves to `operation_aborted`? I know, at a minimum

>If a wait handler is cancelled, ... boost::asio::error::operation_aborted.

But, specifically here, can we not catch more (or all)?

src/client/address_book.cc
EinMByte: Yes, the `!` was a typo.

src/client/address_book.cc
EinMByte: [This](http://www.boost.org/doc/libs/1_61_0/doc/html/boost_asio/reference/basic_deadline_timer/async_wait.html) seems to say that `operation_aborted` is the only possible error. But we should be defensive and check for any error.

src/core/util/http.cc
anonimal: Its hard to argue style. AFAICT, we don't have anything in writing about whether `if` or `else` should handle errors. In this case, `if` handling will look cleaner.

EinMByte: @anonimal The motivation is too avoid the unnecessary `else` altogether. If an error should cause a return, it should be done as quickly as possible IMHO.

anonimal:
There are a few more TODO's I'd like to see resolved before closing #305. Other than that, I'll proceed as planned with opening new related tickets.

PR #364, opened 2016-09-16
Coverity: resolve medium + high defects. Refs #263

anonimal:

- [X] I confirm.

---

Not noted in f0de48f is defect type `OVERRUN`.

References #263.

### DIFF
--- a/src/client/client_context.h
+++ b/src/client/client_context.h
@@ -72,6 +72,7 @@ class ClientContext {
 
   std::shared_ptr<ClientDestination> CreateNewLocalDestination(
       bool is_public = false,
+      // TODO(anonimal): we don't need DSA-SHA1 for default
       i2p::data::SigningKeyType sigType = i2p::data::SIGNING_KEY_TYPE_DSA_SHA1,
       const std::map<std::string, std::string>* params = nullptr);  // transient
 
--- a/src/core/identity.cc
+++ b/src/core/identity.cc
@@ -50,19 +50,26 @@
 namespace i2p {
 namespace data {
 
+// TODO(unassigned): identity implementation needs a big refactor
+
+// TODO(unassigned): keep an eye open for alignment issues and for hacks like:
+// copy public and signing keys together
+//memcpy(public_key, keys.public_key, sizeof(public_key) + sizeof(signing_key));
+
 Identity& Identity::operator=(const Keys& keys) {
-  // copy public and signing keys together
-  memcpy(public_key, keys.public_key, sizeof(public_key) + sizeof(signing_key));
+  memcpy(public_key, keys.public_key, sizeof(public_key));
   memset(&certificate, 0, sizeof(certificate));
   return *this;
 }
 
-size_t Identity::FromBuffer(
+// TODO(unassigned): unused, remove after refactor
+/*size_t Identity::FromBuffer(
     const uint8_t* buf,
     size_t) {
-  memcpy(public_key, buf, DEFAULT_IDENTITY_SIZE);
+  //memcpy(public_key, buf, DEFAULT_IDENTITY_SIZE);
+  memcpy(public_key, buf, sizeof(public_key));
   return DEFAULT_IDENTITY_SIZE;
-}
+}*/
 
 IdentHash Identity::Hash() const {
   IdentHash hash;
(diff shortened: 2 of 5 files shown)


src/client/client_context.h
```diff
@@ -72,6 +72,7 @@ class ClientContext {
 
   std::shared_ptr<ClientDestination> CreateNewLocalDestination(
       bool is_public = false,
+      // TODO(anonimal): we don't need DSA-SHA1 for default
```
EinMByte: Why not just change it then?
We should probably make the whole thing configurable.

src/core/identity.cc
```diff
@@ -50,19 +50,26 @@
...
+// TODO(unassigned): identity implementation needs a big refactor
+
+// TODO(unassigned): keep an eye open for alignment issues and for hacks like:
+// copy public and signing keys together
+//memcpy(public_key, keys.public_key, sizeof(public_key) + sizeof(signing_key));
+
 Identity& Identity::operator=(const Keys& keys) {
-  // copy public and signing keys together
-  memcpy(public_key, keys.public_key, sizeof(public_key) + sizeof(signing_key));
+  memcpy(public_key, keys.public_key, sizeof(public_key));
```
EinMByte: What if we _do_ want to copy the private key? Does that ever happen?

src/core/identity.cc
```diff
@@ -50,19 +50,26 @@
...
   memset(&certificate, 0, sizeof(certificate));
   return *this;
 }
 
-size_t Identity::FromBuffer(
+// TODO(unassigned): unused, remove after refactor
```
EinMByte: Maybe remove implementation altogether , and just leave a comment in the header file?

src/core/transport/upnp.cc
```diff
@@ -327,9 +322,14 @@ void UPnP::TryPortMapping(
...
-      // boost::this_thread::sleep_for(); // pre c++11
-      // sleep(20*60); // non-portable
+      // TODO(unassigned): do we really want to retry on *all* errors? (see upnpcommands.h)
+      LogPrint(eLogError,
+          "UPnP: AddPortMapping (", strPort.c_str(),
+          ", ", strPort.c_str(),
+          ", ", m_NetworkAddr,
+          ") failed with code ", r);
+      // Try again later
+      std::this_thread::sleep_for(std::chrono::minutes(20));
```
EinMByte: Another one of those magic constants, consider adding a TODO.

src/client/client_context.h
anonimal: >Why not just change it then?

Out of scope for the fix. One step at a time.

EinMByte: Fair enough, we should make all signature types configurable, eventually though.

src/core/identity.cc
anonimal: >What if we do want to copy the private key?

It should be done elsewhere. This is an abhorrent implementation.

>Does that ever happen?

Because we currently don't subscribe in-net, address book won't touch this code so, **currently** 100% no. Once `client_context.h:75` is resolved, then **always** 100% no.

After the TODO is resolved, we should also be able to begin a bigger refactor. We should also consider eliminating DSA-SHA1 entirely because **we can**.

src/core/identity.cc
anonimal: I want to leave the implementation until **after** the refactor. I hate deleting code that *may* be even the slightest useful in the future. If not, we can clobber it then when we write a cleaner impl.

src/core/transport/upnp.cc
anonimal: The magician works in mysterious ways.

I hoped that [this TODO](https://github.com/monero-project/kovri/pull/364/commits/150bbc7c1605080e3712ba81e8aab0661d38a7d5#diff-2faf3e582c3080ed0cd80f05dab8f1ceR57) would imply `upnp.cc:332` but we shouldn't take chances. I'll add the TODO.

src/core/identity.cc
anonimal: Referencing #366

src/client/client_context.h
anonimal: Opened #367

src/core/identity.cc
EinMByte: We should still support DSA-SHA1 for now (to be able to communicate with old routers), I think. But it certainly shouldn't be the default signature type for anything in the kovri router.

src/core/identity.cc
EinMByte: Ok

PR #365, opened 2016-09-16
COVERITY :: see #263 ; cid135926

guzzijones:

- [x ] I confirm.

---

Coverity fix cid135926

### DIFF
--- a/src/app/daemon_linux.cc
+++ b/src/app/daemon_linux.cc
@@ -90,11 +90,14 @@ bool DaemonLinux::Start() {
     chdir(d.c_str());
     // close stdin/stdout/stderr descriptors
     ::close(0);
-    ::open("/dev/null", O_RDWR);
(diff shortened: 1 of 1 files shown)


src/app/daemon_linux.cc
```diff
@@ -90,14 +90,14 @@ bool DaemonLinux::Start() {
     chdir(d.c_str());
     // close stdin/stdout/stderr descriptors
     ::close(0);
-    if(::open("/dev/null", O_RDWR)<0)
-	    exit(1);
+    if (::open("/dev/null", O_RDWR) < 0 )
+      exit(1);
```
anonimal: Thanks @guzzijones. We should probably return `false` instead of `exit(1)` so our daemon instance can cleanup with `Stop()` (see `main.cc`)

guzzijones: Ok. I made that adjustment.

PR #371, opened 2016-09-18
COVERITY :: see #263; add move constructor see cid 143597

guzzijones:

- [x] I confirm.

---

add move constructor; fix cid 143597.

### DIFF
--- a/src/core/identity.h
+++ b/src/core/identity.h
@@ -213,6 +213,13 @@ class IdentityEx {
   IdentityEx(
       const IdentityEx& other);
 
+  IdentityEx( IdentityEx && other){
+      m_Verifier=std::move(other.m_Verifier);
+      m_ExtendedLen=other.m_ExtendedLen;
+      m_ExtendedBuffer=std::move(other.m_ExtendedBuffer);
+
+  }
+
   IdentityEx& operator=(const IdentityEx& other);
 
   IdentityEx& operator=(const Identity& standard);


src/core/identity.h
```diff
@@ -213,6 +213,13 @@ class IdentityEx {
   IdentityEx(
       const IdentityEx& other);
 
+  IdentityEx( IdentityEx && other){
+      m_Verifier=std::move(other.m_Verifier);
+      m_ExtendedLen=other.m_ExtendedLen;
+      m_ExtendedBuffer=std::move(other.m_ExtendedBuffer);
+
+  }
+
```
guzzijones: i will update with cpplint fixes

guzzijones: looks like there are some issues with adding explicit to the constructor that takes one argument.

```
/home/ubuntu/kovri/src/core/garlic.cc: In member function ‘void i2p::garlic::GarlicDestination::AddSessionKey(const uint8_t*, const uint8_t*)’:
/home/ubuntu/kovri/src/core/garlic.cc:396:27: error: no matching function for call to ‘i2p::crypto::CBCDecryption::SetKey(const uint8_t*&)’
     decryption->SetKey(key);

```

I will take a look at fixing that as well unless you tell me otherwise.  I did some research and it sounds like adding explicit will help avoid bugs.

anonimal:
JFTR: I spoke with @guzzijones and he's working on this but 1316468 is *not* how one does proper move assignment (especially with `IdentityEx`). Merging this will break functionality so let's not merge until this is fixed.

Issue #140, opened 2016-02-18
SSU Refactor


anonimal:
This is in-progress work currently being done by @EinMByte (https://github.com/einmbyte/kovri/tree/ssu). Ticketing for housekeeping.

anonimal:
From [today's meeting](https://github.com/monero-project/kovri/issues/47):
```
+EinMByte Ok, so what needs to be done on SSU:
+EinMByte - Tests for most of the code, this should be easy to do with the new architecture
+EinMByte - A packet builder class.
+EinMByte - More error handling. Currently the new code allows for a lot of error handling but not everything is implemented.
```

EinMByte:
Some more specifics:

On the parser side, I'm currently integrating the new code. SSUSession.h is mostly done, SSUData.h requires some additional work. Some additional required changes for `SSUPacketParser`:
* Currently, it only throws exceptions when something would go fatally wrong. Ideally, we should also do strict checks against the specification. Timestamps etc. also need to be checked.
* Make it handle (AES) decryption and MAC checking. At the moment we're still doing a lot of this stuff in `SSUSession`.
* Write tests for all the packet types that currently don't have any tests.

After that, the main thing is going to be building packets. The idea is to create a `SSUPacketBuilder` class, which converts arbitrary `SSUPacket`s to a byte array.

EinMByte:
Status update:
* Parser code is implemented and seems to be working.
* `SSUData` is not yet using the new code to parse fragments.
* Currently rewriting the code to build messages.

anonimal:
I'm thoroughly impressed with everything I've seen so far. IMHO, this has been much needed work and the surgery you've performed will save a much needed leg. Thank you.

For now, I'd like to assist while you continue the heavier end of the work. Some things I'd like to PR to your branch before this is merged:

- cstdint refactor to fully qualify all integer types
- C++11/14 refactor where appropriate
- [Contributing guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md) refactor
- Documentation/doxygen for each (or most) function/class

Please let me know if you've already done some/most of this work or if you want me to stay away completely for now. I'm proposing this while its fresh on my mind and because you're available for the time being - so it would be nice to be on the same page with you work-wise.

EinMByte:
@anonimal I postponed most of these things. So feel free to make changes, additional tests are also welcome (and maybe best to add those first, since that won't interfere a lot).

anonimal:
[...]

PR #372, opened 2016-09-19
Partially refactored SSU

EinMByte:

- [X] I confirm.

---

Finally, here is the first part of the refactored SSU code.


src/core/transport/ssu_packet.cc
```diff
@@ -889,170 +861,130 @@ std::unique_ptr<SSUSessionDestroyedPacket> SSUPacketParser::ParseSessionDestroye
...
         header->GetExtendedOptionsData(),
         header->GetExtendedOptionsSize());
   }
 }
 
-void WriteSessionRequest(
-    std::uint8_t*& buf,
-    SSUSessionRequestPacket* packet) {
-  WriteData(buf, packet->GetDhX(), static_cast<std::size_t>(SSUSize::DHPublic));
-  // TODO(EinMByte): Check for overflow
```
anonimal: `InputByteStream::ConsumeData(std::size_t amount)` and `OutputByteStream::ProduceData(std::size_t amount)` resolves these TODO's, yes?

src/core/transport/ssu_packet.h
```diff
@@ -769,175 +771,87 @@ class SSUPacketParser {
...
+  /// @brief Gets padded size of message
+  /// @param size Size of message
+  static std::size_t GetPaddedSize(
+      std::size_t size);
+
+  /// @brief Writes an SSU header into a data buffer.
+  /// @pre The data buffer must be sufficiently large.
+  /// @param data Reference to pointer to data
+  /// @param header Pointer to SSU header
+  void WriteHeader(SSUHeader* header);
```
anonimal: Without getting into all the arguments re: passing pointer ownership, was there a particular reason you decided to switch to raw pointers? IIRC, I believe you were passing `unique_ptr` at one point for all current `Write` functions and I do like the homogeneity that was provided, but that's just me. I can't find a particular commit in this branch though, maybe it's just my wishful thinking?

anonimal:
Fantastic work @EinMByte.  Well done if I do say so myself. :+1: :smile: IMHO, this quality of code is what kovri should have inherited from in the beginning. Fortunately for us, we're building a stronger foundation for a better tomorrow! :sunrise: 

JFTR: I've been able to review this in real-time as the branch has progressed. I'm very glad we're finally able to start merging :smiley:

anonimal:
Referencing #140

src/core/transport/ssu_packet.cc
EinMByte: @anonimal Yes, they should be resolved now. Will fix.

src/core/transport/ssu_packet.h
EinMByte: Passing a `unique_ptr` implies that you move ownership, which you don't want to do in this situation. I pass a raw pointer here because I don't need any memory management, and because I want to allow that `header` can be a reference to stack memory (in which case any `delete` would be inappropriate).

What we should consider, though, is to pass `header` as a pointer-to-const.

PR #401, opened 2016-10-15
Added recommendation for users behind NAT

94b1:

- [X] I confirm.

---

Added suggestion about installing MiniUPnP for users behind a restrictive NAT firewall. 

In addition, what about linking to their MiniUPnP's [releases page](https://github.com/miniupnp/miniupnp/releases) directly?


anonimal:
>In addition, what about linking to their MiniUPnP's releases page directly?

Well, the github releases page provides the same source files that are on miniupnp page but the github page also provides zip files (miniupnp page only provides tarballs). Also, miniupnp is so common that it's easily attainable with a package manager, so maybe linking anywhere outside of github is more confusing than not?

PR #412, opened 2016-10-25
NetDB: refactored redundant code checking file path

olarks:
**By submitting this pull-request, I confirm the following:**

- I have read and understood the [Contributing Guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that another pull-request for this purpose does not exist.
- I have considered and confirmed that this submission will be valuable to others.
- I accept that this submission may not be used and that this pull-request may be closed by the will of the maintainer.
- I give this submission freely under the BSD 3-clause license.

*Place an X inside the bracket (or click the box in preview) to confirm*
- [ X] I confirm.

---

*Enter your pull-request summary here*

PR #413, opened 2016-10-27
NetDb: styling fixes and refactoring of arbitrary values

olarks:

- [ X] I confirm.

---

### DIFF
--- a/src/core/net_db.cc
+++ b/src/core/net_db.cc
@@ -163,17 +163,29 @@ void NetDb::Run() {
       if (ts - last_exploratory >=
           static_cast<uint16_t>(NetDbDuration::ExploreTunnelsInterval)) {
         auto num_routers = m_RouterInfos.size();
+        // evaluates if a router has a sufficient number of known routers
+        // to use for building tunnels, if less than 800 routers
+        // are known, then more exploratory tunnels will be created
+        // to find more routers for tunnel building
         // TODO(anonimal): research these numbers
-        if (num_routers < 2500 ||
-            ts - last_exploratory >=
-            static_cast<uint16_t>(NetDbDuration::ExploreTunnelsInterval)) {
+        if (num_routers <
+            static_cast<uint16_t>(NetDbRouterPrefs::RouterNumUpperBoundThreshold)   // 2500
+            || ts - last_exploratory >=
+               static_cast<uint16_t>(NetDbDuration::ExploreTunnelsInterval)) {
           if (num_routers > 0) {
-            num_routers = 800 / num_routers;
+            num_routers =
+              static_cast<uint16_t>(NetDbRouterPrefs::RouterNumLowerBoundThreshold)   // 800
+              / num_routers;
           }
-          if (num_routers < 1)
-            num_routers = 1;
-          if (num_routers > 9)
-            num_routers = 9;
+          if (num_routers <
+              static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterUpperBound)) {    // 1
+            num_routers =
+              static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterUpperBound);
+          }
+          if (num_routers >
+              static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterLowerBound))    // 9
+            num_routers =
+              static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterLowerBound);
           m_Requests.ManageRequests();
           Explore(num_routers);
           last_exploratory = ts;
--- a/src/core/net_db.cc
+++ b/src/core/net_db.cc
@@ -106,7 +106,7 @@ void NetDb::Run() {
       // for messages to be received
       auto msg =
         m_Queue.GetNextWithTimeout(
-            static_cast<std::size_t>(NetDbDuration::WaitForMessageTimeout));
+            static_cast<uint16_t>(NetDbDuration::WaitForMessageTimeout));
       if (msg) {
         int num_msgs = 0;
         while (msg) {
@@ -140,12 +140,12 @@ void NetDb::Run() {
       uint64_t ts = i2p::util::GetSecondsSinceEpoch();
       // builds tunnels for requested destinations every 15 seconds
       if (ts - last_manage_request >=
-          static_cast<std::size_t>(NetDbDuration::ManageRequestsInterval)) {
+          static_cast<uint16_t>(NetDbDuration::ManageRequestsInterval)) {
         m_Requests.ManageRequests();
         last_manage_request = ts;
       }
       // save routers, manage leasesets and validate subscriptions every minute
-      if (ts - last_save >= static_cast<std::size_t>(NetDbDuration::SaveInterval)) {
+      if (ts - last_save >= static_cast<uint16_t>(NetDbDuration::SaveInterval)) {
         if (last_save) {
           SaveUpdated();
           ManageLeaseSets();
@@ -154,19 +154,19 @@ void NetDb::Run() {
       }
       // publishes router info to a floodfill every 40 minutes
       if (ts - last_publish >=
-	  static_cast<std::size_t>(NetDbDuration::PublishRouterInfoInterval)) {
+	  static_cast<uint16_t>(NetDbDuration::PublishRouterInfoInterval)) {
         Publish();
         last_publish = ts;
       }
       // builds exploratory tunnels every 30 seconds to find more peers
       // to be used for tunnel building
       if (ts - last_exploratory >=
-          static_cast<std::size_t>(NetDbDuration::ExploreTunnelsInterval)) {
+          static_cast<uint16_t>(NetDbDuration::ExploreTunnelsInterval)) {
         auto num_routers = m_RouterInfos.size();
         // TODO(anonimal): research these numbers
         if (num_routers < 2500 ||
             ts - last_exploratory >=
-            static_cast<std::size_t>(NetDbDuration::ExploreTunnelsInterval)) {
+            static_cast<uint16_t>(NetDbDuration::ExploreTunnelsInterval)) {
           if (num_routers > 0) {
             num_routers = 800 / num_routers;
           }
--- a/src/core/net_db.cc
+++ b/src/core/net_db.cc
@@ -162,35 +162,25 @@ void NetDb::Run() {
       // to be used for tunnel building
       if (ts - last_exploratory >=
           static_cast<uint16_t>(NetDbDuration::ExploreTunnelsInterval)) {
-        auto num_routers = m_RouterInfos.size();
+        auto known_routers = m_RouterInfos.size();
+        uint16_t num_routers = 0;
         // evaluates if a router has a sufficient number of known routers
         // to use for building tunnels, if less than 800 routers
         // are known, then more exploratory tunnels will be created
         // to find more routers for tunnel building
         // TODO(anonimal): research these numbers
-        if (num_routers <
-            static_cast<uint16_t>(NetDbRouterPrefs::RouterNumUpperBoundThreshold)   // 2500
-            || ts - last_exploratory >=
-               static_cast<uint16_t>(NetDbDuration::ExploreTunnelsInterval)) {
-          if (num_routers > 0) {
+        if (known_routers > 0 && known_routers <
+            static_cast<uint16_t>(NetDbRouterPrefs::RouterNumLowerBoundThreshold)) {    // 800
+          num_routers =
+            static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterLowerBound);    // 9
+        } else {
             num_routers =
-              static_cast<uint16_t>(NetDbRouterPrefs::RouterNumLowerBoundThreshold)   // 800
-              / num_routers;
-          }
-          if (num_routers <
-              static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterUpperBound)) {    // 1
-            num_routers =
-              static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterUpperBound);
-          }
-          if (num_routers >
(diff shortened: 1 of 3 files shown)


src/core/net_db.cc
```diff
@@ -102,7 +102,9 @@ void NetDb::Run() {
            last_manage_request = 0;
   while (m_IsRunning) {
     try {
-      auto msg = m_Queue.GetNextWithTimeout(15000);  // 15 sec
+      auto msg =
+	m_Queue.GetNextWithTimeout(
+	     static_cast<std::size_t>(NetDbDuration::WaitForMessageTimeout));  // 15 sec
```
anonimal: Like I mentioned [here](https://github.com/monero-project/kovri/pull/412#discussion_r85008739), please try to get your editor in check (no single or odd number of spaces).

src/core/net_db.cc
```diff
@@ -309,11 +316,11 @@ bool NetDb::Load() {
           it1 != end;
           ++it1) {
 #if BOOST_VERSION > 10500
-        const std::string& fullPath = it1->path().string();
+        const std::string& full_path = it1->path().string();
```
anonimal: As noted in the contributing guide, please try not to mix formatting work with non-formatting work unless it's unavoidable to do so (in this case, it's not).

src/core/net_db.h
```diff
@@ -55,6 +55,17 @@
 namespace i2p {
 namespace data {
 
+/// @enum NetDbDuration
+/// @brief Constants defining different refresh intervals
+///   for various NetDb operations
+enum struct NetDbDuration : const std::size_t {
```
anonimal: From what I see below, we won't need `std::size_t` when `std::uint16_t` will suffice.

src/core/net_db.cc
```diff
@@ -134,26 +136,31 @@ void NetDb::Run() {
       if (!m_IsRunning)
         break;
       uint64_t ts = i2p::util::GetSecondsSinceEpoch();
-      if (ts - last_manage_request >= 15) {  // manage requests every 15 seconds
+      if (ts - last_manage_request >=
+          static_cast<std::size_t>(NetDbDuration::ManageRequestsInterval)) {  // manage requests every 15 seconds
```
anonimal: We can go ahead and play with the documentation. For us to specify an interval in comments defeats what we're trying to achieve here. I'd suggest to move the comment above the `if` block and describe what will happen within the block and *why* it will happen. This is general documentation advice (and documenting *why* is the fun part :smile:)

olarks: Ok. Thanks for the advice :smile:

src/core/net_db.cc
```diff
@@ -104,7 +104,7 @@ void NetDb::Run() {
     try {
       auto msg =
 	m_Queue.GetNextWithTimeout(
-	     static_cast<std::size_t>(NetDbDuration::WaitForMessageTimeout));  // 15 sec
+            static_cast<std::size_t>(NetDbDuration::WaitForMessageTimeout));  // 15 sec
```
anonimal: If you're going to newline ident for assignments, I'd advise something like this:

```cpp
auto msg =
  m_Queue.GetNextWithTimeout(  // 2 spaces
      static_cast<std::size_t>(NetDbDuration::WaitForMessageTimeout));  // 4 spaces for function parameters
```

olarks: The git diff is not correct. What you advise is exactly what I changed in the file. Not sure why it is not showing up in diff.

olarks:
With the the most recent refactor, NetDbPrefs::RouterNumHigherBoundThreshold is no longer really needed. I don't think it is good to put a "limit" on the number of known routers, even if it only decreases the number of exploratory tunnels built. It is better to identify what is a "minimum" preferred number of known routers is and work towards that. Anything passed that is icing on the cake.

src/core/net_db.cc
```diff
@@ -163,17 +163,29 @@ void NetDb::Run() {
       if (ts - last_exploratory >=
           static_cast<uint16_t>(NetDbDuration::ExploreTunnelsInterval)) {
         auto num_routers = m_RouterInfos.size();
+        // evaluates if a router has a sufficient number of known routers
+        // to use for building tunnels, if less than 800 routers
+        // are known, then more exploratory tunnels will be created
+        // to find more routers for tunnel building
         // TODO(anonimal): research these numbers
```
anonimal: Since you're working in this area, you can remove this TODO.

src/core/net_db.cc
```diff
@@ -163,17 +163,29 @@ void NetDb::Run() {
...
-        if (num_routers < 2500 ||
-            ts - last_exploratory >=
-            static_cast<uint16_t>(NetDbDuration::ExploreTunnelsInterval)) {
+        if (num_routers <
+            static_cast<uint16_t>(NetDbRouterPrefs::RouterNumUpperBoundThreshold)   // 2500
```
anonimal: We don't need this comment. If we change `RouterNumUpperBoundThreshold` then comments like these are harder to maintain.

src/core/net_db.cc
```diff
@@ -163,17 +163,29 @@ void NetDb::Run() {
...
-            num_routers = 9;
+          if (num_routers <
+              static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterUpperBound)) {    // 1
+            num_routers =
+              static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterUpperBound);
+          }
+          if (num_routers >
+              static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterLowerBound))    // 9
+            num_routers =
+              static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterLowerBound);
```
anonimal: Like everywhere before and after this, use fully-qualified integer values (i.e., `std::unit16_t`). You'll see a lot of this throughout the base and it will all need to be refactored.

src/core/net_db.cc
```diff
@@ -163,17 +163,29 @@ void NetDb::Run() {
...
           }
-          if (num_routers < 1)
-            num_routers = 1;
-          if (num_routers > 9)
-            num_routers = 9;
+          if (num_routers <
+              static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterUpperBound)) {    // 1
+            num_routers =
+              static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterUpperBound);
+          }
```
anonimal: Or maybe,
```cpp
auto exp_upper_bound = static_cast<uint16_t>(NetDbRouterPrefs::ExploreNumRouterUpperBound);
if (num_routers < exp_upper_bound)
  num_routers = exp_upper_bound;
```
etc. This also applies elsewhere.

src/core/net_db.h
```diff
@@ -66,6 +66,17 @@ enum struct NetDbDuration : const uint16_t {
   ExploreTunnelsInterval = 30,  // seconds
 };
 
+/// @enum NetDbRouterPrefs
+/// @brief Constants defining NetDb preferences
+/// for how many known routers are wanted for
+/// a large variety of peers to build tunnels
+enum struct NetDbRouterPrefs : const uint16_t {
+  RouterNumLowerBoundThreshold = 800,   // minimum preferred number of known routers
```
anonimal: You can doxygen these comments too.

src/core/net_db.cc
```diff
@@ -108,7 +108,7 @@ void NetDb::Run() {
         m_Queue.GetNextWithTimeout(
             static_cast<uint16_t>(NetDbDuration::WaitForMessageTimeout));
       if (msg) {
-        int num_msgs = 0;
+        uint8_t num_msgs = 0;
```
anonimal: As mentioned in another review, we should use fully-qualified types (i.e., `std::uint8_t`).

Also, you'll want to work on `if (num_msgs > 100)` in terms of this PR refactor as these are both related.

src/core/net_db.h
```diff
@@ -58,7 +58,7 @@ namespace data {
 /// @enum NetDbDuration
 /// @brief Constants defining different refresh intervals
 ///   for various NetDb operations
-enum struct NetDbDuration : const std::size_t {
+enum struct NetDbDuration : const uint16_t {
```
anonimal: As noted before (I know, repeating myself), fully-qualified types when refactoring.

src/core/net_db.h
```diff
@@ -66,6 +66,17 @@ enum struct NetDbDuration : const uint16_t {
...
 
+/// @enum NetDbRouterPrefs
+/// @brief Constants defining NetDb preferences
+/// for how many known routers are wanted for
+/// a large variety of peers to build tunnels
+enum struct NetDbRouterPrefs : const uint16_t {
+  RouterNumLowerBoundThreshold = 800,   // minimum preferred number of known routers
+  RouterNumUpperBoundThreshold = 2500,  // desirable number of known routers, but not limited to
+  ExploreNumRouterLowerBound = 9,   // number of exploratory tunnels being built to find more routers
+  ExploreNumRouterUpperBound = 1,
```
anonimal: Lower bound 9? Upper bound 1? I see what you're getting at, but maybe we can choose other names.

src/core/net_db.h
```diff
@@ -66,6 +66,17 @@ enum struct NetDbDuration : const uint16_t {
   ExploreTunnelsInterval = 30,  // seconds
 };
 
+/// @enum NetDbRouterPrefs
+/// @brief Constants defining NetDb preferences
+/// for how many known routers are wanted for
+/// a large variety of peers to build tunnels
+enum struct NetDbRouterPrefs : const uint16_t {
```
anonimal: Currently, `Router` is implied, `NetDbPreferences` *could* be better for now.

src/core/net_db.cc
```diff
@@ -149,13 +152,16 @@ void NetDb::Run() {
         }
         last_save = ts;
       }
+      // publishes router info to a floodfill every 40 minutes
```
anonimal: >every 40 minutes

Giving static values within comments defeats the purpose of what we're trying to achieve: flexibility :smile: This also applies elsewhere.

src/core/net_db.h
```diff
@@ -58,35 +58,35 @@ namespace data {
...
   /// @brief number of exploratory tunnels
   ///   to be built for < 800 known router
-  ExploreNumRouterLowerBound = 9,
-  /// @var ExploreNumRouterUpperBound
+  MaxExploratoryTunnels = 9,
+  /// @var MinimumExploratoryTunnels
   /// @brief number of exploratory tunnels
   ///   to be built for 800 < known routers < 2500
-  ExploreNumRouterUpperBound = 1,
+  MinimumExploratoryTunnels = 1,
```
anonimal: May as well go with `MinExploratoryTunnels` for consistency.

src/core/net_db.cc
olarks: The code highlighted has been refactored into a less confusing code block and is similar to what you proposed, I believe.

src/core/net_db.cc
```diff
@@ -138,36 +138,34 @@ void NetDb::Run() {
...
       // to be used for tunnel building
       if (ts - last_exploratory >=
-          static_cast<std::uint16_t>(NetDbDuration::ExploreTunnelsInterval)) {
+          static_cast<std::uint16_t>(NetDbDuration::ExploreInterval)) {
         auto known_routers = m_RouterInfos.size();
         std::uint16_t num_routers = 0;
         // evaluates if a router has a sufficient number of known routers
-        // to use for building tunnels, if less than 800 routers
-        // are known, then more exploratory tunnels will be created
-        // to find more routers for tunnel building
```
anonimal: I thought this whole line was well documented. Maybe `if less than Nth routers are known` instead of deleting all of this?

src/core/net_db.cc
```diff
@@ -138,36 +138,34 @@ void NetDb::Run() {
...
       }
-      // publishes router info to a floodfill every 40 minutes
+      // publishes router info to a floodfill
       if (ts - last_publish >=
 	  static_cast<std::uint16_t>(NetDbDuration::PublishRouterInfoInterval)) {
         Publish();
         last_publish = ts;
       }
-      // builds exploratory tunnels every 30 seconds to find more peers
+      // builds exploratory tunnels to find more peers
```
anonimal: Maybe `builds exploratory tunnels at Nth interval to find more peers`?

src/core/net_db.cc
```diff
@@ -138,36 +138,34 @@ void NetDb::Run() {
...
+      // save routers, manage leasesets and validate subscriptions
       if (ts - last_save >= static_cast<std::uint16_t>(NetDbDuration::SaveInterval)) {
         if (last_save) {
           SaveUpdated();
           ManageLeaseSets();
         }
         last_save = ts;
       }
-      // publishes router info to a floodfill every 40 minutes
+      // publishes router info to a floodfill
```
anonimal: Maybe `publishes router info to a floodfill at Nth interval`?

src/core/net_db.cc
```diff
@@ -125,46 +129,59 @@ void NetDb::Run() {
...
+          num_routers =
+            static_cast<std::uint16_t>(NetDbPrefs::MaxExploratoryTunnels);
+        } else if ((known_routers >
+            static_cast<std::uint16_t>(NetDbPrefs::MinimumKnownRouters) &&
+            known_routers <
+            static_cast<std::uint16_t>(NetDbPrefs::FavouredNumKnownRouters)) ||
+            ts - last_exploratory >=
+            static_cast<std::uint16_t>(NetDbDuration::DelayedExploreInterval)) {
+          num_routers = static_cast<std::uint16_t>(NetDbPrefs::MinExploratoryTunnels);
+        }
```
anonimal: Posting for reference to avoid future confusion:

```
&anonimal | olark: but you didn't keep the same logic: the condition was originally *within threshold or if delayed, then explore tunnels*. Now, it's *within threshold, always explore tunnels regardless of delay*.          
&anonimal | olark: take a close look at the placement of the explore calls within the original block, I think you may have gotten confused (based on what I saw on earlier formatting changes).
```

Issue #98, opened 2016-01-29
Improvements to the library organization


EinMByte:
Fragment from IRC:
```
<EinMByte> 1) Add an (actual) API to the core library.
<EinMByte> 2) Put I2PControl, Streaming etc in the client library.
<EinMByte> 3) Make the main application only use the client library and the API
<EinMByte> So the client library would use *only* the API to do core stuff
<EinMByte> (so the API is used instead of I2CP, since we don't have I2CP)
```

Comment from @fluffypony on the matter:
`<fluffypon> I agree with that approach - core shouldn't be able to provide every-interface-    ever when that can be done with stubby clients`

This would mean that the application will consist of 3 rather than 4 components:
* The core library (`libkovri-core`). The only thing that changes here is the addition of an API that allows loose coupling with the client library, and possibly other applications that want to use the core directly.
* The client library (`libkovri-client`). This will also contain what is currently in `libkovri-api`. This library depends on `libkovri-core`, and uses the API to do whatever is necessary.
* The main application, which should contain the daemon and (command line) interface code. This is the front-end only,. My suggestion is that this repository should only contain a minimal user interface, as is the case right now. There can be separate repositories that use the client and core libraries to provide (various) graphical user interfaces.

Please discuss.

anonimal:
Welcome back @EinMByte. A very warm welcome to you.

My decision for our current organization was based on present [I2P specifications](https://geti2p.net/en/docs). The idea was to parallel the specifications because present APIs are not yet necessary to Kovri nor are they necessary to the Kovri client at this point (JFTR: I'm not opposed to reorganization).

@EinMByte I see where you are going with the joining of client and api, but that is under the assumption that we will even keep those APIs. I've given thought to a "kovri-lite" that works specifically with bitmonero and our present "kovri" which could be a more functional i2p client (that is, if we wish to keep everything and put in the work). With a current spec-appropriate organization, I imagine that maintaining the two repos would be easy as "kovri-lite" would only need core (with proposed API) and possibly a custom client bitmonero side.

@EinMByte @fluffypony I'd like to hear more about your ideas on this. @EinMByte can you provide more support for your proposal? What API do you propose for core? How many more API's do we really need? One of our objectives is to provide I2CP; do you disagree with that direction?

As I've said before, I feel that I2P has too many moving parts so I'd like to see some more benefits to this reorganization. None-the-less, I'm glad we're talking about this and I think we're moving in the right direction :smile:

fluffypony:
[...]

PR #420, opened 2016-10-30
App/Core: fix filesystem-related issues introduced in #414

### LINKED ISSUE #330: Refactor app filesystem code for abstraction / remove redundancy

- [X] I confirm.

---

Discussed after #329, referencing #98:

```
&anonimal EinMByte: Hmm. Ok, so ideally all current filesystem code in app should be in core and we just need a way to interface to it through app?
* anonimal should actually pull up the code
&anonimal lol, I think I answered my own question.
&anonimal So, since app is the 'the third wheel', e.g., the throw-away piece that's only needed to run as stand-alone, EinMByte we could just abstract the calls to *something* in core, and from there within core point to the real filesystem implementation.
&anonimal EinMByte: ^ sound about right?
+EinMByte maybe, but make sure that core know nothing about app
+EinMByte (in think that was the main reason for how it's currently done)
&anonimal Yeah, exactly.
&anonimal But the problem with what we have now is that its redundant code that is confusing to maintain.
```

anonimal:

- [X] I confirm.

---

Our data path / data dir is hardcoded, defined at compile time. There is
nothing flexible about it: once you start the router, you it will work
within the same data directory indefinitely (though there are some
run-time options that provide *some* flexibility where some of the
data within the directory can be found/redirected).

Initializing router context (as if the data dir were a mutable type)
is unnecessary and confusing. Ideally, we would *like* as much
flexibility as possible but, for now; this lends itself to breakage.

Also, this PR moves any filesystem related confusion out of app and
into core because of said reasons (and to remove redundancy). This is
not the *ideal* abstraction I had in mind for #330 but it is good
enough for now (and solves the issue at hand).

Run-time breakage was introduced in 95b35bf because of said reasons.

Closes #330

### DIFF
--- a/src/app/daemon.cc
+++ b/src/app/daemon.cc
@@ -72,8 +72,7 @@ bool Daemon_Singleton::Init() {
   LogPrint(eLogInfo, "Daemon_Singleton: initializing");
   kovri::context.Init(
       kovri::app::var_map["host"].as<std::string>(),
-      kovri::app::var_map["port"].as<int>(),
-      kovri::app::GetDataPath());
+      kovri::app::var_map["port"].as<int>());
   m_IsDaemon = kovri::app::var_map["daemon"].as<bool>();
   auto port = kovri::app::var_map["port"].as<int>();
   kovri::context.UpdatePort(port);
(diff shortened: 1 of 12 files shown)


anonimal:
This should be merged a.s.a.p

anonimal:
Referencing #98.

PR #418, opened 2016-10-30
Client: add doxygen documentation to service header

guzzijones:
**By submitting this pull-request, I confirm the following:**

- I have read and understood the [Contributing Guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that another pull-request for this purpose does not exist.
- I have considered and confirmed that this submission will be valuable to others.
- I accept that this submission may not be used and that this pull-request may be closed by the will of the maintainer.
- I give this submission freely under the BSD 3-clause license.

*Place an X inside the bracket (or click the box in preview) to confirm*
- [x ] I confirm.

---

*Enter your pull-request summary here*

PR #423, opened 2016-10-31
Service :: add comments to service.h document base classes

guzzijones:

- [x] I confirm.

---

### DIFF
--- a/src/client/service.h
+++ b/src/client/service.h
@@ -49,53 +49,63 @@ namespace kovri {
 namespace client {
 
 class I2PServiceHandler;
+/// @class I2PService abstract class for service
+/// @brief base class for i2p service
 class I2PService {
  public:
+  /// @brief  constructor takes ClientDestination, defaults to null ptr
+  /// @param localDestination pointer to local destination
   explicit I2PService(
       std::shared_ptr<ClientDestination> local_destination = nullptr);
   explicit I2PService(
       kovri::core::SigningKeyType kt);
   virtual ~I2PService() { ClearHandlers(); }
-
+  /// @brief add a hander to set
+  /// @param conn I2pService pointer handler to add
   inline void AddHandler(
       std::shared_ptr<I2PServiceHandler> conn) {
     std::unique_lock<std::mutex> l(m_HandlersMutex);
     m_Handlers.insert(conn);
   }
-
+  /// @brief remove handler from set
+  /// @param conn I2PServiceHandler pointer to remove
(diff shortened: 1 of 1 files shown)


src/client/service.h
```diff
@@ -104,10 +114,12 @@ class I2PService {
   std::mutex m_HandlersMutex;
 };
 
-/**
- * Simple interface for I2PHandlers.
- * Allows detection of finalization amongst other things.
- */
+/// @class I2PServiceHandler
+/// @brief Simple interface for I2PHandlers. abstract class for handler
+/// Simple interface for I2PHandlers. abstract class for handler
```
anonimal: `Simple interface for I2PHandlers. abstract class for handler`, why repeat the same line?

src/client/service.h
```diff
@@ -49,53 +49,63 @@ namespace kovri {
...
       StreamRequestComplete stream_request_complete,
       const std::string& dest,
       int port = 0);
-
+  /// @brief return io_service reference of member local destination
   inline boost::asio::io_service& GetService() {
     return m_LocalDestination->GetService();
   }
-
+  /// @brief virtual start service
```
anonimal: We can plainly see that this is an interface but, if the documentation were elaborate as to *why* this must be interfaced, that would be more helpful. In actuality, I don't believe we need this interface (but that's beyond the scope of this task).

If you don't want to spend more time in this area then that's fine; we'll most likely refactor the code anyway (but this would be a good learning experience).

src/client/service.h
```diff
@@ -178,35 +188,43 @@ class TCPIPAcceptor : public I2PService {
...
 
  protected:
   std::string m_Address;
 
  private:
+  /// @brief accept connection ; create socket for handler to listen on
+  /// pass handler function; this is what starts the communication
   void Accept();
+  /// @brief callback function to handle data transfers ;
+  /// @param ecode
```
anonimal: I could've sworn I put this in the last PR but it apparently didn't go through.

You should put something like `boost error code` instead of leaving blank.

anonimal:
Referencing #418

src/client/service.h
guzzijones: i will adjust that.

src/client/service.h
guzzijones: i will add "boost error code"

src/client/service.h
guzzijones: i will add a description of what start does.

Issue #457, opened 2016-11-14
Ubuntu 14.04: unit-tests SIGSEGV upon LogPrint


guzzijones:
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

*Place an X inside the bracket to confirm*
- [x ] I confirm.

---
1. make
2. make tests
3. run kovri-tests

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff657db1b in std::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string(std::string const&) ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) bt
#0  0x00007ffff657db1b in std::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string(std::string const&) ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x0000000000fa57e8 in kovri::core::Log::GetGlobalLogEngine () at /home/ubuntu/kovri/src/core/util/impl/log.cc:471
#2  0x0000000000f6ec5f in DeprecatedLogPrint<char const*> (level=kovri::core::eLogLevelDebug) at /home/ubuntu/kovri/src/core/../core/util/log.h:201
#3  0x0000000000f6c249 in kovri::core::HasAESNI () at /home/ubuntu/kovri/src/core/crypto/impl/cryptopp/aes.cc:52
#4  0x0000000000f6c755 in __static_initialization_and_destruction_0 (__initialize_p=1, __priority=65535) at /home/ubuntu/kovri/src/core/crypto/impl/cryptopp/aes.cc:68
#5  0x0000000000f6c770 in _GLOBAL__sub_I_aes.cc(void) () at /home/ubuntu/kovri/src/core/crypto/impl/cryptopp/aes.cc:593
#6  0x00000000013314fd in __libc_csu_init ()
#7  0x00007ffff5bf8ed5 in __libc_start_main (main=0xf64ec5 <main(int, char**)>, argc=1, argv=0x7fffffffe638, init=0x13314b0 <__libc_csu_init>, fini=<optimized out>,
[...]

PR #459, opened 2016-11-15
UNIT TESTS :: fix 457; segfault unit tests

guzzijones:

- [ x] I confirm.

---

fix segfault on unit tests for ubuntu

### DIFF
--- a/tests/unit_tests/core/crypto/aes.cc
+++ b/tests/unit_tests/core/crypto/aes.cc
@@ -36,6 +36,7 @@
 
 #include "core/crypto/aes.h"
 #include "core/util/impl/log.cc"
+extern std::string kovri::core::g_LogFileName;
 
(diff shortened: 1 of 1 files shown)


anonimal:
Referencing #457, I can't reproduce, see comments in issue.

tests/unit_tests/core/crypto/aes.cc
```diff
@@ -35,6 +35,7 @@
 #include <boost/test/unit_test.hpp>
 
 #include "core/crypto/aes.h"
+#include "core/util/impl/log.cc"
```
anonimal: `#include`'ing implementation files is not necessary (and is a long deprecated practice). *If* we need to include an implementation file, it can be included in CMakeLists.txt - though we don't need to because we link with `libcore` which has the logger implementation already built.

Since I cannot reproduce a segfault, I'll need more information in #457.

PR #461, opened 2016-11-15
NetDb/Crypto: refactor rand router implementation

olarks:

- [X ] I confirm.

---

The original code was only randomly choosing a router indice on the first pass and if a router was not found then on the second pass it would start searching from the beginning(not random). This PR fixes that and on the second pass will start from the random router indice and search the routers in reverse to keep the router being chosen random.

### DIFF
--- a/src/core/crypto/rand.h
+++ b/src/core/crypto/rand.h
@@ -82,6 +82,19 @@ namespace core {
     return ret;
   }
 
+  /// @brief CSPRNG shuffle of a sequence container
+  /// @param begin iterator to the first element in container
+  /// @param end iterator beyond the last element in container
+  /// @note IT must meet the requirements of ForwardIterator.
+  ///   begin, end must meet the requirements of Swappable.
+  template <class IT>
+  void Shuffle(IT begin, IT end) {
+    for (; begin != end; ++begin)
+      std::iter_swap(
+          begin,
+          begin + kovri::core::RandInRange<std::size_t>(0, end - begin - 1));
+  }
+
 }  // namespace core
 }  // namespace kovri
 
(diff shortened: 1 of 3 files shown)


anonimal:
```
&anonimal | olark: re: #461 and GetRandomRouter(), there are much simpler solutions but my biggest questions are: 1. why a 2nd pass and why not a 3rd or 4th, etc.  2. if we truly want random then why are we cherry-picking results
```

I can elaborate more on a solution and on these rhetorical questions when I return (I just wanted to throw in a comment for now).

olarks:
@anonimal: The second pass is only to ensure all the routers are being checked against the desired filter. If x is the random indice and num is the number of known routers then if there is no router found in x < num then the second pass will check the routers 0 < x. A third or fourth pass would be unecessary since the second pass would check all the remaining routers the first pass did not look at.

olarks:
In addition GetRandomRouter is also a template for other GetRandomRouter functions searching specifically for a high bandwidth router or a peer test router etc hence the "cherry picking".

anonimal:
Yes, you've described what's already there but that doesn't answer why the need to do more than one pass. Yes, I know it's a template function but that doesn't answer why we need to cherry-pick.

Instead, we can simply `RandomNumberGenerator::Shuffle` or implement similar and iterate through *all* RI's in one pass until the criteria is/is-not found.

olarks:
Hmmm I do like the shuffle idea. Could just copy the map of routerinfos and randomly choose indices removing the routers not matching the criteria until it finds a match. All in one pass. I'll implement that. I can leave this PR open if you like or close once I have it figured out. :)

anonimal:
Open. Just reference commit hash/pull number in the future commit that clarifies the fix for the present commit. I also try to remember to give credit to collaborator's ideas within the commit message, but that's just me.

olarks:
Yeah for sure, credit where credit is due.

anonimal:
>Could just copy the map of routerinfos

You shouldn't need to copy. Also, you don't need to add more cryptopp abstraction overhead (look at `Shuffle`: it's easy enough to implement ourselves).

PR #462, opened 2016-11-16
I2NP: resolve namespace TODO + message type trivial style refactor

olarks:

- [ X] I confirm.

---

### DIFF
--- a/src/client/api/datagram.cc
+++ b/src/client/api/datagram.cc
@@ -119,36 +119,36 @@ void DatagramDestination::SendDatagramTo(
   } else {
     m_Owner.Sign(buf1, len, signature);
   }
-  std::unique_ptr<I2NPMessage> msg
+  std::unique_ptr<kovri::core::I2NPMessage> msg
       = CreateDataMessage(buf, len + header_len, from_port, to_port);
   std::shared_ptr<const kovri::core::LeaseSet> remote
       = m_Owner.FindLeaseSet(ident);
 
-  ReleasableSharedPtr<I2NPMessage> temp_msg
+  ReleasableSharedPtr<kovri::core::I2NPMessage> temp_msg
       = make_releasable_shared_ptr(msg.release());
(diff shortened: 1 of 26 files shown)


olarks:
@anonimal: Would there be any value in putting those values in seperate enum structs? I wanted to document those values anyway and would rather not undo those name stylings. Unless we are going to stick with the ALL_CAPS variable names. ALL_CAPS was used in other files as well that I was going to refactor as well in the future, but if we are going to stick with ALL_CAPS for those values and we don't want them in an enum struct then I won't refactor those files(I can still document them) to keep all the name styling consistent throughout the codebase.

olarks:
The SSU refactor did put all those values into enum structs, but I guess it depends if we want static_casts all over the place.

olarks:
I could also just close the PR and reopen in the future once I have more I2NP refactoring done. I just figured I would get the styling done first and then comb over the code.

anonimal:
I'll repeat again what I said in IRC:

```
&anonimal       The question is not so much how it's contained nor how pretty it is, the question is more of how they are used. I keep saying see SSU refactor because that's how they should be handled.
&anonimal       That type of math for i2np should not be implemented there.
&anonimal       But these constants tie into bigger design changes so they are really just the toes on the leg that needs surgery; that's why I keep suggesting you consider other areas of work for now.
```

The design issue I keep speaking about (which was better resolved in the SSU refactor) has *nothing* to do with enum struct: `&anonimal       That type of math for i2np should not be implemented there.`

We should merge the TODO resolution though so please see my original comment/request.

olarks:
I realize that and I understand what you are getting at, I have looked at the SSU refactor. I'll fix this PR so it only contains the namespace refactor. Thanks.

olarks:
anonimal: Only the namespace refactor now ;)

anonimal:
@olarks please rebase against new HEAD

Issue #507, opened 2016-12-13
Improve HTTP proxy


anonimal:
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

*Place an X inside the bracket to confirm*
- [X] I confirm.

---

General catch-all to cover @guzzijones' current work on the proxy. This includes his refactoring/features/fixes. Related to #11, references #340.

anonimal:
Resolved in #528. Thank you @guzzijones 🎉 👍

Issue #11, opened 2015-11-17
Return correct error codes in HTTPProxy.


anonimal:
Moved here from TODO.

anonimal:
Not just error codes; HTTProxy barely works. Needs full review and improvement. Removing as minor, labeling as medium.

anonimal:
Easy work. Un-assigning myself in case someone else wants to grab this issue before I get to it.

anonimal:
#340

guzzijones:
I did a bare bones fix to add the correct error codes.
It is sitting in my fixhttperrorfixex branch.
I will do some testing early this week and then a pull request.
Anonimal wanted a fix for the error codes minus unit testing and other http proxy server fixes.
I agreed that it would be good to at least have a fix for now.

anonimal:
Great start @guzzijones, keep it up! Also, please continue to reference #11 like you did in 6abd8a7, thanks.

anonimal:
@guzzijones I've opened #507 so you can reference/close that issue when working on non-error code HTTP proxy work. We'll keep this issue solely for error codes and we can close it once we merge your error code branch.

PR #528, opened 2017-01-18
HTTP Proxy : see #11 

guzzijones:

- [x ] I confirm.

---

### DIFF
--- a/src/client/proxy/http.cc
+++ b/src/client/proxy/http.cc
@@ -32,16 +32,22 @@
 
 #include "client/proxy/http.h"
 
+#include <atomic>
+#include <boost/algorithm/string.hpp>
+#include <boost/algorithm/string/regex.hpp>
+
 #include <boost/lexical_cast.hpp>
 #include <boost/regex.hpp>
-
-#include <atomic>
+#include <boost/tokenizer.hpp>
 #include <cassert>
 #include <cstdint>
 #include <cstring>
+#include <map>
 #include <memory>
 #include <string>
+#include <utility>
 
+#include <boost/foreach.hpp>
 #include "client/api/streaming.h"
 #include "client/context.h"
 #include "client/destination.h"
@@ -51,13 +57,8 @@
 #include "core/router/identity.h"
 
 #include "core/util/i2p_endian.h"
-
 namespace kovri {
 namespace client {
-
-//
-// Server
-//
 
 HTTPProxyServer::HTTPProxyServer(
     const std::string& name,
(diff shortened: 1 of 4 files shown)


tests/unit_tests/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 set(TESTS_CLIENT
   "client/address_book/impl.cc"
+  "client/proxy/http.cc"
   "client/reseed.cc"
+  "client/proxy/http.cc"
```
guzzijones: This must have been an artifact of git rebase -i

anonimal: Fix it, rebase, force push.

guzzijones:
I made changes to variables.  I had ran cpplint on files again.  I did a force push. Let me know.  I am not sure why it is showing merge conflicts now as i don't believe it was before..

anonimal:
You'll need to rebase against current master.

guzzijones:
I will adjust the logging to the new system.  That seems to be the reason for the merge conflicts.

anonimal:
Yes, like I said, just rebase against current master.

guzzijones:
I finished up the rebase against the current master.

guzzijones:
looks like the original implementation called async_receive method from the socket variable.  i am passing the socket as a variable.  That is what i can prove so far. :)  I will redo with that style of implementation.  Also i am going to run through a nice tutorial i found so i have good foundational footing.

guzzijones:
switch back to using async_receive is what i see as a possibility.

guzzijones:
i removed the offending header and rebased again.

src/client/proxy/http.cc
```diff
@@ -220,47 +274,71 @@ void HTTPProxyHandler::HandleStreamRequestComplete(
           stream);
     GetOwner()->AddHandler(connection);
     connection->I2PConnect(
-        reinterpret_cast<const std::uint8_t*>(m_Request.data()),
-        m_Request.size());
+        reinterpret_cast<const uint8_t*>(m_Protocol.m_Request.c_str()),
```
anonimal: Please use `std::size_t` and the standard library for all types when possible. Calling `size_t` directly is deprecated and the only reason this compiles now is because of deprecations elsewhere that have included `stddef.h`.

src/client/proxy/http.cc
```diff
@@ -220,47 +274,71 @@ void HTTPProxyHandler::HandleStreamRequestComplete(
...
+  auto it_user_agent = std::find_if(
+      m_HeaderMap.begin(),
+      m_HeaderMap.end(),
+      [](std::pair<std::string, std::string> arg) {
+        boost::trim_left(arg.first);
+        boost::trim_right(arg.first);
+        return arg.first == "User-Agent";
+      });
+  if (it_user_agent != m_HeaderMap.end())  //  found
+    it_user_agent->second = " MYOB/6.66 (AN/ON)";
```
anonimal: >" MYOB/6.66 (AN/ON)"

Watch out for the whitespace in strings, regardless of if they are trimmed or not.

src/client/proxy/http.cc
```diff
@@ -220,47 +274,71 @@ void HTTPProxyHandler::HandleStreamRequestComplete(
...
+    m_Request = m_Request + ii->first + ":" + ii->second + "\r\n";
+  }
+  m_Request = m_Request + "\r\n";
+  // concat body
+  m_Request += m_Body;
   return true;
 }
 
-bool HTTPProxyHandler::ExtractIncomingRequest() {
+bool HTTPProtocol::ExtractIncomingRequest() {
```
anonimal: >HTTPProtocol

I truly don't agree nor understand why you want to call this class `HTTPProtocol`. It's confusing and I don't see the point.

src/client/proxy/http.h
```diff
@@ -33,76 +33,23 @@
 #ifndef SRC_CLIENT_PROXY_HTTP_H_
 #define SRC_CLIENT_PROXY_HTTP_H_
 
+#include <boost/bind.hpp>
+#include <boost/optional.hpp>
 #include <boost/asio.hpp>
-
```
anonimal: Please don't remove lines like this unless absolutely needed.

src/client/proxy/http.h
```diff
@@ -210,73 +156,183 @@ class HTTPProxyHandler
...
+  /// @var m_JumpService
+  /// @brief Address helpers for base64 jump service
+  const std::array<std::string, 4> m_JumpService {
+  {
+    "?i2paddresshelper=",
+    "&i2paddresshelper=",
+    "?kovrijumpservice=",
+    "&kovrijumpservice=",
+  }
+  };
```
anonimal: Please don't mix style changes like this, it was fine the way it was.

src/client/proxy/http.h
```diff
@@ -210,73 +156,183 @@ class HTTPProxyHandler
...
 
-  /// @brief Parses path for base64 address, inserts into address book
-  void HandleJumpService();
+  /// @brief Processes original request: extracts, validates,
+  ///   calls jump service, appends original request
+  /// @return true on success
+  bool CreateHTTPRequest();
+
+  const unsigned int HEADERBODY_LEN = 2;
+  const unsigned int REQUESTLINE_HEADERS_MIN = 1;
```
anonimal: Please use `enum struct` or `enum` for these two.

src/client/proxy/http.h
```diff
@@ -210,73 +156,183 @@ class HTTPProxyHandler
...
+
+  ~HTTPProxyHandler() {
+    Terminate();
+  }
+
+  void set_result(boost::optional<boost::system::error_code> * a,
+      boost::system::error_code b ) {
+    if (b == 0)
+      a->reset(b);
+  }
```
anonimal: >set_result

?

tests/unit_tests/client/proxy/http.cc
```diff
@@ -0,0 +1,60 @@
...
+ */
+#define BOOST_TEST_DYN_LINK
+#include <boost/test/unit_test.hpp>
+#include "client/proxy/http.h"
+
+BOOST_AUTO_TEST_SUITE(HTTPPProtocolTests)
+
+BOOST_AUTO_TEST_CASE(Short) {
+  kovri::client::HTTPProtocol tmp;
+  std::string tmpData = "GET anonimal.i2p http/1.1";
```
anonimal: I use my garlicsite as a reference-point to show that I've been in a certain area. Certainly you don't want me to take credit for your work? Why not `guzzi.i2p`?

Also, doesn't the RFC specify uppercase HTTP?

tests/unit_tests/client/proxy/http.cc
```diff
@@ -0,0 +1,60 @@
...
+  std::string tmpData = "\r\n";
+  BOOST_CHECK(!tmp.HandleData(tmpData));
+}
+BOOST_AUTO_TEST_CASE(ok) {
+  kovri::client::HTTPProtocol tmp;
+  std::string tmpData = "GET anonimal.i2p ";
+  tmpData+="http/1.1\r\nUser-Agent: dummy\r\n\r\n";
+  BOOST_CHECK(tmp.HandleData(tmpData));
+}
+
```
anonimal: You *could* stand to make these tests more robust. I can provide details on IRC.

src/client/proxy/http.cc
```diff
@@ -32,178 +32,232 @@
...
-    }
-    buf++;
-    len--;
-    if (m_State == static_cast<std::size_t>(State::done))
-      return CreateHTTPRequest(buf, len);
+void HTTPProxyHandler::HandleSockRecv(
+    const boost::system::error_code& error,
+    std::size_t bytes_transferred) {
+  if (error) {
+    LOG(debug) << "HTTPSockRecv: error sock read: " << bytes_transferred;
```
anonimal: See above about logging class name/function name.

src/client/proxy/http.cc
```diff
@@ -220,47 +274,71 @@ void HTTPProxyHandler::HandleStreamRequestComplete(
...
 }
 
-bool HTTPProxyHandler::ExtractIncomingRequest() {
+bool HTTPProtocol::ExtractIncomingRequest() {
+  m_ErrorResponse = HTTPResponse(HTTPResponseCodes::status_t::bad_request);
   LOG(debug)
     << "HTTPProxyHandler: method is: " << m_Method
     << " request is: " << m_URL;
-  // Set defaults and regexp
+
```
anonimal: This isn't needed. Please try to be more concise with your commits.

guzzijones:
Thanks for the feedback. All good stuff. I will get on it this weekend.

Enviado con AquaMail para Android http://www.aqua-mail.com


En 27 de enero de 2017 10:29:00 PM 0x914409F1 <notifications@github.com> 
escribió:

src/client/proxy/http.cc
```diff
@@ -32,178 +32,232 @@
...
     std::shared_ptr<boost::asio::ip::tcp::socket> socket) {
   return std::make_shared<HTTPProxyHandler>(this, socket);
 }
 
-//
-// Handler
-//
-
-void HTTPProxyHandler::AsyncSockRead() {
+void HTTPProxyHandler::Handle() {
```
guzzijones: This was not changed.  Both of these functions still exist.

anonimal: I had no where else to comment, this comment was for your handling implementation as it relates to I2PService (and your handling implementation).

src/client/proxy/http.cc
guzzijones: HTTPProxyHandler still exists.  HTTPProtocol has been change to HTTPMessage.  HTTPMessage handles the specifics for http.   This class will continue to be used to decouple the specifics of HTTP from the HTTPHandler.  Let the Handler handle with the appropriate HTTPMessage.  Eventually there will be an HTTPMessage for diff versions of HTTP.  ie the older versions of HTTP handle the body length differently.  At least that was the idea.  Also pulling the functions into HTTPMessage allows me to do unit testing easily.

src/client/proxy/http.h
guzzijones: an enum would imply that they should have mutually exclusive values if they are part of the same enum.  I will do two separate enums inside one struct

guzzijones: getting an error trying to put two enums inside a struct.  I will wait for feedback on this one.

anonimal: >an enum would imply that they should have mutually exclusive values if they are part of the same enum.

```cpp
// TODO(guzzi): doxygen documentation
enum HeaderDefaults : std::uint8_t {
  BodyLength = 2,
  RequestLineMin = 1,
};
```
or similar. Personally, I'd rather see a stronger use-case for these constants - but this would require a different approach to the implementation (that can come later).

>I will do two separate enums inside one struct

Please don't, that's very unnecessary.

src/client/proxy/http.cc
```diff
@@ -144,7 +148,14 @@ void HTTPProxyHandler::AsyncHandleReadHeaders(
         boost::asio::buffers_end(buf));
     // add the additional bytes the m_body
     m_Protocol.m_Body = str;
-  }
+  } else if (num_additional_bytes < 0) {
```
anonimal: `num_additional_bytes` is *unsigned*. Research the difference between signed and unsigned types (and overflow + underflow for both types) to learn why this line makes no sense. Even if this type was signed (but it's not so don't get caught up with what I'm about to say), you don't want to test this way; you'll usually want to throw a class from std::exception.

src/client/proxy/http.cc
```diff
@@ -130,11 +133,12 @@ void HTTPProxyHandler::AsyncHandleReadHeaders(
...
 
   if (!m_Protocol.HandleData(tbuffer)) {
-    LOG(debug) << "AsyncHandleRead: error HandleData() " <<  "check http proxy";
+    LOG(debug) << "HTTPProxy: error HandleData() " <<  "check http proxy";
     HTTPRequestFailed();  // calls Terminate
     return;
   }
-  size_t num_additional_bytes = m_Protocol.m_Buffer.size() - bytes_transfered;
+  std::size_t num_additional_bytes = m_Protocol.m_Buffer.size()
+    - bytes_transfered;
```
anonimal: Still doesn't prevent overflow if bytes transferred is greater than buffer size. Ensure that this will never happen.

src/client/proxy/http.cc
```diff
@@ -154,12 +165,19 @@ void HTTPProxyHandler::AsyncHandleReadHeaders(
...
   if (it != m_Protocol.m_HeaderMap.end()) {
+    // body found
     std::istringstream iss(boost::trim_left_copy(it->second));
-    size_t clen;
+    std::size_t clen;
     iss >> clen;
     clen = clen - num_additional_bytes;
-    // need to call one more function to fill body after this read.
-    if (clen > 0) {
+    if (clen < 0) {
```
anonimal: See my comment above.

src/client/proxy/http.cc
guzzijones: i understand.

src/client/proxy/http.cc
guzzijones: i understand.

src/client/proxy/http.cc
guzzijones: i understand.

src/client/proxy/http.h
guzzijones: I don't really see the use case here for an enum.  The values could just as well be 
headerbodylength = 2
requestlinemin = 2.
etc. etc.
An enum wouldn't work in that case.  I am just trying to avoid magic numbers.

anonimal: >I don't really see the use case

Logical organization which compliments design. In this case, the design is lacking (why you use arbitrary constants which are loosely related but not derived from a larger mathematical constant or why you choose to measure and implement data relationships haphazardly) so, I think learning organization is in your best interest.

These are some of the real issues to consider (not whether it's enum or not) but I need to digress until you're ready to listen and learn. @olarks did similar refactoring work (very well might I add), so he might also have some insight.

For now, let's just get this PR done with.

>I am just trying to avoid magic numbers.
?

guzzijones:
I will go through tonight and add todos. I will go through tonight and run cpplint and test implemetation again.  This push was for review.  (others work during the day time). :)

anonimal:
As discussed with @guzzijones on IRC, even though there are remaining issues to resolve (and he will resolve them with time), this PR will fix our headers issue and implements POST at the same time - so we can merge this now while he continues his TODO work.

anonimal:
Resolves #507.

PR #536, opened 2017-01-24
HTTP: strengthen OpenSSL cipher list and options

anonimal:

- [X] I confirm.

---


anonimal:
All 9/10 reseed servers are confirmed functional with this update; the only exception being the 10th server (https://download.xxlspeed.com) which appears to be having problems - even without this update.

This is not an issue that should prevent us from merging as we technically only need 1 working reseed server.

anonimal:
https://download.xxlspeed.com/ is online again. 10/10.

PR #541, opened 2017-01-30
Introduce Dockerfile

chris-barry:

- [X] I confirm.

---

Introduce a Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+# kovri - I2P but in CPP
+# Copyright (c) 2015-2017, The Kovri I2P Router Project (see LICENSE.md)
+# docker build -t geti2p/kovri .
+# KOVRI_PORT=42085 && sudo docker run -p $KOVRI_PORT --env KOVRI_PORT=$KOVRI_PORT geti2p/kovri
+
+FROM alpine:3.5
+RUN apk add --update --no-cache -t .kovri-dev g++ make cmake binutils boost-dev libressl-dev
+ENV KOVRI_HOST 0.0.0.0
+ENV KOVRI_PORT 24085
+ADD . kovri/
+RUN cd kovri && \
+    make all tests
+RUN adduser -D kovri && \
+    mkdir /home/kovri/.kovri && \
+    /kovri/build/kovri-tests && \
+    cp -r /kovri/pkg/* /home/kovri/.kovri
+RUN install /kovri/build/kovri-tests /usr/bin/kovri-tests && \
+    install /kovri/build/kovri /usr/bin/kovri && \
+    cd .. && \
+    rm -rf ~/kovri && \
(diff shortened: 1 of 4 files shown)


chris-barry:
Please hold merging this. Had an IRC discussion with anonimal. I need to do the following:

1. Check if `Dockerfile` can go in `contrib/`
2. Add notes to `doc/BUILDING.md` or `doc/USER_GUIDE.md`

Dockerfile
```diff
@@ -0,0 +1,25 @@
+# kovri - I2P but in CPP
+# Copyright (c) 2015-2017, The Kovri I2P Router Project (see LICENSE.md)
+# docker build -t geti2p/kovri .
+# KOVRI_PORT=42085 && sudo docker run -p $KOVRI_PORT --env KOVRI_PORT=$KOVRI_PORT geti2p/kovri
+
+FROM alpine:3.5
+RUN apk add --update --no-cache -t .kovri-dev g++ make cmake binutils boost-dev libressl-dev
+ENV KOVRI_HOST 0.0.0.0
+ENV KOVRI_PORT 24085
+ADD . kovri/
```
MoroccanMalinois: IMHO, i think it's better to "git clone" instead of "ADD" (like in monero) to have a self-consistent dockerfile

chris-barry: I am not a Docker guru. What are the pros of that as opposed to adding the source?

MoroccanMalinois: Neither am i :) but i use it a lot. Major point is self-consistency : no additional step. It's supposed to describe the entire process. By using "ADD", somehow you are making a step less explicit. Also easier to deploy to dockerhub : no need to maintain a local copy of the source there.

chris-barry: I don't have strong feeling either way. I am going to do some research and see if there is a docker-community consensus. :)

Right now for dockerhub I have it pointed towards a repo under my account, but you have a good point.

I'll do some changes tomorrow-ish.

chris-barry: This[1] is the best resource I could find on strategies on getting code in to Docker containers. I'm going to add some stuff to the .dockerignore and leave it as is for now. Mainly because I know it works.

[1]: https://ryanfb.github.io/etc/2015/07/29/git_strategies_for_docker.html

chris-barry:
Force pushed to my branch with additions to the documentation.

chris-barry:
@anonimal The short story of what I found is, it depends if you want automated builds on dockerhub (probably a good thing). The automated builds can scan git and make automatic builds based on tags.

Unfortunately dockerhub does not allow you to build with a `Dockerfile` in a non-root directory.

If automated builds are not a priority, then it can be stored anywhere.

anonimal:
>automated builds on dockerhub

Sounds like a good decision.

Issue #538, opened 2017-01-25
NetDb: HandleDatabaseStoreMsg: 'GunzipImpl: Put caught exception 'Inflator: error in compressed block''


anonimal:
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

*Place an X inside the bracket to confirm*
- [X] I confirm.

---

Roughly ~5 days ago (~2 days after writing [this comment](https://trac.i2p2.de/ticket/1915#comment:8) in java I2P's trac), all test routers began segfaulting with similar (gunzip) related errors (for us, at the core level).  I've never seen this issue before, and we've been enjoying weeks/months+ of segfault-free runtime. Logs + backtraces show that this is a new issue:

```
[2017.01.24 22:58:51.069654] [0x00007fffebfff700] [debug]  NetDb: DatabaseStore
[2017.01.24 22:58:51.069718] [0x00007fffebfff700] [debug]  NetDb: RouterInfo
[2017.01.24 22:58:51.069851] [0x00007fffebfff700] [error]  GunzipImpl: Put caught exception 'Inflator: error in compressed block'
[2017.01.24 22:58:51.069958] [0x00007fffebfff700] [error]  IdentityEx: identity buffer length 0 is too small
```
```
[2017.01.24 22:14:46.072220] [0x0000000805008000] [debug]  NetDb: DatabaseStore
[2017.01.24 22:14:46.072240] [0x0000000805008000] [debug]  NetDb: RouterInfo
[2017.01.24 22:14:46.073597] [0x0000000805008000] [error]  GunzipImpl: Put caught exception 'Inflator: error in compressed block'
[2017.01.24 22:14:46.073631] [0x0000000805008000] [debug]  NetDb: new RouterInfo added
[2017.01.24 22:14:46.073648] [0x0000000805008000] [error]  IdentityEx: identity buffer length 0 is too small
```
```
[2017.01.22 20:54:07.787485] [0xb42fab40] [error]  GunzipImpl: Put caught exception 'Gunzip: CRC check error'
[2017.01.22 20:54:07.787554] [0xb42fab40] [error]  IdentityEx: identity buffer length 1 is too small
[...]

Issue #533, opened 2017-01-24
Log: move global logger storage to logger object + enable log-to-file in daemon mode


anonimal:
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

*Place an X inside the bracket to confirm*
- [X] I confirm.

---

As noted in c8870c0905cf79b35a4d9fd667a5e64af3cad45c
```
// TODO(anonimal):
// Boost.Log uses an "application-wide singleton" (note: our logger/sink setup applies globally from instance configuration)
// As a result, logging will not work when in daemon mode. http://www.boost.org/doc/libs/1_63_0/libs/log/doc/html/log/rationale/fork_support.html
// We've worked around this problem in the past by using some very gross hacking but we may be able to apply a cleaner work-around so we can set this up entirely in the core namespace
// (we could create an (inheritable?) logging class with overloaded stream operator and adjust our logging initialization and macro accordingly, or consider other options)
// Also note that a singleton will effect having multiple logging library options (there's no need to do that though when we have huge flexibility with sinks)
```

While we here (unrelated), we should add tags and a global exception handler (if this is needed sooner then we should open a new feature request).

coneiric:
> (we could create an (inheritable?) logging class with overloaded stream operator and adjust our logging initialization and macro accordingly, or consider other options)

One idea I had along these lines was to create a base logging class, like you suggest, and inherit+instantiate in each module namespace.

So each nampespace's macro wouldn't need to change when calling it, and would get initialized with that module's logger.

How I'm thinking of it so far:

In `core/util/base_logger.h/cc`:
```
all the current definitions in core/util/log.h/cc, without the macros
```

Then in `core/util/log.h/cc`:
```
namespace kovri {
namespace core {

class CoreLogger : public BaseLogger {...};
CoreLogger co_Logger;
#define LOG(severity)  BOOST_LOG_SEV(co_Logger, boost::log::trivial::severity)
```

Would take a similar approach for client, app, and util. Does this design make sense,  or follow your suggestions?

Also, thought about giving each class it's own logger, but that seemed to get out of control quickly (already over 64 classes that use logging).

[...]

PR #544, opened 2017-02-06
Create exception dispatcher utility class + implement for crypto impl

anonimal:

- [X] I confirm.

---

As noted in the TODO's, most of this refactor is immediately ineffective
if the exception isn't handled at a higher calling level *but*, when/if
things are thrown and *not* caught, this refactor should get us started
in the right direction in terms of debugging and subsequent higher-level
re-design.

As I noted in yesterday's meeting:

"JFTR: of the thousands of lines of code that we forked from, almost
none of it handled exceptions. All this spaghetti code, and that guy
did not even have a basic grasp of design with exception handling in
mind. Literally, much of the core code is still stringed together with
the hope that it won't fail. I hate it. It's always been like walking
on glass, but a real fix will take time. So, I've started in that
direction at least in terms of getting exceptions handled or handled
better."

Also note the TODO's for the exception class: this work is a WIP so I do not want to refactor the remaining code-base with this class until it's solidified (and after the logging implementation is solidified as it will handle class name attribute) (#533).

Note: this refactor still catches #538

Note: FYI, I'll most likely force push this branch after it's rebased against my ntcp branch or ntcp branch is merged before this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,6 +188,7 @@ if(WITH_CRYPTOPP)
   else()
     message(STATUS "Found Crypto++: ${CryptoPP_INCLUDE_DIR}, ${CryptoPP_LIBRARIES}")
     include_directories(${CryptoPP_INCLUDE_DIR}/../)  # We back-track for <cryptopp/*>
+    add_definitions(-DWITH_CRYPTOPP)
   endif()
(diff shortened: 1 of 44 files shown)


anonimal:
Live-tests pass on all platforms.

PR #548, opened 2017-02-08
Add the packaging metadata to build the kovri snap

elopio:

- [X] I confirm.

---

This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds that then can be distributed to many users through the Ubuntu store.

Kudos to @anonimal for making the snapcraft.yaml. I only added a few touches and sent the PR for some stolen karma :)

### DIFF
--- a/snap/kovri-wrapper
+++ b/snap/kovri-wrapper
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+if [ ! -d "$SNAP_USER_DATA/.kovri" ]; then
+  mkdir $SNAP_USER_DATA/.kovri/
+  cp -R $SNAP/config $SNAP_USER_DATA/.kovri/
+  cp -R $SNAP/client $SNAP_USER_DATA/.kovri/
+fi
+
+exec "$SNAP/bin/kovri" "$@"
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,28 @@
+name: kovri
+version: master
+summary: The Kovri I2P Router Project
+description: |
+  Kovri is a security-focused, community-driven C++ implementation of I2P
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: strict
+
+apps:
+  kovri:
+    command: kovri
+    plugs: [network, network-bind]
+
+parts:
+  kovri:
+    source: .
+    plugin: make
+    build-packages:
+      - g++
+      - libboost-all-dev
+      - libssl-dev
+      - libssl1.0.0
+    artifacts:
+      - build/kovri
+      - pkg
+    organize:
+      build/kovri: bin/
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -9,7 +9,7 @@ confinement: strict
 
 apps:
   kovri:
-    command: kovri
+    command: kovri-wrapper
     plugs: [network, network-bind]
 
 parts:
@@ -23,6 +23,12 @@ parts:
       - libssl1.0.0
     artifacts:
       - build/kovri
-      - pkg
+      - pkg/config
+      - pkg/client
     organize:
       build/kovri: bin/
+      pkg/*: .
+  wrapper:
+    plugin: dump
+    source: snap/
+    stage: [kovri-wrapper]


elopio:
You can find more information about snapcraft here: http://snapcraft.io/

To test it in an Ubuntu 16.04 machine:

```
$ sudo apt install git snapcraft
$ git clone https://github.com/elopio/kovri
$ cd kovri
$ git checkout snapcraft
$ snapcraft
$ sudo snap install *.snap --dangerous
```

There are a few things to discuss here before the snap is ready to be published.

I'm not sure where do you want to put the config data. Currently, you would have to call kovri like this:
`$ kovri -c /snap/kovri/current/pkg/config/kovri.conf -t /snap/kovri/current/pkg/config/tunnels.conf`

And it still fails with:
`No such file or directory: "/home/ubuntu/snap/kovri/x1/.kovri/client/certificates/su3"`

There doesn't seem to be a flag to specify the certificates path.

anonimal:
@elopio thank you for doing this!

>And it still fails with:
>>No such file or directory: "/home/ubuntu/snap/kovri/x1/.kovri/client/certificates/su3"

>There doesn't seem to be a flag to specify the certificates path.

snapcraft needs to somehow run `make install-resources`. We currently don't use `make install` because we've wanted to refrain from using system directories until we were closer to release - but we're getting closer so now is a good time for us to make that change.

If running `make install-resources` doesn't solve the issue, then possibly the following:

Currently, `KOVRI_DATA_PATH` is a hardcoded build option which means the binary will need *all* non-config `pkg/`resources (and run-time files) to be in said path - regardless of the config file switches (the current default hardcoded data path on Linux is `~/.kovri/`).

The ideal solution is to simply implement an optional data directory switch but what we can easily do for now, before building, is export an environment variable `KOVRI_DATA_PATH="~/snap/kovri/x1/.kovri"` (or whatever directory is used for snap's filesystem).

Once #549 is merged, I can also add a `make snapcraft` target which will add this definition automatically. Or, if there's a way for snapcraft to export env variables before building then that should work too.

elopio:
@anonimal when the snap is run, it has a slightly modified environment to properly work in confinement. So from the point of view of the kovri binary, $HOME is ~/snap/kovri/{revision}, which is readable and writable only by kovri. I think that part works fine here.

The problem we are hitting is that during install, snaps are mounted read-only in /snap/{name}/current. This is by design, it's secure, allows for delta downloads and simple rollbacks when the install fails. So the resources and default configs are bundled inside the snap, and during the install they are mounted somewhere in /snap/kovri/current/. We need to move them from this path to $HOME.

As the configure hook runs during install, I think we can make a simple script that does something like:

```
if [ ! -d "$SNAP_USER_DATA/.kovri/config" ]; then
    cp -R $SNAP/config $SNAP_USER_DATA/.kovri/config
fi
```

And the same for the client certificates.

Note that this also has the potential to work for migrations. If ever you need to change the config format, or regenerate the certificates, or something, you could make the script smarter. Check if old config exists, and apply a transformation.

Let me know what do you think.

anonimal:
Ok. What if:
```bash
if [ ! -d "$SNAP_USER_DATA/.kovri" ]; then
    cp -R $SNAP $SNAP_USER_DATA/.kovri
fi
```
and take out the config file switches from the command line? All resources (by default) are read from `.kovri`.  Side-note: will kovri see `$SNAP_USER_DATA` as `$HOME/snap/kovri/{revision}`? I've read https://snapcraft.io/docs/reference/env but am not sure if this is the case.

>If ever you need to change the config format, or regenerate the certificates, or something, you could make the script smarter. Check if old config exists, and apply a transformation.

This happens frequently. Do we need to apply any transformations or can we simply overwrite the files (or will that not be an issue because of snapshots)?

elopio:
@anonimal yes, with the hook copying the config to the home, there will be no need to pass the values when calling kovri from the command line. And yes, $SNAP_USER_DATA is /home/{user}/snap/kovri/{revision}.

Also, the hook script is yours to do anything you want. You can just overwrite, no problem with that. I thought that users might have tweaked the default config, so if you just overwrite, those changes will be lost. Or well, not lost because the old snap is always kept in case the user wants to rollback. But note that the flexibility of the hook lets you figure that out later. You can always push a revision that the only thing that changes is the place where the config is stored, or something like that.

anonimal:
>Or well, not lost because the old snap is always kept in case the user wants to rollback. But note that the flexibility of the hook lets you figure that out later.

Ah, ok. I'll need to RTFM, thank you for clarifying.

Forgive my ignorance, but where/how do we implement this? With https://snapcraft.io/docs/build-snaps/hooks or https://snapcraft.io/docs/build-snaps/scriptlets?

elopio:
@anonimal the scriptlets run at build time, so they can take care of anything that needs to live inside the snap. The hooks run at install time, so they can take care of things that need to live in $HOME, outside of the snap, and thus writable.

elopio:
@anonimal sorry it took so long, I got stuck on Friday. Take a look at the solution suggested by @kyrofa, and please let me know if it works for you.

anonimal:
@elopio awesome! It works! Thank you so much!!!!!!! 🎉 👍

elopio:
Thanks for merging @anonimal.
I think that a useful next step is to push the snap to the edge channel in the store:
https://snapcraft.io/docs/build-snaps/publish

That way, both the community of testers from monero and Ubuntu can start trying it and provide feedback. I'll be around again on monday, send me a ping if I can help you with something else.

anonimal:
Ok, sounds great. I'll need work out details internally with @danrmiller and @fluffypony but I'll give you a ping once that's resolved.

PR #578, opened 2017-02-25
http: const getters + split HostIsI2P and CompleteURI

MoroccanMalinois:

- [X] I confirm.

---

That has nothing to do with what i am doing currently, so dunno if it's appropriate to PR this.

### DIFF
--- a/src/client/util/http.cc
+++ b/src/client/util/http.cc
@@ -67,17 +67,27 @@ bool HTTP::Download() {
   }
   // TODO(anonimal): ideally, we simply swapout the request/response handler
   // with cpp-netlib so we don't need two separate functions
-  if (!HostIsI2P())
-    return DownloadViaClearnet();
-  return DownloadViaI2P();
+  if (HostIsI2P())
+    {
+      AmendURI();
+      return DownloadViaI2P();
+    }
+  return DownloadViaClearnet();
 }
 
-bool HTTP::HostIsI2P() {
+bool HTTP::HostIsI2P() const
+{
   auto uri = GetURI();
   if (!(uri.host().substr(uri.host().size() - 4) == ".i2p"))
     return false;
+  return true;
+}
+
+void HTTP::AmendURI()
+{
+  auto uri = GetURI();
   if (!uri.port().empty())
-    return true;
+    return;
   // We must assign a port if none was assigned (for internal reasons)
   std::string port;
   if (uri.scheme() == "https")
@@ -94,7 +104,6 @@ bool HTTP::HostIsI2P() {
       + uri.host() + ":" + port
       + uri.path() + uri.query() + uri.fragment());
   SetURI(new_uri);
-  return true;
 }
 
 bool HTTP::DownloadViaClearnet() {
--- a/src/client/util/http.cc
+++ b/src/client/util/http.cc
@@ -46,6 +46,7 @@
 
 #include "core/router/context.h"
 
+#include "core/util/byte_stream.h"
 #include "core/util/filesystem.h"
 #include "core/util/log.h"
 
@@ -89,6 +90,7 @@ void HTTP::AmendURI()
   if (!uri.port().empty())
     return;
   // We must assign a port if none was assigned (for internal reasons)
+  LOG(trace) << "HTTP : Amending URI";
   std::string port;
   if (uri.scheme() == "https")
     port.assign("443");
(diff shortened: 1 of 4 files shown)


anonimal:
>so dunno if it's appropriate to PR this.

No problem! It's open source and you're not under any contract to complete any specific work. I'm glad to see you're exploring the codebase.

src/client/util/http.h
```diff
@@ -172,15 +176,19 @@ class HTTP : public HTTPStorage {
...
-  /// @brief Tests if TLD is I2P and set default ports for in-net downloading
+  /// @brief Tests if TLD is I2P
   /// @return True if TLD is .i2p
+  bool HostIsI2P() const;
+
+  /// @brief Set default ports for in-net downloading
   /// @notes The default port is 80, for HTTPS it is 443
   /// @notes Removing this will require refactoring stream implementation
-  bool HostIsI2P();
+  void CompleteURI();
```
anonimal: >CompleteURI();

I like how you want to decouple host type check from the amending of the URI. I must point out though that the only use-case we have for amending the URI is for in-net downloading and the only reason we need to amend the URI for in-net downloading is because the cpp-netlib refactor isn't finished. Also, adding this function implies that there is a use-case to amend the URI elsewhere - which there isn't.

If we want to decouple this, maybe we should also change the logic test for host type (e.g., `if (HostIsI2P()) { AmendURI(); return DownloadViaI2P(); }`. Ultimately, I think the best solution is for the refactor to be finished but I think that would be fine for now.

Did you have any other thoughts on this area?

MoroccanMalinois:
> Did you have any other thoughts on this area?

Only one thing that is neither in this PR nor in a TODO(i think) : if the flag `--reseed-skip-ssl-check` is activated, all calls to HTTP::Download with a clearnet url will skip ssl checking. Looks like today this is only used for reseed (assuming publishers are part of the reseed process), so it is not a current pb, but if it's meant to also be used for other purposes, it might be misleading. 

(will add more log DownloadViaI2P() when i'll get to take a deep look into it)

src/client/util/http.cc
```diff
@@ -135,6 +139,10 @@ bool HTTP::DownloadViaClearnet() {
           | SSL_OP_NO_TLSv1 | SSL_OP_NO_TLSv1_1
           | SSL_OP_NO_SSLv2 | SSL_OP_NO_SSLv3);
   }
+  else
+    LOG(debug) << "HTTP: Skipped Reseed SSL checking ";
```
anonimal: FYI: because the if block has braces, else should also have braces; even if one line (google style thing).

src/client/util/http.cc
```diff
@@ -111,6 +112,8 @@ bool HTTP::DownloadViaClearnet() {
   // Create and set options
   Options options;
   options.timeout(static_cast<std::uint8_t>(Timeout::Request));
+  LOG(debug) << "HTTP: Download Clearnet with timeout : "
+             << std::to_string(static_cast<std::uint8_t>(Timeout::Request));
```
anonimal: Some alternatives:

- `#include "core/util/byte_stream.h"` `kovri::core::GetType(Timeout::Request)`
- Refactor `Timeout` as it currently doesn't need to be scoped *if* refactored as a member of class `HTTP`, then simply use `Timeout::Request`

src/core/util/log.cc
```diff
@@ -46,8 +48,82 @@ BOOST_LOG_GLOBAL_LOGGER_INIT(g_Logger, boost::log::sources::severity_logger_mt)
...
+     << (opt.openssl_ciphers() ? *opt.openssl_ciphers() : std::string("No"))
+     << ", openssl_sni_hostname ? "
+     << (opt.openssl_sni_hostname() ? *opt.openssl_sni_hostname()
+                                    : std::string("No"))
+     << ", always_verify_peer ? " << opt.always_verify_peer()
+     << ", options : " << opt.openssl_options()
+     << ", timeout : " << opt.timeout() << ", io_service ? "
+     << (opt.io_service() ? "Yes" : "No");
+  return ss.str();
+}
```
anonimal: This is a lot of manual work; especially if we add/remove options. I'm not sure how useful logging these options are right now but if this is what you want then ok.

As a whole, I think we should also consider a more unified system of formatting. All these dangling `?` and `<>` and `:` in terms of English, I find quite confusing.

PR #579, opened 2017-02-25
Add consts in various places where it makes sense

selene-kovri:

- [ ] I confirm.

---

### DIFF
--- a/src/app/config.cc
+++ b/src/app/config.cc
@@ -159,8 +159,8 @@ void Configuration::ParseKovriConfig() {
 
 // TODO(unassigned): improve this function and use-case
 void Configuration::ParseKovriConfigFile(
-    std::string& file,
-    bpo::options_description& options,
+    const std::string& file,
+    const bpo::options_description& options,
     bpo::variables_map& var_map) {
   std::ifstream filename(file.c_str());
   if (!filename)
@@ -330,7 +330,7 @@ void Configuration::ParseTunnelsConfig() {
   }
 }
 
-const std::string Configuration::GetAttribute(Key key) {
+const std::string Configuration::GetAttribute(Key key) const {
   switch (key) {
     // Section types
     case Key::Type:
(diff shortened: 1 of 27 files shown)


selene-kovri:
Understood, I will work more on this to use the const versions of those two.

anonimal:
Hi @selene-kovri, how is this coming along? Running into any snags?

anonimal:
@selene-kovri I'm going to merge this now. If the dead code lingers we'll just remove it or refactor. You're still welcome to PR new/continuing work though.

selene-kovri:
Thanks, and sorry I've not been able to work on this yet. Not much time :)

selene-kovri:
I have looked some more, and I need an opinion on which path to take.
In order to use the const GetAddress/GetAddressBook functions, CheckAddressIdentHashFound and GetLoadedAddressIdentHash ought to be const. However, GetLoadedAddressIdentHash calls LoadSubscriptionFromPublisher, which is plainly non-const, even though the first two are logically const, and the loading does lazy loading of data.
One way to do this is to have a new function EnsureSubscriptionFromPublisherLoaded (which would be const) and make the loaded data mutable (I've not looked yet at exactly what would need to be mutable). I'm asking because having mutable data might be something that's unwanted (there's very little mutable use in the source).

anonimal:
>GetLoadedAddressIdentHash calls LoadSubscriptionFromPublisher

Ideally, `LoadSubscriptionFromPublisher` should not be called from `GetLoadedAddressIdentHash` because the checking of ident hash should not rely on the ensuring of a subscription being loaded - as that is done within the addressbook's `SubscriberUpdateTimer`. What we do now is a hack to work around the original hacked design that we forked from - but - what *we* do will ensure that a subscription is loaded *before* the timer begins (because client tunnels are brought up on startup).

>I'm asking because having mutable data might be something that's unwanted (there's very little mutable use in the source).

I think the issue is endemic of design and could be resolved by decoupling the aforementioned *without* the need for enforcing `mutable` members (this decoupling work was put on the backburner but you're more than welcome to work on it).

Regarding both points: there are more chicken-or-the-egg design issues, one more being subscription download as done in `LoadSubscriptionFromPublisher`. If needed, I can give details in IRC if you decide to move forward on any of these issues.

Issue #563, opened 2017-02-16
Config: implement custom data directory


anonimal:
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

*Place an X inside the bracket to confirm*
- [X] I confirm.

---

The benefits are multitudinous; including allowing reads + writes to tmpfs/memory-only filesystem.

`KOVRI_DATA_PATH` is hardcoded for cross-platform compatibly. There is a custom data dir feature but only available at compile-time. I believe we can keep the hard-coded defaults but also allow a `--data-dir` run-time feature (note: other hard-coded areas will need refactoring).

In addition to this feature, we should also consider adding a quick note to the user-guide on how to use said virtual filesystems.

anonimal:
Resolved in #569

PR #580, opened 2017-03-03
kovri-util: add logs and data-dir options

MoroccanMalinois:

- [X] I confirm.

---
su3file need the certificates in data-dir and log options are necessary to be able "to pipe" the output.  

Also Referencing #563


anonimal:
Please rebase.

MoroccanMalinois:
Rebased

src/util/base.cc
```diff
@@ -104,18 +104,15 @@ bool process(
 
 bool BaseCommand::Impl(
     const std::string& cmd_name,
-    int argc,
```
guzzijones: Why the move away from argc and argv?  I now cannot parse out -a as an option for benchmark .  It thinks it is -all for kovri-util, when i run kovri-util benchark -a the help prints out for me and the -all is NOT run anyway.

guzzijones: I think the correct solution was to use all options AFTER the positional argument as options for the benchmark test. The rerun the bpo::command_line_parser. As it currently stands i wont be able to use any of the options for main --all, -a, etc.  I would rather put it back before I push.

MoroccanMalinois:
I replaced (argc, argv) with a vector. In fact, i changed the logic in main because otherwise the arguments managed in main had to be put first, which is not how most cli works. I think it's better now. But the effect is that we can't use the same flag (-a) in the main and in a sub command. I suggest you remove the option -a in the main, since it was not a terrible idea to begin with :) 

Sorry for the inconvenience.

guzzijones:
' i changed the logic in main because otherwise the arguments managed in main had to be put first,'  I actually thought that was great.  I am going to wait until anonimal chimes in on this.  I am waiting for a fix for the log level anyway.

MoroccanMalinois:
The fix for the log-level is simply changing 1 by 3 in https://github.com/monero-project/kovri/blob/master/src/util/main.cc#L70
you can also use "./build/kovri-util --log-level 3 ...."

PR #598, opened 2017-03-18
kovri-tests :: fix unmatched curly address book unit tests

guzzijones:
---

---

fix unmatched bracket in address book unit test

### DIFF
--- a/tests/unit_tests/client/address_book/impl.cc
+++ b/tests/unit_tests/client/address_book/impl.cc
@@ -41,7 +41,6 @@
 
 #include "core/crypto/rand.h"
 
-BOOST_AUTO_TEST_SUITE(AddressBook);
 
 /// @class SubscriptionFixture
 struct SubscriptionFixture {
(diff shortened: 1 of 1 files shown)


tuanveo:
Vào CN, 19 thg 3 2017 lúc 06:22 AJ <notifications@github.com> đã viết:

guzzijones:
BOOST_FIXTURE_TEST_SUITE replaces BOOST_AUTO_TEST_SUITE when used according to docs.  http://www.boost.org/doc/libs/1_59_0/libs/test/doc/html/boost_test/tests_organization/fixtures/case.html
"Once again the only difference from the macro BOOST_AUTO_TEST_SUITE usage is the presence of an extra argument - the fixture name."

and here is the very easy to follow tutorial i have been using:
http://www.alittlemadness.com/2009/03/31/c-unit-testing-with-boosttest/

anonimal:
>"Once again the only difference from the macro BOOST_AUTO_TEST_SUITE usage is the presence of an extra argument - the fixture name."

Because there's no code before the fixture (which requires the suite), this is doable. Not just because the macro merely has an "extra argument".

This should be good enough for now then.

Issue #96, opened 2016-01-29
Improve configuration file handling


anonimal:
So, 2429f3a in #61 gets some important work started with regard to updating the router context when reloading the tunnels config file.

The following issues still need to be tackled:

1. Config files aren't reloaded automatically. SIGHUP is still needed to reload files. To reload automatically would require a little more client->router design work.
2. tunnels config: after changing a port value for an existing tunnel and then pkilling with SIGHUP, the previously defined port is never closed. This should be implemented (not a bug since we don't claim to do it yet).
3. 2429f3a copy/pastes previous orginal code (e.g, unneeded redundancy between ```ReadTunnels()``` and ```ReloadTunnels()```) and has an overall orignal approach to implementation. [Boost.Program_options](http://www.boost.org/doc/libs/1_60_0/doc/html/program_options.html) can reduce the amount of code needed by a noticeable difference and it will better abstract the implementation (that's what it's designed for). [variables_map](http://www.boost.org/doc/libs/1_60_0/doc/html/program_options/reference.html#header.boost.program_options.variables_map_hpp) can easily map everything needed inside the tunnels config and can easily be cleared/refilled when reloaded. Code should be reused and our [current config handler](https://github.com/monero-project/kovri/blob/master/src/client/util/Config.cpp) (with a little work) can easily handle both kovri and tunnels configuration files. Once that's done, satisfying ```ReadTunnels()/ReloadTunnels()``` is trivial.
4. ```Daemon_Singleton::Reload()``` doesn't actually reload the daemon; it only reloads configuration file(s) (currently only tunnels config). When reloading kovri conf, changing port value for example would require **real** client->router design overhaul to implement a "graceful restart" (see [I2PControl](https://github.com/monero-project/kovri/blob/master/src/api/I2PControl/I2PControl.cpp) for an approach to router shutdown).  We need to do this (more client->router separation) anyway to make room for a future auto-update implementation (see #48). This last point may go in its own ticket once (at least) the previous 3 are resolved.

anonimal:
Assigning myself point 3 so I can finish that [point 3] along with #42.

anonimal:
I've unassigned myself because these various issues can be resolved when refactoring app and client (and each issue can be resolved separately by any number of developers).

anonimal:
#340

PR #609, opened 2017-03-28
Minor client/server tunnel refactor

MoroccanMalinois:
---

---
Minor changes, just preparation for a future discussion on IRC.
Referencing #96

### DIFF
--- a/src/app/instance.cc
+++ b/src/app/instance.cc
@@ -84,6 +84,7 @@ void Instance::Reload() {
   // TODO(unassigned): locking etc.
   // TODO(unassigned): client/router contexts
   m_IsReloading = true;
+  GetConfig().ParseTunnelsConfig();
   SetupTunnels();
   m_IsReloading = false;
 }
--- a/src/app/instance.cc
+++ b/src/app/instance.cc
@@ -207,13 +207,13 @@ void Instance::SetupTunnels() {
         else
           LOG(error) << "Instance: Failed to add server tunnel";
       }
-    } catch (const std::exception& ex) {
-      LOG(error) << "Instance: exception during tunnel setup: " << ex.what();
-      return;
-    } catch (...) {
-      LOG(error) << "Instance: unknown exception during tunnel setup";
-      return;
     }
+    catch (...)
+      {
+        kovri::core::Exception ex;
+        ex.Dispatch(__func__);
+        return;
+      }
   }  // end of iteration block
   if (m_IsReloading) {
     LOG(info) << "Instance: " << client_count << " client tunnels updated";
--- a/src/app/instance.cc
+++ b/src/app/instance.cc
@@ -165,8 +165,7 @@ void Instance::InitClientContext() {
 
 void Instance::SetupTunnels() {
   // List of tunnels that exist after update
-  // TODO(unassigned): ensure that default IRC and eepsite tunnels aren't removed?
-  std::vector<std::string> updated_tunnels;  // TODO(unassigned): this was never fully implemented
+  std::vector<std::string> updated_tunnels;
   // Count number of tunnels
   std::size_t client_count = 0, server_count = 0;
   // Iterate through each section in tunnels config
@@ -187,6 +186,7 @@ void Instance::SetupTunnels() {
             continue;
           }
           kovri::client::context.UpdateClientTunnel(tunnel);
+          updated_tunnels.push_back(tunnel.name);
           ++client_count;
           continue;
         }
@@ -200,6 +200,7 @@ void Instance::SetupTunnels() {
         bool is_http = (tunnel.type == GetConfig().GetAttribute(Key::HTTP));
         if (m_IsReloading) {
           kovri::client::context.UpdateServerTunnel(tunnel, is_http);
+          updated_tunnels.push_back(tunnel.name);
           ++server_count;
           continue;
         }
@@ -219,7 +220,6 @@ void Instance::SetupTunnels() {
   if (m_IsReloading) {
     LOG(info) << "Instance: " << client_count << " client tunnels updated";
     LOG(info) << "Instance: " << server_count << " server tunnels updated";
-    // TODO(unassigned): this was never fully implemented
     RemoveOldTunnels(updated_tunnels);
     return;
   }
--- a/src/app/instance.cc
+++ b/src/app/instance.cc
@@ -189,22 +189,12 @@ void Instance::SetupTunnels() {
           ++client_count;
           continue;
         }
-        // Get local destination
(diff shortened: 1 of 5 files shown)


src/client/context.cc
```diff
@@ -385,6 +371,50 @@ void ClientContext::UpdateClientTunnel(
...
+  // Insert client tunnel
+  auto client_tunnel = std::make_unique<kovri::client::I2PClientTunnel>(
+      tunnel, local_destination);
+  if (!InsertClientTunnel(tunnel.port, std::move(client_tunnel)))
+    {
+      LOG(error) << "Instance: failed to insert new client tunnel";
+      return false;
+    }
+  std::lock_guard<std::mutex> lock(m_ClientMutex);
+  m_ClientTunnels[tunnel.port]->Start();
```
MoroccanMalinois: This is the only logic change. ClientContext::AddClientTunnel is called from 2 places : one of them didn't have the Start call.  I don't know if it should be called, but i found it weird that they didn't do the same.

src/app/instance.cc
```diff
@@ -212,17 +186,67 @@ void Instance::SetupTunnels() {
...
+              kovri::client::context.UpdateServerTunnel(tunnel, is_http);
+              ++server_count;
+            }
+        }
+      catch (const std::exception& ex)
+        {
+          LOG(error) << "Instance: exception during tunnel setup: "
+                     << ex.what();
+          return;
+        }
```
anonimal: You can get rid of this block and implement the exception dispatcher in the next block.

MoroccanMalinois:
@anonimal : it's little better, still multiple todo's in this area but i think it's good enough for a first review.

src/app/instance.h
```diff
@@ -81,9 +81,7 @@ class Instance {
   void SetupTunnels(bool is_reloading = false);
 
   /// @brief Should remove old tunnels after tunnels config is updated
-  /// TODO(unassigned): not fully implemented
-  void RemoveOldTunnels(
-      std::vector<std::string>& updated_tunnels);
+  void RemoveOldTunnels(std::vector<std::string>& updated_tunnels);
```
anonimal: The commit says `Instance: fix RemoveOldTunnels` but `RemoveOldTunnels()` is still not fully implemented (see definition). Can you please keep the TODO (or fully implement)?

src/app/instance.cc
```diff
@@ -80,12 +78,11 @@ void Instance::Initialize() {
 
 void Instance::Reload() {
   LOG(info) << "Instance: reloading";
-  // TODO(unassigned): reload kovri.conf
   // TODO(unassigned): locking etc.
   // TODO(unassigned): client/router contexts
-  m_IsReloading = true;
-  SetupTunnels();
-  m_IsReloading = false;
+  GetConfig().ParseKovriConfig();
```
anonimal: This doesn't work (`kovri.conf` re-loading). It won't work without work elsewhere (see other TODO's). Can you please remove this line or implement?

src/client/context.cc
```diff
@@ -370,8 +356,9 @@ void ClientContext::UpdateClientTunnel(
     if (ec)  // New address is not an IP address, compare strings
       rebind = (tunnel.address != current_addr);
     else  // New address is an IP address, compare endpoints
-      rebind = (client_tunnel->GetEndpoint() == boost::asio::ip::tcp::endpoint(
-          next_addr, tunnel.port));
+      rebind =
+          (client_tunnel->GetEndpoint()
+           != boost::asio::ip::tcp::endpoint(next_addr, tunnel.port));
```
anonimal: 👍

src/app/instance.h
```diff
@@ -76,23 +76,17 @@ class Instance {
   void InitClientContext();
 
   /// @brief Sets up (or reloads) client/server tunnels
+  /// @param is_reloading Are tunnels configuration in the process of reloading?
   /// @warning Configuration files must be parsed prior to setup
-  void SetupTunnels();
+  void SetupTunnels(bool is_reloading = false);
```
anonimal: >removing member m_IsReloading

1. Asumming that locks are not where they should be, if `Instance::Reload()` is called consecutively (multiple consecutive `SIGHUP`'s), I think we'll be looking at possible race conditions

2. `SetupTunnels()` with member `m_IsReloading` is a cheap way to resolve point 1 above

3. I find bool `is_reloading` as function arg to be somewhat confusing but I think the real issue is *how* we setup/reload tunnels.

Any other thoughts? How else could we approach these issues without more patching?

src/client/context.h
```diff
@@ -174,20 +186,16 @@ class ClientContext {
...
 
   /// @return the client tunnel with the given name, or nullptr
-  std::unique_ptr<I2PClientTunnel> GetClientTunnel(
-      const std::string& name);
+  I2PClientTunnel* GetClientTunnel(const std::string& name);
 
   /// @return the client tunnel with the given (local) port
-  std::unique_ptr<I2PClientTunnel> GetClientTunnel(
-      int port);
+  I2PClientTunnel* GetClientTunnel(int port);
```
anonimal: >raw pointer refactoring

I'm glad my suggestions on IRC worked out 😄

src/client/context.h
```diff
@@ -138,12 +138,24 @@ class ClientContext {
...
+  /// @param http true if server tunnel is an HTTP tunnel
+  /// @return true if the tunnel was inserted, false otherwise
+  bool AddServerTunnel(
+      const TunnelAttributes& tunnel,
+      bool is_http);  // TODO(anonimal): this isn't ideal
+
+  /// @brief Creates the specified client tunnel and tries to insert it
+  /// @param tunnel Const reference to populated/initialized tunnel attributes class
+  /// @return true if the tunnel was inserted, false otherwise
+  bool AddClientTunnel(const TunnelAttributes& tunnel);
```
anonimal: >pulling out tunnel creation functionality
👍

src/app/instance.h
MoroccanMalinois: Sorry @anonimal , but i really don't see what's left to do. This function removes all tunnels that are not present in "updated_tunnels". What am i missing ?

anonimal: AFAICT, nothing! I was distracted and overlooked the 2 important lines that were already implemented. My sincerest apologies.

src/app/instance.h
MoroccanMalinois: I removed this commit. I guess this falls in "// TODO(unassigned): locking etc."

src/app/instance.cc
MoroccanMalinois: Line removed

Issue #362, opened 2016-09-15
Packaging for release


anonimal:
**By submitting this issue, I confirm the following:**
- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

_Place an X inside the bracket to confirm_
- [X] I confirm.

---

Opening now so we don't forget later.

I'm assuming we'll require #46 to be resolved first?
For distros/platforms, I imagine at a minimum; Arch, Debian/Ubuntu (PPA too?), and OSX.

@fluffypony said he'd look into this. I'd rather not have to deal with packaging and neither does @EinMByte I'll take this ticket at the last minute if no one else gets to it in time.

anonimal:
AUR is easy. The others not so much because they tend to lag on dependency versions.

For starters, we can simply focus on static binaries. The only issue is that we need to ship extra clutter (certs, etc.) so we'll need an easy **cross-platform** solution to handle that. One possible solution:

```
&anonimal Self-extracting binary would be nice.
&anonimal I think one cross-platform solution could be to 7z an sfx archive that does a recursive self-install.
&anonimal We can adjust the details as necessary per platform since the site will be per-platform downloads.
&anonimal Really not the most fun to maintain but so far the most compact and user-friendly option I can think of right now.
```

anonimal:
Note, as referenced in the above meta repository commits/PR, for now we're going to work with makeself for nightly/branch-tip static releases.

anonimal:
```
&anonimal Alright, as I mentioned in #monero-dev, the win build is now officially static (minus the usual system32 dll's).
&anonimal The problem being openssl and the fact we *don't* want it static but really have no choice at this time (afaict).
&anonimal So, that just makes for more work for me or whoever wants to keep up with openssl stable releases.
&anonimal (or minimum hot fixes)
&anonimal pigeons: is there a way we could automate some of this task? RSS feed + email notification + triggers buildbot + uploads to wherever? (something like that maybe?)
&anonimal And then there's the question of making openssl static for all non-windows platforms too (for static builds). Maybe update them all in one fell swoop when the time comes if we can get the process automated?
+pigeons ok we can set something up, but seems like something to manually trigger, but still setup the rest of the automation
[...]

PR #610, opened 2017-04-02
Packager: add bash completion script for kovri

rakhimov:
This is a script for kovri command-line completion.
It's nice to have for packaging,
relevant to issue #362.

Adopted from [SCRAM](https://github.com/rakhimov/scram)

---

---

### DIFF
--- a/pkg/kovri-bash.sh
+++ b/pkg/kovri-bash.sh
@@ -1,0 +1,91 @@
+# Bash completion for kovri.
+# source this script or put it in /usr/share/bash-completion/completions/
+
+########################################
+# Searches for short and long options
+# starting with "-" and prints them
+#
+# Globals:
+#   None
+# Arguments:
+#   A line containing options
+# Returns:
+#   Short option
+#   Long option
+########################################
+_parse_boost_program_options() {
+  local short_option long_option w
+  for w in $1; do
+    case "${w}" in
+      -?) [[ -n "${short_option}" ]] || short_option="${w}" ;;
+      --*) [[ -n "${long_option}" ]] || long_option="${w}" ;;
+    esac
+  done
+
+  [[ -n "${short_option}" ]] && printf "%s\n" "${short_option}"
+  [[ -n "${long_option}" ]] && printf "%s\n" "${long_option}"
+}
+
+########################################
+# Parses Boost program options
+# generated help description to find
+# options
+#
+# Globals:
+#   None
+# Arguments:
+#   None
+# Returns:
+#   Options
+########################################
+_parse_kovri_help() {
+  local line
+  kovri --help | while read -r line; do
+    [[ "${line}" == *-* ]] || continue
+    _parse_boost_program_options "${line}"
+  done
+}
+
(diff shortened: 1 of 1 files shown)


anonimal:
Other solutions are welcome too.

rakhimov:
@anonimal I agree it is a nasty approach to hardcode and keep in sink.
It would be nice to find "single source of truth" for options and their argument types.
Unfortunately, Boost does not expose argument types (or possible values) by default;
it needs some manual code with 'value_type' and custom conventions for types.

The hardcoding can be swapped with the following code:
```bash
-*)
      (kovri --help | grep -q -E -- "${prev}.*\sarg\s") && return
      ;;
```
But again, it is not possible to deduce what to suggest.
Is it a file directory, string, number, boolean?

Note that the hardcoding is only for argument suggestion (i.e., not the completion/deduction),
so if options are added or changed, nothing really breaks.

anonimal:
>(kovri --help | grep -q -E -- "${prev}.*\sarg\s") && return
>But again, it is not possible to deduce what to suggest.
Is it a file directory, string, number, boolean?

If we *really* wanted to keep `COMPREPLY`, maybe this is where a config file addendum could come in (alla quasi-INI):

```
# [String]

# comment = "This is log-level description, etc."
# option = "0 1 2 3 4 5"
log-level = 3

# comment = "This is port help"
port = 12345

# [Bool]

# comment = "This is enable-ssu description, etc."
# option = "on off yes no true false 1 0"
enable-ssu = 1
```

and then adjust the script to parse as needed. Syntax being flexible. Would this be worth the effort?

> so if options are added or changed, nothing really breaks.

Indeed, but I still think we should avoid any future potential dead code when possible.

rakhimov:
Extra info in conf will help,
but now you have the same problem of keeping Boost options and conf options in sync.

Why not autogenerate the conf example from Boost (or vice versa)?
(or be implicit about it and document the association between cmd options and config vars)

This could be done by adding help-info into Boost options and consistently specifying their type in code
with value_name,
so the following pattern would be expected:
```bash
--option type  (=default) description
```
The type could be 'path, string, int, float'.

rakhimov:
This duplication issue may come up again with man pages as well.

anonimal:
Ok. How about we merge what we have so far and open a new issue that will tackle the issues above? If so, will you open the new issue after I merge?

rakhimov:
@anonimal Sure, I will.

anonimal:
Two sidenotes:

1. If this is to make it into the binary distribution or even `make install`, the installers will need to know of this file. We should also put a little note in documentation because end-users may not know of the purpose of this script (and then get confused when they see it).

2. The script should be set executable.

rakhimov:
I don't think the script must be executable.
The completion scripts of other packages in /usr/share/bash-completion/completions don't have such property.
It may cause packaging issues.

anonimal:
>I don't think the script must be executable.

Sorry! I was completely distracted and not thinking. That is correct.

PR #613, opened 2017-04-05
Instance: fixes for reload client tunnel

MoroccanMalinois:
---

---

Referencing #96

### DIFF
--- a/src/app/config.cc
+++ b/src/app/config.cc
@@ -209,6 +209,20 @@ void Configuration::ParseTunnelsConfig() {
         tunnel.keys = value.get<std::string>(GetAttribute(Key::Keys), "");
         // Parse for CSV destinations + dest:port, then set appropriately
         ParseClientDestination(&tunnel);
+        // Check for conflicting port
+        if (std::find_if(
+                m_TunnelsConfig.begin(),
+                m_TunnelsConfig.end(),
+                [&tunnel](
+                    kovri::client::TunnelAttributes const& tunnel_attribute) {
+                  return tunnel.port == tunnel_attribute.port;
+                })
+            != m_TunnelsConfig.end())
+          {
+            LOG(error) << "Config: " << tunnel.name
+                       << " will not be loaded, conflicting port";
+            continue;
+          }
       } else if (tunnel.type == GetAttribute(Key::Server)
                 || tunnel.type == GetAttribute(Key::HTTP)) {
         tunnel.in_port = value.get<std::uint16_t>(GetAttribute(Key::InPort), 0);
--- a/src/app/instance.cc
+++ b/src/app/instance.cc
@@ -176,15 +176,17 @@ void Instance::SetupTunnels() {
           ||tunnel.type == GetConfig().GetAttribute(Key::IRC)) {  // TODO(unassigned): see #9
         if (m_IsReloading) {
           auto client_tunnel = kovri::client::context.GetClientTunnel(tunnel.port);
-          if (client_tunnel && client_tunnel->GetName() != tunnel.name) {
-            // Conflicting port
-            // TODO(unassigned): what if we interchange two client tunnels' ports?
-            // TODO(unassigned): the addresses could differ
-            LOG(error)
-              << "Instance: " << tunnel.name
-              << " will not be updated, conflicting port";
-            continue;
-          }
+          if (client_tunnel && client_tunnel->GetName() != tunnel.name)
+            {
+              // Check for conflicting port done in ParseTunnelsConfig
+              // early deletion of client_tunnel to avoid temp duplicate port bind
+              std::string name = client_tunnel->GetName();
+              LOG(debug) << "ClientContext: Premature delete tunnel " << name;
+              kovri::client::context.RemoveClientTunnels(
+                  [&name](kovri::client::I2PClientTunnel* old_tunnel) {
+                    return name == old_tunnel->GetName();
+                  });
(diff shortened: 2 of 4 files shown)


src/app/config.cc
```diff
@@ -209,6 +209,19 @@ void Configuration::ParseTunnelsConfig() {
         tunnel.keys = value.get<std::string>(GetAttribute(Key::Keys), "");
         // Parse for CSV destinations + dest:port, then set appropriately
         ParseClientDestination(&tunnel);
+        // Check for conflicting port
+        if (std::find_if(
+                m_TunnelsConfig.begin(),
+                m_TunnelsConfig.end(),
+                [&tunnel](kovri::client::TunnelAttributes const& rhs) {
+                  return tunnel.port == rhs.port;
```
anonimal: Can we please continue to keep naming conventions as specified in ["General Naming Rules"](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules)?

src/app/instance.cc
```diff
@@ -176,15 +176,17 @@ void Instance::SetupTunnels() {
...
-          }
+          if (client_tunnel && client_tunnel->GetName() != tunnel.name)
+            {
+              // Check for conflicting port done in ParseTunnelsConfig
+              // early deletion of client_tunnel to avoid temp duplicate port bind
+              std::string name = client_tunnel->GetName();
+              LOG(debug) << "ClientContext: Premature delete tunnel " << name;
+              kovri::client::context.RemoveClientTunnels(
+                  [&name](kovri::client::I2PClientTunnel* rhs) {
+                    return name == rhs->GetName();
```
anonimal: >Can we please continue to keep naming conventions as specified in ["General Naming Rules"](https://google.github.io/styleguide/cppguide.html#General_Naming_Rules)?

Here too please.

src/app/instance.cc
```diff
@@ -176,15 +176,17 @@ void Instance::SetupTunnels() {
...
-            // TODO(unassigned): the addresses could differ
-            LOG(error)
-              << "Instance: " << tunnel.name
-              << " will not be updated, conflicting port";
-            continue;
-          }
+          if (client_tunnel && client_tunnel->GetName() != tunnel.name)
+            {
+              // Check for conflicting port done in ParseTunnelsConfig
+              // early deletion of client_tunnel to avoid temp duplicate port bind
```
anonimal: >ParseTunnelsConfig
>client_tunnel

FYI, in the future, I think it's best to avoid specifying names in comments because A) they can be missed during future refactoring B) IDEs may not catch them during automatic name-changing

src/client/context.cc
```diff
@@ -359,6 +358,29 @@ void ClientContext::UpdateClientTunnel(
...
            != boost::asio::ip::tcp::endpoint(next_addr, tunnel.port));
+    if (!rebind)  // check for change in destination
+      {
+        LOG(trace)
+            << "ClientContext: update checking for change in destination";
+        rebind = (client_tunnel->GetTunnelAttributes().dest != tunnel.dest)
+                 || (client_tunnel->GetTunnelAttributes().dest_port
+                     != tunnel.dest_port);
+      }
+    if (!rebind)  // check for change in keys
```
anonimal: Why can't we keep this block within the above block?

src/client/context.cc
```diff
@@ -359,6 +358,29 @@ void ClientContext::UpdateClientTunnel(
...
+      {
+        LOG(trace) << "ClientContext: update checking for change in keys";
```
anonimal: `debug` please. I think we should save `trace` for encoded binary output and finer debugging.

src/client/tunnel.h
```diff
@@ -190,6 +190,13 @@ class I2PClientTunnel : public TCPIPAcceptor {
     return m_TunnelAttributes.name;
   }
 
+  /// @brief Set tunnel attributes
+  /// @param tunnel Initialized/populated tunnel attributes class
+  void SetTunnelAttributes(const TunnelAttributes& tunnel) noexcept
+  {
+    m_TunnelAttributes = tunnel;
+  }
```
anonimal: Yay, a use-case for not only server tunnel 😃

MoroccanMalinois:
@anonimal Ready for another round

src/client/context.cc
```diff
@@ -359,6 +358,26 @@ void ClientContext::UpdateClientTunnel(
...
+        if (local_destination->GetIdentHash()
+            != client_tunnel->GetLocalDestination()->GetIdentHash())
+          {
+            client_tunnel->SetLocalDestination(local_destination);
+            rebind = true;
+          }
+      }
+    client_tunnel->SetTunnelAttributes(tunnel);
     if (rebind) {
       // The IP address has changed, rebind
```
anonimal: This comment needs updating or removal.

src/client/context.cc
```diff
@@ -359,6 +358,26 @@ void ClientContext::UpdateClientTunnel(
       rebind =
           (client_tunnel->GetEndpoint()
            != boost::asio::ip::tcp::endpoint(next_addr, tunnel.port));
+    if (!rebind)  // check for change in destination
```
anonimal: Can we add a quick note here about *why* we want to rebind when these tunnel attributes change (e.g., a quick/easy way to update attributes even if local address/port hasn't changed)?

MoroccanMalinois:
Done

PR #615, opened 2017-04-12
Configuration: check user input port in valid range

MoroccanMalinois:
---

---

Not sure about this one, but makes sense to me
#96

### DIFF
--- a/pkg/config/kovri.conf
+++ b/pkg/config/kovri.conf
@@ -47,8 +47,8 @@
 #  Port to listen on
 #  =================
 #
-#  Default: a random port will be generated with each use if its not
-#  specified by command line or here on the configuration file
+#  Default: a random port between 9111 and 30777 will be generated with each
+#  use if its not specified by command line or here on the configuration file
 #
 
 #port =
--- a/src/app/config.cc
+++ b/src/app/config.cc
@@ -168,6 +168,14 @@ void Configuration::ParseKovriConfigFile(
     throw std::runtime_error("Could not open " + file + "!\n");
   bpo::store(bpo::parse_config_file(filename, options), var_map);
   bpo::notify(var_map);
+  // Ensure port in valid range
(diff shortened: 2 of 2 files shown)


NanoAkron:
Users are the worst people to ask about security :) Make things configurable for experts, but use defaults for the average joe.

Easy beats secure every time in userland

anonimal:
>Make things configurable for experts, but use defaults for the average joe

@NanoAkron indeed! FYI, the default random generated port will use the appropriate range *without* this PR. This PR only enforces the config option; thus circumventing the ability for an "expert" to use a custom port out of range for whatever reason.

MoroccanMalinois:
hmm ... in a way, this looks kinda similar to the monero static ring sizes discussion, in the sense that the protocol allows for the user to choose the value he wants and people should be able to do what the protocol allows. But in the same time, this is a domain when you hurt other's privacy when you hurt yourself, so we are kinda incentivized to protect people from themselves. 

The default range looks large enough. So far, i can't think of any advantage to use a custom port outside the one used by the vast majority. I tend to think that people will mostly do it not knowing about the default port range. But this is not a good reason. I guess my main pb is that i still have hard time believing that it would not effect someone's anonymity if he was the only one choosing a custom port number. 

    ultimately, the decision should be defined by the end-user

IMHO, this depends on the ability of the majority of the end-users to understand the consequences of their decisions.

Having said this, i don't have a strong opinion on this point.

anonimal:
>monero static ring sizes discussion

Fortunately, what were dealing with is infinitesimally less important in terms of security.

>the protocol allows

Note: there is no specification for this, I believe it is entirely impl-defined.

anonimal:
As discussed on IRC, putting aside the fact that we have free-range to define a new standard, since we already have these parameters in place, I think we should merge and revisit/pursue the question in the future should we ever fork the network. Until then, if the PR'd code becomes a problem, we can easily remove said code.

PR #616, opened 2017-04-12
use the recently added GetAddressBook function

selene-kovri:
Notably, this involves calling LoadSubscriptionsFromPublisher
at startup, rather than lazily.

---

---

### DIFF
--- a/src/client/address_book/impl.cc
+++ b/src/client/address_book/impl.cc
@@ -90,6 +90,7 @@ void AddressBook::Start(
         m_SharedLocalDestination->GetService());
   m_SubscriberUpdateTimer->expires_from_now(
       boost::posix_time::minutes(SubscriberTimeout::InitialUpdate));
+  LoadSubscriptionFromPublisher();
   m_SubscriberUpdateTimer->async_wait(
       std::bind(
           &AddressBook::SubscriberUpdateTimer,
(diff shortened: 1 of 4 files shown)


selene-kovri:
That's a followup to 579, but I feel like I'm missing something since it was trivial to change, and it works.

src/client/address_book/impl.cc
```diff
@@ -408,9 +410,9 @@ bool AddressBook::CheckAddressIdentHashFound(
 
 // For in-net download only
 std::unique_ptr<const kovri::core::IdentHash> AddressBook::GetLoadedAddressIdentHash(
-    const std::string& address) {
+    const std::string& address) const {
   if (!m_SubscriptionIsLoaded)
-    LoadSubscriptionFromPublisher();
```
anonimal: Removing both?! `m_SubscriptionIsLoaded` may never be true...

anonimal:
@selene-kovri status?

selene-kovri:
Hello,

I'm really sorry, I think I vastly overestimated the time I'd have available.
I had a look at your comments and tried to understand how it all fits together to see when it makes sense to "lazy load" and not, but I'll need a lot more time to see where everything fits I think. If and when I do, I'll post again, but I think it's best to assume it won't be soon indeed.

anonimal:
Hi @selene-kovri. Ok, thanks for the message. In the meantime, if you want to work on other areas of the project then that is great too.

PR #621, opened 2017-04-14
RouterInfo: various refactoring

anonimal:
WIP. Live tests pass. Referencing #617. This WIP is in pursuit of isolating #187.

**Note for any reviewers**: if skimming through the GitHub *"Files Changed"* tab, a clickable *"Large diffs are not loaded by default"* for `router/info.cc` is tucked into the diff.

---

---

### DIFF
--- a/src/core/router/info.h
+++ b/src/core/router/info.h
@@ -70,29 +70,37 @@ const int MAX_RI_BUFFER_SIZE = 2048;
 
 class RouterInfo : public RoutingDestination {
  public:
-  enum SupportedTransports {
-    eNTCPV4 = 0x01,
-    eNTCPV6 = 0x02,
-    eSSUV4 = 0x04,
-    eSSUV6 = 0x08
+  /// @enum Transport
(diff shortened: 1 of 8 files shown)


src/core/router/info.h
```diff
@@ -51,67 +51,181 @@
...
     boost::asio::ip::address host;
-    int port;
+    std::uint16_t port{};
     Tag<32> key;
-    std::uint32_t tag;
-
-    /// @brief Human readable description of this struct
-    /// @param prefix for tabulations
-    /// @returns human readable string
-    std::string GetDescription(const std::string& tabs = std::string()) const;
```
anonimal: @MoroccanMalinois FYI, OOP-wise, I preferred your approach but, considering the use-case, the extra hoops we'd have to jump through to implement nested scope access to enclosed members would probably not have been in our best interest.

MoroccanMalinois: @anonimal i totally agree :)

PR #641, opened 2017-05-16
Reseed: skip SU3 checks when reseed-from .zip

MoroccanMalinois:
---

---

### DIFF
--- a/src/client/reseed.cc
+++ b/src/client/reseed.cc
@@ -237,22 +237,31 @@ bool Reseed::FetchStream(
  *   - Extract RI's for Reseed
  *
  */
-bool SU3::SU3Impl() {
-  /**
-   * TODO(unassigned): when the --reseed-from file is a .zip, (or non-su3 type),
-   * it would be nice to skip su3 all-together, decompress as appropriate, and
-   * validate for RI type - then insert into NetDb.
-   */
-  LOG(debug) << "SU3: preparing stream...";
-  if (!PrepareStream()) {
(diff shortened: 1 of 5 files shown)


src/client/reseed.cc
```diff
@@ -237,22 +238,31 @@ bool Reseed::FetchStream(
...
-    return false;
-  }
-  LOG(debug) << "SU3: verifying stream...";
-  if (!VerifySignature()) {
-    LOG(error) << "SU3: verification failed";
-    return false;
-  }
+bool SU3::SU3Impl()
+{
+  if (boost::algorithm::ends_with(kovri::context.GetOptionReseedFrom(), ".zip"))
```
anonimal: I avoided something like this because we should determine filetype not by extension but by specification. If you'd rather not implement, could you add a TODO for this instead?

I also think we should consider implementing another run-time switch to ensure that the user knows the risks involved (e.g., `--disable-su3-verification`) and the switch would need to be flipped in order to reseed purely from zip. What do you think?

MoroccanMalinois: added option --disable-su3-verification. i think with this option, an early detection is not needed anymore since it's the first thing done in ZIP impl and we only support zip, so the check will be redundant. Or maybe i should add a todo with "detection an implementation of other format" ?

anonimal: Yes, I think that TODO is a great idea.

MoroccanMalinois: Done

PR #645, opened 2017-05-22
kovri-install.sh: fix option checking

pensanto:
The current comparison between $_create_checksum_file and false at line 137 will always yield false since $_create_checksum_file at that point is either empty or true.  I am guessing that the intent is to fail if the caller supplies -c, but not -p.  If this is the case, then the script should fail if $_create_checksum_file is true, which is what this commit does.

(I ran into this when trying to port the install script to a shell different from bash for my own purposes.  Replacing the "== false" with "!= true", which is the right way to do this comparison results in the script failing the option checks when called without arguments from 'make install'.)

---

---

### DIFF
--- a/pkg/kovri-install.sh
+++ b/pkg/kovri-install.sh
@@ -134,7 +134,7 @@ PrepareOptions() {
     fi
   else
     # Ensure proper command line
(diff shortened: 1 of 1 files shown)


anonimal:
Thanks @pensanto

Issue #399, opened 2016-10-14
A new, separate binary for an anytime (offline/online) utility


anonimal:
**By submitting this issue, I confirm the following:**
- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

_Place an X inside the bracket to confirm_
- [X] I confirm.

---

```
&anonimal       Ooo, guzzi I just thought of a great side-project I think you'd love to do.
&anonimal       Converting a few of the tools in i2p.jar until a utility binary that we can use.
&anonimal       Or simply working out what we already have in lib form and making a utility binary that hooks into our lib.
&anonimal       (or something similar, probably more on the lib side so we can use it elsewhere)
&anonimal       guzzi: $ java -jar lib/i2p.jar <-- running this command in your I2P directory will show you the commands available
&anonimal       addresses/base32/base64/cpuid/eepget/su3file <-- maybe just these to start
&anonimal       We already have most of the code (it's just spattered everywhere)
```

This also includes `routerinfo` in `router.jar`.

Basically, nearly all of the implementation code is done; the issue is mostly a matter of hooking into both the client and core libraries and processing what's returned, etc. I had kept eepget in mind when rewriting our download impl (I believe that's noted somewhere in there). The same goes for SU3 when I did that rewrite/refactor a while ago.

More work would need to go into a cpuid utility but, for x86/x86_64 AES-NI, that part's already implemented core-side.

Base32/64 should be all there so it's simply a matter of io/stream handling. I believe an addresses and routerinfo option will take more work as there isn't a clear separation of functionality within the router (surprise!).

@guzzijones feel free to ping me if you want to pursue this and/or need any direction

anonimal:
Note: when API's are resolved (#350 #351), perhaps we could simply use those.

anonimal:
In our most recent meeting, @MoroccanMalinois offered to take this issue and @guzzijones will do some work in this area too. +1

anonimal:
The utility is functional and well into development. Additional features can be added as needed. Closing as resolved.

PR #659, opened 2017-06-18
kovri-util: fix uncaught exception for routerinfo create

MoroccanMalinois:
---

---
#399

### DIFF
--- a/src/util/routerinfo.cc
+++ b/src/util/routerinfo.cc
@@ -77,19 +77,133 @@ bool RouterInfoCommand::Impl(
     const std::string& cmd_name,
     const std::vector<std::string>& args)
 {
-  std::vector<std::string> inputs;
-  bpo::positional_options_description pos;
-  pos.add("args", -1);
-
-  bpo::variables_map vm;
   try
     {
+      std::vector<std::string> inputs;
+      bpo::positional_options_description pos;
+      pos.add("args", -1);
+
+      bpo::variables_map vm;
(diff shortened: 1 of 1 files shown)


MoroccanMalinois:
Done

src/util/routerinfo.cc
```diff
@@ -77,157 +77,139 @@ bool RouterInfoCommand::Impl(
...
+      for (const auto& arg : inputs)
         {
+          core::InputFileStream input(arg, std::ios::in | std::ios::binary);
+          if (input.Fail())
+            {
+              LOG(error) << "routerinfo: Failed to open input " << arg;
+              return false;
+            }
+          // Read all input
+          std::size_t length(0);
```
anonimal: JFTR, since c++11:

>If T is a scalar type, the object's initial value is the integral constant zero explicitly converted to T.

So, `length()` or `length{}` should also work.

src/util/routerinfo.cc
```diff
@@ -77,157 +77,139 @@ bool RouterInfoCommand::Impl(
...
-        {
-          LOG(error) << "routerinfo: Failed to open input " << arg;
-          return false;
-        }
-      // Read all input
-      std::size_t length(0);
-      std::unique_ptr<std::uint8_t[]> buffer =
-          input.ReadAll<std::uint8_t, std::size_t>(&length);
-      if (!buffer)
+      if (inputs.size() == 0)
```
anonimal: In the future, `if (inputs.empty())`

PR #660, opened 2017-06-18
Daemon: call Stop on Init exception

MoroccanMalinois:
---

---
Not sure it's the best fix, but the app didn't exits with the following mistake (stays hanging after return EXIT_FAILURE)

    ./build/kovri --host blabla

### DIFF
--- a/src/app/config.cc
+++ b/src/app/config.cc
@@ -172,6 +172,12 @@ void Configuration::ParseKovriConfigFile(
     throw std::runtime_error("Could not open " + file + "!\n");
   bpo::store(bpo::parse_config_file(filename, options), var_map);
   bpo::notify(var_map);
+  // Check host syntax
+  boost::system::error_code ec;
+  boost::asio::ip::address::from_string(
+      m_KovriConfig["host"].as<std::string>(), ec);
+  if (ec)
+    throw std::runtime_error("Invalid host: " + ec.message());
   // Ensure port in valid range
   if (!m_KovriConfig["port"].defaulted())
     {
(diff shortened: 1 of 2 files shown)


src/app/daemon.cc
```diff
@@ -81,6 +81,7 @@ bool DaemonSingleton::Init() {
     m_Instance->Initialize();
   } catch (const std::exception& ex) {
     LOG(error) << "DaemonSingleton: exception during initialization: " << ex.what();
+    Stop();
```
anonimal: >Not sure it's the best fix

Yes, I agree. We shouldn't need to run this call if we've technically never `Start`ed 😏 

For the time being, will you consider applying at least one of these two options (or both)?:

1. Adding a quick check for `host` validity during instance *configuration* and bail there if `host` is invalid (can be ensured with members from `boost::asio::ip::address`)
2. Adding a TODO above this line (Instance and RAII refactoring is unfinished)

Any other ideas for a quick fix?

MoroccanMalinois:
Done

src/app/config.cc
```diff
@@ -172,6 +172,12 @@ void Configuration::ParseKovriConfigFile(
     throw std::runtime_error("Could not open " + file + "!\n");
   bpo::store(bpo::parse_config_file(filename, options), var_map);
   bpo::notify(var_map);
+  // Check host syntax
+  boost::system::error_code ec;
+  boost::asio::ip::address::from_string(
+      m_KovriConfig["host"].as<std::string>(), ec);
+  if (ec)
+    throw std::runtime_error("Invalid host !\n");
```
anonimal: Can you use`ec.message()` for the string?

PR #663, opened 2017-06-22
SSU: fix SendSessionCreated endianness

MoroccanMalinois:
---

---

Can't find this precisely in the specs, but [this is how](https://github.com/monero-project/kovri/blob/master/src/core/router/transports/ssu/session.cc#L402) the receiver is processing it (this difference between ProcessSessionCreated and SendSessionCreated is what's causing the error descrived in #661 )

Ref #140

### DIFF
--- a/src/core/router/transports/ssu/session.cc
+++ b/src/core/router/transports/ssu/session.cc
@@ -469,7 +469,7 @@ void SSUSession::SendSessionCreated(
     packet.SetIPAddress(remote_address.to_v6().to_bytes().data(), 16);
     s.Insert(remote_address.to_v6().to_bytes().data(), 16);
   }
-  s.Insert<std::uint16_t>(packet.GetPort());  // remote port
+  s.Insert<std::uint16_t>(htobe16(packet.GetPort()));  // remote port
   if (address->host.is_v4())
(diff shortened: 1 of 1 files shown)


anonimal:
From IRC:

```
@anonimal       MoroccanMalinois: #663 is a logical observation but the revision appears to not resolve failed sig verification on livenet.
```
```
@anonimal       MoroccanMalinois: since it's not in the spec, looking at PacketBuilder.java: their implementation assumes that the data is big endian. 
 *      anonimal basing this off of DataHelper.java
@anonimal       So, with that assumption made, I think we can assume that #663 *should* be in place.
@anonimal       MoroccanMalinois: what are you seeing on your testnet? 
@anonimal       Because I still see failed sig verification, and without further debugging of SSU issues, I cannot easily confirm that #663 will have any noticeable effect (positive or negative) without more test-cases and logging in place.
@anonimal       I'll start some work on the testnet script. Do you want me to work with you on #140? The need to resolve SSU issues is on par with the NTCP/NetDb/I2NP work that I'd been doing.
```

anonimal:
Thanks @MoroccanMalinois 👍 #140 is looking forward to your work 😄

PR #664, opened 2017-06-24
SSU:  fix ProcessSessionConfirmed SignedOnTime

MoroccanMalinois:
---

---
Ref #140 

Looks like we were appending SignedOnTime instead of replacing the previous.

### DIFF
--- a/src/core/router/transports/session.h
+++ b/src/core/router/transports/session.h
@@ -70,6 +70,18 @@ class SignedData {
     m_Stream.write(reinterpret_cast<char *>(&type), sizeof(T));
   }
 
+  /// @brief Insert at a specific position
+  /// @param way beginning, current or end
+  /// @param offset value relative to the seekdir parameter
+  /// @param buf buffer
+  /// @param len length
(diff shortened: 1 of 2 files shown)


src/core/router/transports/session.h
```diff
@@ -70,6 +70,18 @@ class SignedData {
     m_Stream.write(reinterpret_cast<char *>(&type), sizeof(T));
   }
 
+  /// @brief Insert at a specific position
+  /// @param way beginning, current or end
+  /// @param offset value relative to the seekdir parameter
+  /// @param buf buffer
+  /// @param len length
+  template <typename T>
+  void InsertAt(std::ios_base::seekdir way, std::streamoff offset, T type)
```
anonimal: JFTR: because of the use-case, we shouldn't *need* to insert at a specific position - but a bigger fix would require a bigger refactor (and this patch fixes the appending issue).

Until then, could you please overload `Insert` for consistency?

MoroccanMalinois: done

PR #665, opened 2017-07-01
SSU: fix SessionConfirmed padding

MoroccanMalinois:
---

---
2 errors in padding calc:
- OutputByteStream::m_Length is the remaining length and not the current size, so (m_Length - init_length) is a negative value.
- It didn't take into account header size in ParseSessionConfirmed (as opposed to SSUPacketBuilder::WriteSessionConfirmed)

Jftr, from [spec](https://geti2p.net/spec/ssu#ssu) : 

> Since the signature is at the end, the padding in the last or only packet must pad the total packet to a multiple of 16 bytes, or the signature will not get decrypted correctly. This is different from all the other message types, where the padding is at the end.

Ref #140


src/core/router/transports/ssu/packet.cc
```diff
@@ -774,7 +776,8 @@ std::unique_ptr<SSUSessionConfirmedPacket> SSUPacketParser::ParseSessionConfirme
     throw std::length_error("SSUPacketParser: invalid length within identity");
   packet->SetRemoteRouterIdentity(identity);
   packet->SetSignedOnTime(ReadUInt32());
-  const std::size_t padding_size = ((m_Length - init_length) + identity.GetSignatureLen()) % 16;
+  const std::size_t padding_size = SSUPacketBuilder::GetPaddingSize(
+      header->GetSize() + init_length - m_Length + identity.GetSignatureLen());
```
anonimal: Much better 😄 👍

src/core/router/transports/ssu/packet.h
```diff
@@ -755,8 +755,11 @@ class SSUPacketParser : private kovri::core::InputByteStream {
   std::unique_ptr<SSUSessionCreatedPacket> ParseSessionCreated();
 
   /// @brief Parses a session confirmed packet, without the header
+  /// @param header Packet header
+  /// @note header is required to compute padding
   /// TODO: Support multiple fragments?
-  std::unique_ptr<SSUSessionConfirmedPacket> ParseSessionConfirmed();
+  std::unique_ptr<SSUSessionConfirmedPacket> ParseSessionConfirmed(
+      const SSUHeader* header);
```
anonimal: I think we should consider using a member instead of an argument. Benefits include easier access to parsed header data (so we're not passing header data to *packet* parsing functions) and class design consistency (since we *do* do header parsing here after all).

Any thoughts?

tests/unit_tests/core/router/transports/ssu/packet.cc
```diff
@@ -36,14 +36,38 @@
...
+    // Build session_confirmed
+    core::OutputByteStream output(session_confirmed.data(), session_confirmed.size());
+    // 1 byte info
+    output.WriteUInt8(0x01);
+    // 2 byte identity size (0x01, 0x87)
+    output.WriteUInt16(m_AliceIdentity.size());
+    // Append identity
+    output.WriteData(m_AliceIdentity.data(), m_AliceIdentity.size());
+    // Signed on time (0x57, 0x69, 0x04, 0xAA)
+    output.WriteUInt32(1466500266);
```
anonimal: >1466500266

Could you please create a definition since this is used throughout several areas?

tests/unit_tests/core/router/identity.cc
```diff
@@ -0,0 +1,88 @@
...
+
+BOOST_AUTO_TEST_CASE(ParseIdentity)
+{
+  // Parse
+  core::IdentityEx identity;
+  BOOST_CHECK(
+      identity.FromBuffer(m_AliceIdentity.data(), m_AliceIdentity.size()));
+  // Check that FromBuffer + ToBuffer == original buffer
+  // TODO(anonimal): review the following arbitrary size (must be >= 387)
+  std::array<std::uint8_t, 1024> output{{}};
```
anonimal: >{{}}

👍

tests/unit_tests/core/router/transports/ssu/packet.cc
```diff
@@ -551,4 +592,44 @@ BOOST_AUTO_TEST_CASE(SessionCreatedPlain) {
...
+  packet.SetSignedOnTime(1466500266);
+  std::size_t sig_position =
+      session_confirmed.size() - identity.GetSignatureLen();
+  packet.SetSignature(&session_confirmed.at(sig_position));
+  // Output to buffer
+  auto buffer = std::make_unique<std::uint8_t[]>(packet.GetSize());
+  core::SSUPacketBuilder builder(buffer.get(), packet.GetSize());
+  builder.WriteSessionConfirmed(&packet);
+  // Padding is randomized, so check everything before and after
+  std::size_t padding_position = 1  // Info
```
anonimal: >const correctness

Can you please add `const`?

anonimal:
Quick note, I forgot to mention:

> so (m_Length - init_length) is a negative value.

I don't see how that's possible since both types are unsigned (an underflow, yes). Did you mean something else?

MoroccanMalinois:
> Did you mean something else?

Nope, underflow is the correct term

src/core/router/transports/ssu/packet.h
MoroccanMalinois: > Any thoughts?

In addition, imho, the only public member in SSUPacketParser should be ParsePacket. Ideally, parser should be destroyed just after return the newly created packet (since the header moves from the parser to packet). 
Likewise, SSUPacketBuilder::WriteSession[...] should be private or protected. 
That's why i changed a little the unit-test to use ParsePacket and WritePacket instead of ParseSessionConfirmed and WriteSessionConfirmed.

MoroccanMalinois:
Sure. Done

PR #669, opened 2017-07-07
SSU: fix and document raw packet buffer size

MoroccanMalinois:
---

---
Ref #140 

IIUC, it should be 1502  and not 1500 

Jftr, in practice, AESAlignedBuffer has room for alignment (15 bytes) so the overflow should have been extremely rare (a packet with max mtu size + max alignment).


anonimal:
JFTR: I hope with #140 we can remove the alignment question out much of the implementation. IIRC, grievances were well-voiced over IRC throughout the early stages of #140.

PR #658, opened 2017-06-17
SSU: fix SessionCreated Alice's IP

MoroccanMalinois:
---
**By submitting this pull-request, I confirm the following:**

- I have read and understood the [Contributing Guide](https://github.com/anonimal/kovri-docs/blob/master/developer/CONTRIBUTING.md).
- I have checked that another pull-request for this purpose does not exist.
- I have considered and confirmed that this submission will be valuable to others.
- I accept that this submission may not be used and that this pull-request may be closed by the will of the maintainer.
- I give this submission freely under the BSD 3-clause license.
---

https://github.com/monero-project/kovri/blob/master/src/core/router/transports/ssu/session.cc#L474 overwrites the data being pointed by https://github.com/monero-project/kovri/blob/master/src/core/router/transports/ssu/session.cc#L466, so we were sending Bob's IP instead of Alice's IP

PR #671, opened 2017-07-11
SSU: better fix for #658

MoroccanMalinois:
---

---
IMHO, i think this is a better fix for #658 "design-wise" (with the disclamer that i'm usually not very good at design)

IIUC, the idea behind the builder is that all data are already available and we only need to find pointers to these data. The problem is that the internal data of boost::asio::ip::address is not accessible, only a copy through `to_bytes().data()`. So i think, to be consistent design-wise, it is better to treat "ip" like other elements in the builder and leave the memory allocation outside. Or said otherwise, it was a bug in the usage of the class, not in the class itself ... i think ;) 

Ref #140

### DIFF
--- a/src/core/util/byte_stream.cc
+++ b/src/core/util/byte_stream.cc
@@ -158,5 +158,18 @@ const std::string GetFormattedHex(const std::uint8_t* data, std::size_t size) {
   return hex.str();
 }
 
+std::unique_ptr<std::vector<std::uint8_t>> AddressToByteVector(
+    const boost::asio::ip::address& address)
+{
+  bool is_v4(address.is_v4());
+  auto data = std::make_unique<std::vector<std::uint8_t>>(is_v4 ? 4 : 16);
+  std::memcpy(
+      data->data(),
+      is_v4 ? address.to_v4().to_bytes().data()
+            : address.to_v6().to_bytes().data(),
+      data->size());
+  return data;
+}
+
 } // namespace core
 } // namespace kovri
--- a/src/core/router/transports/ssu/packet.cc
+++ b/src/core/router/transports/ssu/packet.cc
@@ -217,13 +217,12 @@ std::uint8_t const* SSUSessionCreatedPacket::GetDhY() const {
 void SSUSessionCreatedPacket::SetIPAddress(
     std::uint8_t* address,
     std::size_t size) {
-  m_IPAddress.reset(new std::uint8_t[size]);
-  std::memcpy(m_IPAddress.get(), address, size);
+  m_IPAddress = address;
   m_AddressSize = size;
 }
 
 std::uint8_t const* SSUSessionCreatedPacket::GetIPAddress() const {
-  return m_IPAddress.get();
+  return m_IPAddress;
 }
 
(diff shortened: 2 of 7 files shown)


src/core/util/byte_stream.cc
```diff
@@ -158,5 +158,20 @@ const std::string GetFormattedHex(const std::uint8_t* data, std::size_t size) {
...
+  std::unique_ptr<std::uint8_t[]> buffer =
+      std::make_unique<std::uint8_t[]>(*size);
+  std::memcpy(
+      buffer.get(),
+      address.is_v4() ? address.to_v4().to_bytes().data()
+                      : address.to_v6().to_bytes().data(),
+      *size);
+  return buffer;
+}
+
```
anonimal: >std::unique_ptr<<std::uint8_t[]>>
>std::size_t* size)

At this point, why not refactor/implement/return a vector? Simply pass the address object, do the checks, create vector from `to_bytes().data()` (or `assign` if passing arg reference to *caller* vector object), and return the vector (*if* creating the object locally in the *callee*)? From there, `Insert` the vector data pointer and vector size. We could probably implement all of that in just a few lines (including the return) while reducing the session code by a couple of lines.

MoroccanMalinois:
Thanks @anonimal. Ready

MoroccanMalinois:
Just force pushed a mini update (start fixture variable name with m_)

src/core/util/byte_stream.cc
```diff
@@ -158,5 +158,19 @@ const std::string GetFormattedHex(const std::uint8_t* data, std::size_t size) {
   return hex.str();
 }
 
+std::unique_ptr<std::vector<std::uint8_t>> AddressToByteVector(
```
anonimal: I like this revision even though I had other thoughts in mind. Assuming we trust RVO, we may have also considered:

```cpp
std::vector<std::uint8_t> AddressToByteVector(
    const boost::asio::ip::address& address) 
{                                                   
  bool is_v4(address.is_v4());                      
  std::vector<std::uint8_t> vec(is_v4 ? 4 : 16);
  std::copy_n(
      vec.begin(),
      (is_v4 ? 4 : 16),
      (is_v4 ? address.to_v4().to_bytes().begin()
             : address.to_v6().to_bytes().begin()));
  return vec;                                       
}
```

or avoid the return:

```cpp
void AddressToByteVector(
    std::vector<std::uint8_t>& vec,
    const boost::asio::ip::address& address) 
{ 
  bool is_v4(address.is_v4());
  vec.resize((is_v4 ? 4 : 16));
  std::copy_n(
      vec.begin(), 
      (is_v4 ? 4 : 16),
      (is_v4 ? address.to_v4().to_bytes().begin()
             : address.to_v6().to_bytes().begin()));
}
```

but either of these would've required interface refactoring elsewhere (and I also prefer the move semantics of `unique_ptr`).

With that said, I think we should use our collaborated revision (https://github.com/monero-project/kovri/pull/671/commits/02ccfa87a1c996762ff752faba4794a5c3cf00b9) 👍

MoroccanMalinois:
Thanks @anonimal. Ready

PR #672, opened 2017-07-14
Util: ByteStream members init and unit-test

MoroccanMalinois:
---

---
For `Util: OutputByteStream add check for empty data` : 

I saw in [std::memcpy spec](http://en.cppreference.com/w/cpp/string/byte/memcpy) : 

    If either dest or src is a null pointer, the behavior is undefined, even if count is zero. 

But I would like to write things like

    packet->SetIPAddress(nullptr, 0)

and avoid to have to do the checks in the caller

### DIFF
--- a/src/core/util/byte_stream.cc
+++ b/src/core/util/byte_stream.cc
@@ -38,6 +38,7 @@
 #include <iomanip>
 
 #include "core/util/i2p_endian.h"
+#include "core/util/log.h"
 
 namespace kovri {
 namespace core {
@@ -100,6 +101,13 @@ void OutputByteStream::ProduceData(std::size_t amount) {
 }
 
 void OutputByteStream::WriteData(const std::uint8_t* data, std::size_t len) {
+  if (!len)
+    {
+      LOG(debug) << "OutputByteStream: skip empty data";
+      return;
+    }
+  if (!data)
+    throw std::runtime_error("OutputByteStream: null data");
   std::uint8_t* ptr = m_Data; 
   ProduceData(len);
   std::memcpy(ptr, data, len);
--- a/tests/unit_tests/core/util/byte_stream.cc
+++ b/tests/unit_tests/core/util/byte_stream.cc
@@ -33,6 +33,9 @@
 
 #include <boost/test/unit_test.hpp>
 
+#include <array>
+#include <limits>
+
 #include "core/util/byte_stream.h"
 
 namespace core = kovri::core;
@@ -40,7 +43,7 @@ namespace core = kovri::core;
 struct ByteStreamFixture
 {
   const std::string m_IPv4String = "10.11.12.13";
-  const std::array<std::uint8_t, 4> m_IPv4Array{{0x0A, 0x0B, 0x0C, 0x0D}};
+  std::array<std::uint8_t, 4> m_IPv4Array{{0x0A, 0x0B, 0x0C, 0x0D}};
 
   const std::string m_IPv6String = "fe80::42:acff:fe11:2";
   const std::array<std::uint8_t, 16> m_IPv6Array{{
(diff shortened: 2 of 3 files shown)


src/core/util/byte_stream.cc
```diff
@@ -100,6 +100,8 @@ void OutputByteStream::ProduceData(std::size_t amount) {
 }
 
 void OutputByteStream::WriteData(const std::uint8_t* data, std::size_t len) {
+  if (!len)
+    return;
```
anonimal: >But I would like to write things like
> packet->SetIPAddress(nullptr, 0)

I'm not 100% sure why'd you want to do this or why you *don't* want to throw on 0 length or `nullptr`. If we must do this, I think we should add debug logging for the odd times that we *would* want to do this. What do you think?

Also, while we're here: `if (!data || !len)`.

tests/unit_tests/core/util/byte_stream.cc
```diff
@@ -50,6 +50,93 @@ struct ByteStreamFixture
...
+{
+  std::uint64_t value = 16777216;
+  std::size_t size = sizeof(std::uint64_t);
+  std::uint8_t buffer[sizeof(std::uint64_t)];
+  core::OutputByteStream output(buffer, size);
+  BOOST_CHECK_NO_THROW(output.WriteUInt64(value));
+  BOOST_CHECK_THROW(output.WriteUInt8(0), std::length_error);
+  core::InputByteStream input(buffer, size);
+  BOOST_CHECK_EQUAL(input.ReadUInt64(), value);
+  BOOST_CHECK_THROW(input.ReadUInt8(), std::length_error);
```
anonimal: For the sake of simplicity, how about something like:

```cpp
diff --git a/tests/unit_tests/core/util/byte_stream.cc b/tests/unit_tests/core/util/byte_stream.cc
index 7981de8e..5ab88c53 100644
--- a/tests/unit_tests/core/util/byte_stream.cc
+++ b/tests/unit_tests/core/util/byte_stream.cc
@@ -33,6 +33,9 @@
 
+#include <array>
+#include <limits>
+
 #include "core/util/byte_stream.h"
 
 namespace core = kovri::core;
@@ -126,13 +129,14 @@ BOOST_AUTO_TEST_CASE(Bits32Test)
 
 BOOST_AUTO_TEST_CASE(Bits64Test)
 {
-  std::uint64_t value = 16777216;
-  std::size_t size = sizeof(std::uint64_t);
-  std::uint8_t buffer[sizeof(std::uint64_t)];
-  core::OutputByteStream output(buffer, size);
+  std::array<std::uint8_t, sizeof(std::uint64_t)> buffer {{}};
+  std::uint64_t value = std::numeric_limits<std::uint64_t>::max();
+
+  core::OutputByteStream output(buffer.data(), buffer.size());
   BOOST_CHECK_NO_THROW(output.WriteUInt64(value));
   BOOST_CHECK_THROW(output.WriteUInt8(0), std::length_error);
-  core::InputByteStream input(buffer, size);
+
+  core::InputByteStream input(buffer.data(), buffer.size());
   BOOST_CHECK_EQUAL(input.ReadUInt64(), value);
   BOOST_CHECK_THROW(input.ReadUInt8(), std::length_error);
 }
```

and refactor where appropriate for the previous cases?

tests/unit_tests/core/util/byte_stream.cc
```diff
@@ -50,6 +50,93 @@ struct ByteStreamFixture
...
+  BOOST_CHECK_THROW(input.ConsumeData(1), std::length_error);
+}
+
+BOOST_AUTO_TEST_CASE(OutputByteStream)
+{
+  std::array<std::uint8_t, 4> buffer;
+  core::OutputByteStream output(buffer.data(), buffer.size());
+  BOOST_CHECK_NO_THROW(output.WriteData(nullptr, 0));
+  BOOST_CHECK_NO_THROW(output.WriteUInt8(m_IPv4Array.at(0)));
+  BOOST_CHECK_EQUAL(output.GetSize(), 4);
```
anonimal: >4

For the sake of maintainability, how about `buffer.size()`?

tests/unit_tests/core/util/byte_stream.cc
```diff
@@ -50,6 +50,93 @@ struct ByteStreamFixture
...
+  BOOST_CHECK_EQUAL(output.GetData(), buffer.data());
+  BOOST_CHECK_EQUAL(output.GetPosition(), buffer.data() + 1);
+  BOOST_CHECK_NO_THROW(output.WriteData(&m_IPv4Array.at(1), 3));
+  BOOST_CHECK_EQUAL(output.GetPosition(), buffer.data() + 4);
```
anonimal: >4

For the sake of maintainability, how about `buffer.size()`?

src/core/util/byte_stream.cc
MoroccanMalinois: > not 100% sure why'd you want to do this or why you don't want to throw on 0 length or nullptr

Here is a sample use case : from [SSU RelayRequest spec](https://geti2p.net/spec/ssu), 

> the IP length is always 0 and the port is always 0, and the receiver should use the packet's source address and port

Or 

> Challenge is unimplemented, challenge size is always zero

IMHO, the packet builder should not care about these kind of rules. They should only be expressed in session.cc. To kinda explicitly say that it's the intended behavior, and not an oversight or whatever, i want to put something like this in session.cc

> packet.SetIPAddress(nullptr, 0);
> packet.SetChallenge(nullptr, 0);

But that would require code like 

> if(packet->GetIPAddressSize()) 
> WriteData(packet->GetIPAddress(), packet->GetIPAddressSize());
> if(packet->GetChallengeSize()) 
> WriteData(packet->GetChallenge(), packet->GetChallengeSize());

So what do you think about debug log ***if*** data null **and** len null, **otherwise** error log + throw  exception ?

anonimal: >IMHO, the packet builder should not care about these kind of rules. They should only be expressed in session.cc.

Sure, but the code you've changed is generic byte stream code that can (and will) be applied throughout the codebase - and is not specific to SSU. If you want SSU specificity, why not work on the packet builder specifically?

>So what do you think about debug log if data null and len null, otherwise error log + throw exception ?

This sounds like an anti-pattern in the making. If we write data and expect said data to *not* fulfill those parameters but said so data so happens *to* fulfill those parameters, we'll probably spend excessive time debugging instead of implementing any needed design changes. Can you think of any other use-case for this byte stream diff outside of SSU?

MoroccanMalinois: You are totally right !
Ready for review

src/core/util/byte_stream.cc
```diff
@@ -100,6 +101,11 @@ void OutputByteStream::ProduceData(std::size_t amount) {
 }
 
 void OutputByteStream::WriteData(const std::uint8_t* data, std::size_t len) {
+  if (!data || !len)
+    {
+      LOG(debug) << "OutputByteStream: skip null data or size";
+      return;
+    }
```
anonimal: >BOOST_CHECK_NO_THROW(output.WriteData(nullptr, 1));

As mentioned here https://github.com/monero-project/kovri/pull/672#discussion_r127573419, I'm having a real hard time *not* justifying we throw. Example:

```cpp
std::array<std::uint8_t, 1> arr {{}};
OutputByteStream obs(array.data(), array.size());

auto data_ptr = pointer_that_we_expect_to_NEVER_be_nullptr_but_for_some_extraneous_reason_is_unwittingly_nullptr;
obs.WriteData(data_ptr, 1);  // Doesn't throw - but we really should
```

Outside of the use-case example you proposed in the above issue comment, when would we need to *purposefully* write any non-zero length of a nullptr (it sounds crazy!). And now we'll need to do checks in *all* caller functions? I *completely* understand what you want to do with SSU but, like I mentioned in the above comment, maybe that should be done SSU-side and not with byte stream.

Am I misunderstanding your intention? If not, can we devise a less questionable solution?

MoroccanMalinois: Yeah, that was retarded, sorry. After i agreed with your idea, i just copy/pasted your previous instruction. And after that, i changed the unit-test accordingly with the code (u can never be wrong this way :man_facepalming: )

src/core/util/byte_stream.cc
```diff
@@ -100,6 +101,13 @@ void OutputByteStream::ProduceData(std::size_t amount) {
 }
 
 void OutputByteStream::WriteData(const std::uint8_t* data, std::size_t len) {
+  if (!len)
```
anonimal: I still strongly disagree with not throwing here (I'm not yet convinced as to why we should allow 0 length data) but, since you've provided debug logging, I'll pick and choose my battles 😆

PR #703, opened 2017-09-06
Testnet: add new container for build

MoroccanMalinois:
---

---

To avoid rebuilding the docker image for each git revision, the user can now choose to "Use binaries from repo" during testnet creation.

To rebuild the binaries:

    testnet.sh exec make

Or get a shell first:

    testnet.sh exec sh

and then

     make

### DIFF
--- a/contrib/docker/testnet.sh
+++ b/contrib/docker/testnet.sh
@@ -41,17 +41,14 @@ seq_start=10  # Not 0 because of port assignments, not 1 because we can't use IP
 seq_end=$((${seq_start} + 19))  # TODO(unassigned): arbitrary end amount
 sequence="seq -f "%03g" ${seq_start} ${seq_end}"
 
-#Note: this can avoid to rebuild the docker image
-#custom_build_dir="-v /home/user/kovri/build/kovri:/usr/bin/kovri -v /home/user/kovri/build/kovri-util:/usr/bin/kovri-util"
-
 reseed_file="reseed.zip"
 
 PrintUsage()
 {
-  echo "Usage: $ $0 {create|start|stop|destroy}" >&2
+  echo "Usage: $ $0 {create|start|stop|destroy|exec}" >&2
 }
 
-if [ "$#" -ne 1 ]
+if [[ $# -lt 1 ]]
 then
   PrintUsage
   exit 1
@@ -74,6 +71,7 @@ Prepare()
   # Set environment
   set_repo
   set_image
+  set_bins
   set_workspace
   set_args
   set_network
@@ -99,6 +97,22 @@ set_repo()
   if [[ ! -d $KOVRI_REPO ]]; then
     false
     catch "Kovri not found. See building instructions."
+  fi
+}
+
+set_bins()
+{
+  read_bool_input "Use binaries from repo?" KOVRI_USE_REPO_BINS ""
+
+  if [[ $KOVRI_USE_REPO_BINS = true ]];then
+    echo "Using binaries in ${KOVRI_REPO}/build"
+    mount_repo_bins="-v ${KOVRI_REPO}/build/kovri:/usr/bin/kovri \
+      -v ${KOVRI_REPO}/build/kovri-util:/usr/bin/kovri-util"
+
+    read_bool_input "Build repo binaries?" KOVRI_BUILD_REPO_BINS "Exec make release-static"
+    if [[ $KOVRI_BUILD_REPO_BINS = false ]];then
+      echo "Please ensure that the binaries are built statically if not built within a container"
+    fi
   fi
 }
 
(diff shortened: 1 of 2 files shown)


contrib/docker/testnet.sh
```diff
@@ -74,6 +72,28 @@ set_repo()
...
+      [yY])
+        KOVRI_USE_REPO_BINS=true
+        ;;
+      *)
+        KOVRI_USE_REPO_BINS=false
+      ;;
+     esac
+  fi
+
+  mount_repo_bins=""
```
anonimal: >mount_repo_bins=""

I don't think this is necessary. Did you find this necessary?

MoroccanMalinois:
In addition, in the new dockerfile, maybe we could also add gdb (i took the liberty to add iptables as my next PR will require it). 

Just an fyi, whatever the size of the image, it doesn't change the size of instances (as it uses [AUFS](https://en.wikipedia.org/wiki/Aufs)), so the impact of the docker image size on testnet is quite limited.

contrib/docker/testnet.sh
```diff
@@ -74,6 +72,28 @@ set_repo()
...
+  if [[ "$KOVRI_USE_REPO_BINS" = true ]]; then
+    mount_repo_bins="-v ${KOVRI_REPO}/build/kovri:/usr/bin/kovri -v ${KOVRI_REPO}/build/kovri-util:/usr/bin/kovri-util"
+
```
anonimal: If we're going to start bool'ing, how about we create bool vars for building? This [bool'ing] can then be applied elsewhere so we never have to enter keyboard input once they are all set.

What do you think? If we apply elsewhere, I think it should be in a separate commit.

contrib/docker/testnet.sh
```diff
@@ -101,7 +121,16 @@ set_image()
...
-  read_input "Build Kovri Docker image? [$KOVRI_IMAGE]" NULL "docker build -t $KOVRI_IMAGE $KOVRI_REPO"
+  # Select Dockerfile
+  local _default_dockerfile="Dockerfile"
+  if [[ -z $KOVRI_DOCKERFILE ]]; then
+    KOVRI_DOCKERFILE=${_default_dockerfile}
+    read_input "Change Dockerfile?: [KOVRI_DOCKERFILE=${KOVRI_DOCKERFILE}]" KOVRI_DOCKERFILE
+  fi
+
+  local _dockerfile_path="${KOVRI_REPO}/contrib/docker/${KOVRI_DOCKERFILE}"
+
```
anonimal: Same here.

contrib/docker/testnet.sh
```diff
@@ -314,6 +343,17 @@ Destroy()
...
 
+Exec()
+{
+  docker run -i -t \
+    --rm \
+    -v $KOVRI_REPO:/home/kovri/kovri \
+    -w /home/kovri/kovri \
+    $KOVRI_IMAGE \
+    $@
+  catch "Failed to run a shell"
```
anonimal: >"Failed to run a shell"

But this could be a failed docker run too?

anonimal:
Hi @MoroccanMalinois, I just saw these:

>In addition, in the new dockerfile, maybe we could also add gdb (i took the liberty to add iptables as my next PR will require it).

Ok, sounds good. Since it's dev-only, let's do what we need.

>Just an fyi, whatever the size of the image, it doesn't change the size of instances (as it uses AUFS EDIT: or OverlayFS now ), so the impact of the docker image size on testnet is quite limited.

I'm more concerned about how we manage logging. Even without debug levels, there's a lot of writing. I have some ideas on potential resolutions, will talk more in `#kovri-dev`.

MoroccanMalinois:
Thanks @anonimal. Ready for review

contrib/docker/testnet.sh
```diff
@@ -74,6 +100,19 @@ set_repo()
...
+set_bins()
+{
+  read_bool_input "Use binaries from repo?" KOVRI_USE_REPO_BINS ""
+
+  if [[ $KOVRI_USE_REPO_BINS = true ]];then
+    echo "Using binaries in ${KOVRI_REPO}/build"
+    mount_repo_bins="-v ${KOVRI_REPO}/build/kovri:/usr/bin/kovri \
+      -v ${KOVRI_REPO}/build/kovri-util:/usr/bin/kovri-util"
+
+    read_bool_input "Build repo binaries?" KOVRI_BUILD_REPO_BINS "Exec make"
```
anonimal: Re: our debugging in `#kovri-dev`, any objection to making this a `release-static` build?

contrib/docker/testnet.sh
```diff
@@ -74,6 +100,19 @@ set_repo()
   fi
 }
 
+set_bins()
+{
+  read_bool_input "Use binaries from repo?" KOVRI_USE_REPO_BINS ""
```
anonimal: Can you add a note to the question (or somewhere) about ensuring that the binaries are built statically if not built within the container?

contrib/docker/testnet.sh
```diff
@@ -40,12 +65,13 @@ Prepare()
 
   # Cleanup for new testnet
   if [[ $KOVRI_WORKSPACE || $KOVRI_NETWORK ]]; then
-    read_input "Kovri testnet environment detected. Attempt to destroy previous testnet?" NULL cleanup_testnet
+    read_bool_input "Kovri testnet environment detected. Attempt to destroy previous testnet?" KOVRI_CLEANUP cleanup_testnet
```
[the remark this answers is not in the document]
MoroccanMalinois: yes. you can set it to `true` or `false`, and the script will destroy or not without asking.
At least, it's what it's supposed to do :)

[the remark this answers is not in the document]
anonimal: >  if [ "${!2}" = true ];then
>    $3
>  fi

Sorry, old comment

PR #710, opened 2017-09-09
Testnet: create named pipes for log output + refactor RI and container creation + logic layout refactor

anonimal:
@MoroccanMalinois let's hold on merging until we figure out why docker appears to not auto-flush the pipe (docker appears to only flush once the instance is stopped).

Note: #706 confirms that auto-flushing works (at least on mainnet).

---

---

### DIFF
--- a/contrib/docker/testnet.sh
+++ b/contrib/docker/testnet.sh
@@ -376,6 +376,7 @@ create_ri()
 create_instance()
 {
   local _seq=${1}
+  local _docker_opts=${2}
 
   # Create named pipe for logging
   local _pipe="${KOVRI_WORKSPACE}/${kovri_base_name}${_seq}/${pipe_base_name}"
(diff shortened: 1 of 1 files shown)


MoroccanMalinois:
@anonimal  I don't think it's docker related. I'm getting the same behavior directly on the host.
Maybe i'm missing something. Here is how i tested:

    mkfifo test
    tail -f test

On another terminal

    cat LICENSE.md > test

The first time i run `cat`, i only see the last ~10 lines.
All the next time, i see all the file. 
Same for testnet, after a second start, logs are auto flushed (assuming `tail` stayed open between the runs).

anonimal:
>I'm getting the same behavior directly on the host.

What's your host? I can only reproduce your issue on FreeBSD and MSYS2 (MSYS2 looks stupidly broken); I can't reproduce your issue on Arch, Ubuntu, or OSX (kovri auto-flushing to pipe also works as expected on these systems).

On my Arch docker host, creating/tailing the pipe + cat'ing big file into the pipe *within* the container works as well.  I don't think docker is completely innocent here.

MoroccanMalinois:
My host is a debian jessie (not updated). Going to install and test a fresh stretch.

MoroccanMalinois:
Tested with debian stretch, ubuntu 16.04 and (fedora-23 in Qubes). All the same.

anonimal:
@MoroccanMalinois then let's not use tail. `cat`'ing the pipe works fine (I imagine that tail is blocking). Have you tried `s/tail -f/cat/` on your affected systems?

anonimal:
Note: rebased some trivial changes not related to this PR discussion, also added 34a55a5 to save opening a new PR.

PR #716, opened 2017-09-15
README: remove mastodon.social point of contact

anonimal:
RFC. Do we really need this? I think not.

The service appears useless. No better than GNU social - and has tighter speech restrictions.

Also, `@getkovri` on twitter gets 45 followers in just a couple of days but `@kovri@mastadon.social` only gets 2 (one a community member and the other appears to be a bot). Not to mention, mastodon keeps spamming my bitlbee with posts from instance admins of whom I'm not following.

In all, not worth maintaining IMHO but we can keep as 'reserved' account.

Referencing https://github.com/monero-project/meta/issues/113.

---

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,6 @@ Alpha release coming soon
 
 ## Contact
 - Twitter: [@getkovri](https://twitter.com/getkovri)
-- Mastodon: [@kovri@mastodon.social](https://mastodon.social)
 - Email:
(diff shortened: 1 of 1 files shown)


anonimal:
Note: apparently, if deleting the account: "username will remain reserved to prevent future impersonations."

anonimal:
Account deleted.

PR #723, opened 2017-09-23
Testnet: implement webserver for subscription publishing

### LINKED ISSUE #718: Testnet: internal HTTP/S webserver for publishing subscription
Opposing approaches to a resolution:

1.
```
@anonimal cpp-netlib has an http/s server API. It seems trivial to implement.
@anonimal We can run a testnet-only server to host as an addressbook publisher.
@anonimal This way we don't do anything crazy like try to setup a webserver within docker for something this small.
@anonimal I say this because right now we are polling downloads.getmonero.org for our default subscription,
@anonimal and, even within testnet, these requests are going live onto the internet so, Nth requests to that host at the same == not very nice.
@anonimal The host is also hardcoded so I think it's best to take advantage of creating a realistic environment rather than having everything be customized.
@anonimal (when we want to change publisher to https://127.0.0.1, we simply > the new host into publishers.txt after we copy resources to workspace. this will override hardcoded value)
```

2.
```
@anonimal Hmm, on 2nd thought, MoroccanMalinoi-,
@anonimal but easier to simply setup a testnet network webserver? Probably, and we would avoid furthering our dependency upon a library dependency.
@anonimal We would just have to adjust the webserver config after we've acquired appropriate network octets, then copy over, etc.
@anonimal The C++ in me wants to get away from Docker but the Docker in me wants to get away from C++ for this task ;)
```

Both have selling points, but which has more virtue? Note: a WebUI is not a virtue, so please don't.

anonimal:
Also implements kovri `--enable-ssl` option.

Resolves #718.

---

---

### DIFF
--- a/contrib/docker/Dockerfile.dev.apache
+++ b/contrib/docker/Dockerfile.dev.apache
@@ -1,0 +1,35 @@
+# Copyright (c) 2015-2017, The Kovri I2P Router Project
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are
+# permitted provided that the following conditions are met:
+#
+# 1. Redistributions of source code must retain the above copyright notice, this list of
+#    conditions and the following disclaimer.
+#
+# 2. Redistributions in binary form must reproduce the above copyright notice, this list
+#    of conditions and the following disclaimer in the documentation and/or other
+#    materials provided with the distribution.
+#
+# 3. Neither the name of the copyright holder nor the names of its contributors may be
+#    used to endorse or promote products derived from this software without specific
+#    prior written permission.
+#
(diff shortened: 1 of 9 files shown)


anonimal:
@MoroccanMalinois I don't like having to add an httpd config file. I'm also questioning the Apache License and if that license applies to a config file (I added it at the top).

Because containers aren't intended to be static, is there anyway we can sed `ServerName` within the httpd container at run-time *before* httpd is started? If so, I think we should do that so we can drop this in-tree config.

anonimal:
Working on an entrypoint to replace the conf file.

anonimal:
Revised/rebased. Thanks to @MoroccanMalinois for entrypoint tips in `#kovri-dev`.

anonimal:
Some commits were accidentally rebased instead of stashed. Everything's cleared up now, good to go.

Issue #534, opened 2017-01-24
Log: cleanup formatting


anonimal:
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

*Place an X inside the bracket to confirm*
- [X] I confirm.

---

c8870c0905cf79b35a4d9fd667a5e64af3cad45c allows us to do more with our logger. The current formatting is annoying, and we can also benefit from some colored output (ideally, we autodetect color capable terminals if boost doesn't do this by default). Also note we need to *not* output color codes to file sink or to any other future sink/channel that doesn't support color codes.

anonimal:
```
selsta  so aligned [log level] next?
@anonimal       Sure, that would be neat. Formatter does it by default but apparently is overwritten/removed when we use custom formatter?
```

anonimal:
Resolved in https://github.com/monero-project/kovri/pull/731.

PR #729, opened 2017-09-26
Log: add colored severity levels

selsta:
---

---

### DIFF
--- a/src/core/util/log.cc
+++ b/src/core/util/log.cc
@@ -41,6 +41,7 @@
 #include <boost/log/sources/severity_channel_logger.hpp>
 #include <boost/log/support/date_time.hpp>
 #include <boost/make_shared.hpp>
+#include <boost/phoenix/bind.hpp>
 #include <boost/property_tree/ini_parser.hpp>
 #include <boost/property_tree/ptree.hpp>
 #include <boost/shared_ptr.hpp>
@@ -64,12 +65,63 @@ namespace kovri
 {
 namespace core
 {
+namespace logging = boost::log;
+namespace expr = boost::log::expressions;
+namespace attrs = boost::log::attributes;
+
+std::string SetColor(
+    logging::value_ref<
+        logging::trivial::severity_level,
+        logging::trivial::tag::severity> const& severity,
+    bool has_color)
+{
+  if (has_color)
+    {
+      if (severity)
+        {
+          switch (severity.get())
+            {
+              case logging::trivial::fatal:
+              case logging::trivial::error:
+                return "\033[31m";  // Red
+              case logging::trivial::info:
+                return "\033[32m";  // Green
+              case logging::trivial::warning:
+                return "\033[33m";  // Yellow
+              case logging::trivial::debug:
+                return "\033[34m";  // Blue
+              case logging::trivial::trace:
+                return "\033[36m";  // Cyan
+            }
+        }
+      return "\033[0m";  // Reset
+    }
+  return "";
+}
+
+logging::formatter GetFormat(bool has_color)
+{
+  logging::formatter format =
+      expr::stream
+      << boost::phoenix::bind(  // TODO(unassigned): replace with std::bind
+             &SetColor,
+             logging::trivial::severity.or_none(),
+             has_color)
+      << "["
+      << expr::format_date_time(
+             expr::attr<boost::posix_time::ptime>("TimeStamp"),
+             "%Y.%m.%d %T.%f")
+      << "]"
+      << " [" << expr::attr<attrs::current_thread_id::value_type>("ThreadID")
+      << "]"
+      << " [" << logging::trivial::severity << "]"
+      << "  " << expr::smessage << (has_color ? "\033[0m" : "");  // Reset
+  return format;
+}
+
 void SetupLogging(const boost::program_options::variables_map& kovri_config)
 {
-  namespace logging = boost::log;
-  namespace expr = boost::log::expressions;
   namespace sinks = boost::log::sinks;
-  namespace attrs = boost::log::attributes;
   namespace keywords = boost::log::keywords;
   // Get global logger
   // TODO(unassigned): depends on global logging initialization. See notes in log impl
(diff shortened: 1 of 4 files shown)


src/core/util/log.cc
```diff
@@ -131,6 +156,19 @@ void SetupLogging(const boost::program_options::variables_map& kovri_config)
...
+                          expr::attr<boost::posix_time::ptime>("TimeStamp"),
+                          "%Y.%m.%d %T.%f")
+                   << "]"
+                   << " [" << expr::attr<attrs::current_thread_id::value_type>(
+                                  "ThreadID")
+                   << "]"
+                   << " [" << logging::trivial::severity << "]"
+                   << "  " << expr::smessage
+                   << "\033[0m"; // Reset
   logging::formatter format =
```
anonimal: To eliminate the copypasta, one possible solution is to abstract the formatting process. Perhaps with a template function `template<typename Sink> SetFormatter(Sink& sink, bool is_file)` or local Boost.Variant object. Ideally, with either, the objective would be to test the sink object attributes (e.g., is it file or text sink?) internally so passing a bool would unnecessary.

There are probably more efficient solutions available though I don't know of any Boost.Log solutions off the top of my head (could be in the docs somewhere though).

src/core/util/log.cc
```diff
@@ -41,6 +41,7 @@
 #include <boost/log/sources/severity_channel_logger.hpp>
 #include <boost/log/support/date_time.hpp>
 #include <boost/make_shared.hpp>
+#include <boost/phoenix.hpp>
```
anonimal: As discussed in IRC, `boost::phoenix::bind` is a "lazy" evaluator so it apparently works well with boost expressions. We should aim for a `std::bind` or lambda replacement + refactor though (if possible) in order to reduce build overhead.

src/core/util/log.cc
```diff
@@ -131,6 +156,19 @@ void SetupLogging(const boost::program_options::variables_map& kovri_config)
   auto file_sink = boost::shared_ptr<text_file_sink>(
       std::make_unique<text_file_sink>(file_backend));
   // Set sink formatting
+  logging::formatter color_format =
```
anonimal: Ideally, in a separate commit, we provide a config option to disable color log `--log-enable-color` which would default to true (so, enabled by default). If there is ample evidence that boost is smart enough to deal with non-color terminals, then nevermind.

anonimal:
This should resolve https://github.com/monero-project/kovri/issues/534 though we should try to get the log alignment fixed (between severity and log text) before closing that issue.

src/core/util/log.cc
selsta: I think these color escape codes are not boost related.

src/core/util/log.cc
```diff
@@ -64,12 +65,62 @@ namespace kovri
...
+                          "%Y.%m.%d %T.%f")
+                   << "]"
+                   << " [" << expr::attr<attrs::current_thread_id::value_type>(
+                                  "ThreadID")
+                   << "]"
+                   << " [" << logging::trivial::severity << "]"
+                   << "  " << expr::smessage
+                   << (color == true ? "\033[0m" : ""); // Reset
+
+  return fmt;
```
anonimal: As noted in our [style guide](https://github.com/monero-project/kovri-docs/blob/master/i18n/en/style.md), will you please consider the following?:

- running `clang-format` on all of your two new functions
- as noted in [Naming](https://google.github.io/styleguide/cppguide.html#Naming), `color` and `fmt` could be more descriptive (e.g., `has_color` and `format`)

Also, including this PR number in the commit log (e.g., `Referencing #729` or similar).

src/core/util/log.cc
```diff
@@ -64,12 +65,62 @@ namespace kovri
...
+        return "\033[31m"; // Red
+      case logging::trivial::info:
+        return "\033[32m"; // Green
+      case logging::trivial::warning:
+        return "\033[33m"; // Yellow
+      case logging::trivial::debug:
+        return "\033[34m"; // Blue
+      case logging::trivial::trace:
+        return "\033[36m"; // Cyan
+    }
```
anonimal: This could be simplified with:

```diff
diff --git a/src/core/util/log.cc b/src/core/util/log.cc
index 1aba1994..2dfce01a 100644
--- a/src/core/util/log.cc
+++ b/src/core/util/log.cc
@@ -74,28 +74,28 @@ std::string SetColor(
     logging::value_ref<logging::trivial::severity_level,
     logging::trivial::tag::severity> const& severity, bool color)
 {
-  if (!color)
+  if (color)
   {
-    return "";
-  }
-  else if (severity)
-  {
-    switch (severity.get())
+    if (severity)
     {
-      case logging::trivial::fatal:
-      case logging::trivial::error:
-        return "\033[31m"; // Red
-      case logging::trivial::info:
-        return "\033[32m"; // Green
-      case logging::trivial::warning:
-        return "\033[33m"; // Yellow
-      case logging::trivial::debug:
-        return "\033[34m"; // Blue
-      case logging::trivial::trace:
-        return "\033[36m"; // Cyan
+      switch (severity.get())
+      {
+        case logging::trivial::fatal:
+        case logging::trivial::error:
+          return "\033[31m"; // Red
+        case logging::trivial::info:
+          return "\033[32m"; // Green
+        case logging::trivial::warning:
+          return "\033[33m"; // Yellow
+        case logging::trivial::debug:
+          return "\033[34m"; // Blue
+        case logging::trivial::trace:
+          return "\033[36m"; // Cyan
+      }
     }
+    return "\033[0m"; // Reset
   }
-  return "\033[0m"; // Reset
+  return "";
 }
 
 logging::formatter GetFormat(bool color)
```

src/core/util/log.cc
```diff
@@ -64,12 +65,62 @@ namespace kovri
...
+      case logging::trivial::error:
+        return "\033[31m"; // Red
+      case logging::trivial::info:
+        return "\033[32m"; // Green
+      case logging::trivial::warning:
+        return "\033[33m"; // Yellow
+      case logging::trivial::debug:
+        return "\033[34m"; // Blue
+      case logging::trivial::trace:
+        return "\033[36m"; // Cyan
```
anonimal: Note: I haven't tested on non-bash yet, but I imagine these escape codes only apply to bash? We should implement `--log-disable-color` for non-bash users if this indeed rings true.

selsta: https://stackoverflow.com/a/2616912 according to him garbage will show up when colour sequences are not supported.

selsta: Added --log-disable-color

src/core/util/log.cc
selsta: Improved variable names and ran `clang-format` on my functions.

>Also, including this PR number in the commit log (e.g., Referencing #729 or similar).

Asked you about that on IRC.

src/core/util/log.cc
selsta: Applied, thanks

src/core/util/log.cc
selsta: I changed <boost/phoenix.hpp> to <boost/phoenix/bind.hpp>. Does that reduce build overhead?

(Asked that on IRC too.)

src/core/util/log.cc
```diff
@@ -131,19 +183,9 @@ void SetupLogging(const boost::program_options::variables_map& kovri_config)
...
-                   << "]"
-                   << " [" << expr::attr<attrs::current_thread_id::value_type>(
-                                  "ThreadID")
-                   << "]"
-                   << " [" << logging::trivial::severity << "]"
-                   << "  " << expr::smessage;
-  text_sink->set_formatter(format);
-  file_sink->set_formatter(format);
+  text_sink->set_formatter(
+      GetFormat(true && !kovri_config["log-disable-color"].as<bool>()));
```
anonimal: I think this line points out a problem elsewhere. We should have enabled-by-default options (I should've recommended that before, sorry). If we simply refactor to `--log-enable-color` and set default to true, then we can `GetFormat(kovri_config["log-disable-color"].as<bool>()`.  What do you think?

PR #735, opened 2017-10-13
kovri-util: add cpuid func

selsta:
TODO future:

-  Better error handling. Currently I’m returning zeros and empty strings when CpuId() fails. I think it would be better to check for CpuId() in the beginning and write an error if it fails.

---

---

### DIFF
--- a/src/util/cpuid.cc
+++ b/src/util/cpuid.cc
@@ -1,0 +1,216 @@
+/**                                                                                           //
+ * Copyright (c) 2015-2017, The Kovri I2P Router Project                                      //
+ *                                                                                            //
+ * All rights reserved.                                                                       //
+ *                                                                                            //
+ * Redistribution and use in source and binary forms, with or without modification, are       //
+ * permitted provided that the following conditions are met:                                  //
+ *                                                                                            //
+ * 1. Redistributions of source code must retain the above copyright notice, this list of     //
+ *    conditions and the following disclaimer.                                                //
+ *                                                                                            //
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this list     //
+ *    of conditions and the following disclaimer in the documentation and/or other            //
+ *    materials provided with the distribution.                                               //
+ *                                                                                            //
+ * 3. Neither the name of the copyright holder nor the names of its contributors may be       //
+ *    used to endorse or promote products derived from this software without specific         //
+ *    prior written permission.                                                               //
+ *                                                                                            //
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY        //
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF    //
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL     //
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,       //
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,               //
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS    //
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,          //
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF    //
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.               //
+ */
+
+#include "util/cpuid.h"
+
+#ifdef WITH_CRYPTOPP
+#include <cryptopp/cpu.h>
+#endif
+
+#include "core/util/log.h"
+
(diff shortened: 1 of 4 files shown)


src/util/cpuid.cc
```diff
@@ -0,0 +1,218 @@
...
+#include "core/util/log.h"
+
+namespace bpo = boost::program_options;
+
+// Cpuid command
+CpuidCommand::CpuidCommand() {}
+
+std::string CpuidCommand::GetCpuVendor()
+{
+  CryptoPP::word32 cpuid[4] = {0};
```
anonimal: There's no conditional preprocessor for every instance of `word32`. These should simply be refactored to `unsigned int` to avoid breaking a build. Also note that `cpuid[4] = {0}` is an unnecessary syntax for initialization in c++11 (`cpuid[4]{}` will suffice). cryptopp maintains c03 compatibility but we aim for c++14 compliance, so feel free to use the perks of c++11/14.

src/util/cpuid.cc
```diff
@@ -0,0 +1,218 @@
...
+          cpu_model_name << (char)(cpuid[j] & 0xFF);
+          cpu_model_name << (char)(cpuid[j] >> 8 & 0xFF);
+          cpu_model_name << (char)(cpuid[j] >> 16 & 0xFF);
+          cpu_model_name << (char)(cpuid[j] >> 24 & 0xFF);
+        }
+    }
+  return cpu_model_name.str();
+}
+
+int CpuidCommand::GetCpuFamily()
```
anonimal: >`int`

You'll never return a signed type with these functions.

src/util/cpuid.cc
```diff
@@ -0,0 +1,218 @@
...
+    }
+  if (cpu_family == 15)
+    {
+      cpu_family += GetCpuExtendedFamily();
+    }
+  LOG(info) << "CPU Family: " << cpu_family;
+  LOG(info) << "CPU Model: " << cpu_model;
+  LOG(info) << "CPU Stepping: " << GetCpuStepping();
+  LOG(info) << "CacheLineSize: " << CryptoPP::GetCacheLineSize();
+  LOG(info) << "Has SSE2?: " << CryptoPP::HasSSE2();
```
anonimal: The question marks aren't needed.

src/util/cpuid.cc
```diff
@@ -0,0 +1,218 @@
...
+{
+#ifdef WITH_CRYPTOPP
+#if defined(__x86_64__) \
+    || defined(_M_X64)  // TODO(selsta): _M_IX86 and _X86_ ??
+  LOG(info) << "CPU Vendor: " << GetCpuVendor();
+  LOG(info) << "CPU Model Name: " << GetCpuModelName();
+  int cpu_family = GetCpuFamily();
+  int cpu_model = GetCpuModel();
+  if (cpu_family == 15
+      || (cpu_family == 6 && GetCpuVendor().compare("GenuineIntel") == 0))
```
anonimal: >15
>6

Constants should be `const` or `constexpr` variables with human-meaningful names instead of raw values.

src/util/cpuid.cc
```diff
@@ -0,0 +1,218 @@
...
+
+#include "util/cpuid.h"
+
+#ifdef WITH_CRYPTOPP
+#include <cryptopp/cpu.h>
+#endif
+
+#include "core/util/log.h"
+
+namespace bpo = boost::program_options;
```
anonimal: Unused.

anonimal:
The ARM builds should be fixed too before this is merged.

anonimal:
Referencing https://github.com/weidai11/cryptopp/issues/521.

anonimal:
#737 has been merged.

selsta:
Added your suggestions

src/util/cpuid.cc
```diff
@@ -0,0 +1,220 @@
...
+#include <cryptopp/cpu.h>
+#endif
+
+#include "core/util/log.h"
+
+// Cpuid command
+CpuidCommand::CpuidCommand() {}
+
+#if defined(__x86_64__) \
+    || defined(_M_X64)  // TODO(selsta): _M_IX86 and _X86_ ??
```
anonimal: When you're fixing this for x86, use `defined(__i386__)`. Also we don't need `_M_X64` since no devs are using MSVC (I had left that in the AES-NI code in the rare chance we'd pickup a MSVC-using dev by now but we haven't so it's no longer applicable).

selsta:
Anything still missing?

anonimal:
Reference this PR number in the git log.

selsta:
done

src/util/cpuid.cc
```diff
@@ -0,0 +1,218 @@
...
+  LOG(info) << "Has PadlockPMM: " << CryptoPP::HasPadlockPMM();
+#elif defined(__arm__)
+  LOG(info) << "Has NEON: " << CryptoPP::HasNEON();
+  LOG(info) << "Has PMULL: " << CryptoPP::HasPMULL();
+  LOG(info) << "Has CRC32: " << CryptoPP::HasCRC32();
+  LOG(info) << "Has AES: " << CryptoPP::HasAES();
+  LOG(info) << "Has SHA1: " << CryptoPP::HasSHA1();
+  LOG(info) << "Has SHA2: " << CryptoPP::HasSHA2();
+#endif
+  return true;
```
anonimal: LOG that other processors are unsupported.

src/util/cpuid.cc
```diff
@@ -0,0 +1,218 @@
...
+#endif
+  LOG(warning) << "cpuid requires Crypto++";
```
anonimal: Will there be a non-cryotopp implementation? If not, maybe we should reconsider how we treat pre-processing and simply leave out the command option all-together if cryptopp is not available.

anonimal: Note: doing so would cut down on the number of directives needed and keep things generally cleaner.

selsta:
addressed both comments.

anonimal:
Referencing #399.

Issue #498, opened 2016-12-11
Remove support for all signatures except Ed25519 + keep RSA-SHA512 raw verifier for reseed


anonimal:
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

*Place an X inside the bracket to confirm*
- [X] I confirm.

---

As discussed in today's meeting (and earlier), the general consensus is "yes".

References #485 #345 #369 #160.

anonimal:
Note: will remove:
- DSA-SHA1
- ECDSA-SHA256-P256
- ECDSA-SHA384-P384
- ECDSA-SHA512-P521
- RSA-SHA256-2048
- RSA-SHA384-3072

anonimal:
NOTICE: THIS ISSUE HAS BEEN MOVED TO GitLab. Please continue the discussion there. See https://github.com/monero-project/kovri/issues/1013 for details.

Issue #643, opened 2017-05-19
NetDb: rewrite/refactor


anonimal:
Not nearly as bad of shape as #627 was and as #511 currently is, the network-database has always needed work (but this issue is mostly for housekeeping).

Since NetDb/RI/I2NP/Tunnels/Transports are all related (too closely related in our implementation), any work here will help clarify any questions / resolve any potential issues related to #187.

---
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.
---

anonimal:
NOTICE: THIS ISSUE HAS BEEN MOVED TO GitLab. Please continue the discussion there. See https://github.com/monero-project/kovri/issues/1013 for details.

Issue #366, opened 2016-09-16
Rewrite Identity implementation


anonimal:
**By submitting this issue, I confirm the following:**
- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

_Place an X inside the bracket to confirm_
- [X] I confirm.

---

If I had to vote for one of the most questionably dangerous, ugliest, least functional, and least maintainable areas of this codebase, I could give a strong vote for our current identity implementation. It's error prone and confusing (to say the least). It's not a lot of code but it's important code - so it should be solid code (as should the other areas that tie into it).

Noticed especially during #364, we can put this issue toward a later release milestone but ideas for improvements in the meantime are welcome. One solution includes over-arching libcore design refactor but let's not get carried away (for now) 😄

anonimal:
Github closed this because, in a comment, I said the word "resolve" before #366? 😆

anonimal:
NOTICE: THIS ISSUE HAS BEEN MOVED TO GitLab. Please continue the discussion there. See https://github.com/monero-project/kovri/issues/1013 for details.

PR #755, opened 2017-11-20
Remove unnecessary signature algorithms

coneiric:
---

---

Remove signatures according to #498


src/core/router/info.h
```diff
@@ -60,8 +60,8 @@ struct RouterInfoTraits
   /// @brief Router Info size constants
   enum Size : std::uint16_t
   {
-    MinBuffer = core::DSA_SIGNATURE_LENGTH,  // TODO(unassigned): see #498
-    MaxBuffer = 2048,  // TODO(anonimal): review if arbitrary
+    MinBuffer = core::EDDSA25519_SIGNATURE_LENGTH,  // TODO(unassigned): see #498
+    MaxBuffer = core::RSASHA5124096_KEY_LENGTH,  // TODO(anonimal): review if arbitrary
```
anonimal: >MaxBuffer = core::RSASHA5124096_KEY_LENGTH

This quickly breaks functionality. See the use-case for MaxBuffer.

coneiric: I didn't realize this breaks functionality. I'll revert MaxBuffer, and do some reading. Is the MinBuffer value still fine?

coneiric:
I haven't checked with a full live test yet. Was able to start Kovri, and got it to connect to peers. Will test out if tunnel creation works.

Any tips for writing integration tests?

anonimal:
>Any tips for writing integration tests for Kovri?

Aside from the prelim Boost.Python work and testnet, there's nothing currently in-place aside from running the router and testing tunnels manually.

coneiric:
Just ran into a connection bug with a pull from the latest master. Thought it might have been something from my changes, so cloned a fresh copy, and same bug came up with building tunnels.

A few different errors keep popping up, (some rejections, timeouts, missing parts, etc.). The project builds fine, and runs without crashing.

Here is a gist of the [error logs](https://gist.github.com/coneiric/b3023dac2ad6d173342b35ec704c8a74)

Should I open an issue for this?

anonimal:
>Just ran into a connection bug with a pull from the latest master

Those are all known and actual error reporting needs fine-tuning (tunnel building works regardless). Not related to this PR, no need for a new issue.

anonimal:
```
@anonimal       oneiric: I'm doing a hefty amount of related review + related work on your PR. I'll end up committing something to your branch instead of sending diffs.
@anonimal       Essentially one of the problems is how we don't properly handle silent fails (somewhat silent) when encountering older routers in our netdb + when we want to build tunnels.
@anonimal       These older routers are signing with DSA-SHA1 and we never remove them from the netdb once parsed/created.
@anonimal       That and more. I've patched the other areas you missed but there's still more to do and I haven't caught it all yet.
@anonimal       This work would've been easier to finish if that identity rewrite was resolved first. Oh well, here we are.
@anonimal       s/how we don't/how we don't with this PR/
@anonimal       I've added much needed trace logging too, really makes things easier.
@anonimal       Dumping the RIs in hex and running kovri-util should be reduced to simply running GetDescription during runtime. I'll make the changes and more.
@anonimal       Anyway, next week (Monday).
```

The above reference was regarding nightmare code #366.

anonimal:
An update/clarifications:

1. I've put this PR on the back-burner and will probably continue to do so for the rest of the month because its resolution is not a requisite for release. We'll see.
2. After reseed, most (but not all) of the time, we remove NetDb RI entries once they are confirmed DSA (set as unreachable, the NetDb eventually removes) but even when they are successfully removed, this isn't done quickly enough (saved ~1 minute later, deleted ~3/4 minutes after that) so, transports attempt to connect anyway - later failing during signature verification.
3. When both client / server tunnel Ed25519 keys are in use, and we attempt to use tunnels, there are silent failures that need to be logged and fixed
4. When server tunnel keys are DSA, obvious errors:
   - When HandlingDatabaseStoreMessage / creating remote LeaseSet from DatabaseStore, lease(s) is(are) DSA (keytype 0). We don't handle appropriately in LeaseSet's ReadFromBuffer (patched to catch, but not PR'd; if any lease is DSA, the entire leaseset is null - this is not a real solution), nor elsewhere (not patched).
     - Other non-Ed25519 behavior is TBD but it should be behave similar so long as the aforementioned isn't fixed.

coneiric:
Thanks for your notes and work on this PR. Apologies for lack of reply, was unable to access my accounts.

> I've put this PR on the back-burner

Totally understand. I'll keep working on the points you brought up, hopefully ready by release after this one :)

> After reseed, most (but not all) of the time, we remove NetDb RI entries

Will look for areas in code to shorten the gap between confirming DSA reseed entries. Is there a minimum amount of time the entries need to stay in the NetDb?

> When both client / server tunnel Ed25519 keys are in use ... there are silent failures that need to be logged and fixed

Will look into sections of the code dealing with this scenario, and get an understanding of what those silent failures are.

> When HandlingDatabaseStoreMessage / creating remote LeaseSet from DatabaseStore, lease(s) is(are) DSA (keytype 0) ... if any lease is DSA, the entire leaseset is null - this is not a real solution)

Is it possible to separate the DSA entry from the rest of the LeaseSet? Does one member rely on all others in the set? Like in the catch scenario, was thinking about carving out the DSA entry(s), and replacing them with valid entries from elsewhere / doing a DatabaseStore reorg to only include ed25519 in LeaseSet. Not sure how that would work in code, or if it is even a good idea on the conceptual level.

Many thanks for the pointers on where to go from here.

anonimal:
>Is it possible to separate the DSA entry from the rest of the LeaseSet?

IIRC, should be if implemented.

>Like in the catch scenario, was thinking about carving out the DSA entry(s), and replacing them with valid entries from elsewhere / doing a DatabaseStore reorg to only include ed25519 in LeaseSet.

Ok. Let's see what you come up with.

src/core/router/identity.cc
```diff
@@ -435,60 +354,25 @@ SigningKeyType IdentityEx::GetSigningKeyType() const {
...
 
 CryptoKeyType IdentityEx::GetCryptoKeyType() const {
   if (m_StandardIdentity.certificate.type ==
       CERTIFICATE_TYPE_KEY && m_ExtendedBuffer)
     return bufbe16toh(m_ExtendedBuffer.get() + 2);  // crypto key
-  return CRYPTO_KEY_TYPE_ELGAMAL;
+  // TODO(unassigned): should an exception be raised here?
+  // since an unsupported setup is used when the above condition is false?
+  std::unique_ptr<std::uint8_t[]> null_key = std::make_unique<std::uint8_t[]>(m_ExtendedLen - 2);
+  return bufbe16toh(null_key.get());  // null crypto key
```
anonimal: >Likely not the correct solution, only fixes SIGILL runtime
interrupt.
>std::unique_ptr<std::uint8_t[]> null_key = std::make_unique<std::uint8_t[]>(m_ExtendedLen - 2);
>return bufbe16toh(null_key.get());  // null crypto key

Not a real fix. If you drop this commit and rebase the branch, I can start sending in the patches I mentioned earlier.

src/core/router/identity.cc
```diff
@@ -435,60 +354,25 @@ SigningKeyType IdentityEx::GetSigningKeyType() const {
   if (m_StandardIdentity.certificate.type ==
       CERTIFICATE_TYPE_KEY && m_ExtendedBuffer)
     return bufbe16toh(m_ExtendedBuffer.get());  // signing key
-  return SIGNING_KEY_TYPE_DSA_SHA1;
+  // TODO(unassigned): should an exception be raised here?
+  // since an unsupported setup is used when the above condition is false?
+  std::unique_ptr<std::uint8_t[]> null_key = std::make_unique<std::uint8_t[]>(m_ExtendedLen);
+  return bufbe16toh(null_key.get());  // signing key
```
anonimal: >Likely not the correct solution, only fixes SIGILL runtime
interrupt.
>std::unique_ptr<std::uint8_t[]> null_key = std::make_unique<std::uint8_t[]>(m_ExtendedLen);
>return bufbe16toh(null_key.get());  // signing key

Not a real fix. If you drop this commit and rebase the branch, I can start sending in the patches I mentioned earlier.

coneiric: I can drop that commit no problem, where should I rebase the branch to? Just don't want all the comments to get deleted, unless that's not a problem.

coneiric:
wtf, just tried to reset the last two commits, and Github auto-close the PR...

anonimal:
Hmm, can you reopen?

coneiric:
It says there are no commits. I'll try adding something simple, then reopen

anonimal:
>Is there a minimum amount of time the entries need to stay in the NetDb?

@coneiric I missed this question. See `Time::RouterExpiration` and its implementation (both are in `src/core/router/net_db`). Also see the [docs](https://geti2p.net/en/docs/how/network-database) for details. Also note LeaseSet behavior and expiration time.

anonimal:
#498 will be vastly easier to resolve once #366 and #643 and resolved.

coneiric:
Closing until #366 and #643 are resolved.

PR #761, opened 2017-12-01
Core: replace stray cout in config

selsta:
`[2017-11-04 01:22:00] <anonimal> selsta: it was probably just missed, I don't know`

---

---

### DIFF
--- a/src/core/util/config.cc
+++ b/src/core/util/config.cc
@@ -161,7 +161,7 @@ void Configuration::ParseConfig()
   // Help options
   if (m_Map.count("help"))
     {
-      std::cout << config_options << std::endl;
+      LOG(info) << config_options;
       throw std::runtime_error(
           "for more details, see user-guide or config file");
     }


src/core/util/config.cc
```diff
@@ -161,7 +161,9 @@ void Configuration::ParseConfig()
   // Help options
   if (m_Map.count("help"))
     {
-      std::cout << config_options << std::endl;
+      // Logging is not setup yet
```
anonimal: Comments should describe what's being done or will be done, not what's *not* being done.

src/core/util/config.cc
```diff
@@ -161,7 +161,9 @@ void Configuration::ParseConfig()
...
+      core::SetupLogging(m_Map);
```
anonimal: Another option is to move logging initialization from core instance initialization into configuration initialization (e.g., in this function right after the map is notified). This way we only need to call the function once. An argument against doing so would be that the logging setup is then not decoupled from config parsing - but initialization requires a parsed map and per-instance configurations would require per-logging configurations anyway (eventually, when we're not using the global logging singleton).

src/core/util/config.cc
selsta: I always write comments to why I’m doing things.. what’s being done can be read from the code anyway?

src/core/util/config.cc
selsta: Changed it as you suggested.

src/core/instance.cc
```diff
@@ -56,10 +56,6 @@ Instance::Instance(const std::vector<std::string>& args) try
     : m_Exception(__func__),
       m_Config(args)
   {
-    // TODO(anonimal): Initializing of sources/streams/sinks must come after we've properly configured the logger.
-    //   we do this here so we can catch debug logging before instance "initialization". This is not ideal
```
anonimal: Please don't remove this TODO. It is not yet resolved. You can move it along with the function.

src/core/util/config.cc
anonimal: >I always write comments to why I’m doing things

Well it's good to get into better habits then. And if you're going to describe why, then describe why. Obviously "logging is not setup yet" but that doesn't explain *why* you need to setup logging.

selsta: Yea, that comment could have been better.

selsta:
sorry, fixed both things

src/core/util/config.cc
```diff
@@ -158,10 +158,14 @@ void Configuration::ParseConfig()
   bpo::store(
       bpo::command_line_parser(m_Args).options(cli_options).run(), m_Map);
   bpo::notify(m_Map);
+
+  // TODO(anonimal): Initializing of sources/streams/sinks must come after we've properly configured the logger.
+  //   we do this here so we can catch debug logging before instance "initialization". This is not ideal
+  core::SetupLogging(m_Map);
```
anonimal: Oops! Slight problem: now log options won't be set by config file. My fault! I should've looked before commenting.

We have two options now:
1. Setup logging after we parse config file *then* run help. This is a waste of cycles when running help.
2. Revert all this work and keep your original PR (minus the comment and call to SetupLogging). If we don't initialize custom logging, boost.log will use perfectly fine defaults. So, unless you want a colorful help section, just use boost.log defaults by calling LOG without SetupLogging.

Does that make sense?

selsta: I changed it back to the 2. option.

PR #778, opened 2017-12-25
Testnet: add monitoring

MoroccanMalinois:
---

---
if `KOVRI_DISABLE_MONITORING` is set to `false`, it will create 4 additional containers:
- 1 for a time series database ([influxdb](https://www.influxdata.com/))
- 1 for collecting statistics from kovri instances via kovri-util
- 1 for collecting container statistics ([cadvisor](https://github.com/google/cadvisor)) 
- 1 for visualization ([grafana](https://grafana.com/))

There is a variable at the top left to choose what it is included (standard regex).

So out of the box, it produces things like:
containers
instance

Dashboards can be edited via the UI and then backuped via the script contrib/testnet/monitoring/dashboards/backup.py

### DIFF
--- a/contrib/testnet/monitoring/dashboards/details-netdb.json
+++ b/contrib/testnet/monitoring/dashboards/details-netdb.json
@@ -1,0 +1,635 @@
+{
+    "dashboard": {
+        "annotations": {
+            "list": [
+                {
+                    "builtIn": 1,
+                    "datasource": "-- Grafana --",
+                    "enable": true,
+                    "hide": true,
+                    "iconColor": "rgba(0, 211, 255, 1)",
+                    "name": "Annotations & Alerts",
+                    "type": "dashboard"
+                }
+            ]
+        },
+        "editable": true,
+        "gnetId": null,
+        "graphTooltip": 0,
+        "hideControls": false,
+        "id": null,
+        "links": [],
+        "refresh": "10s",
+        "rows": [
+            {
+                "collapse": false,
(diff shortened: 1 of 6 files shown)


contrib/testnet/monitoring/dashboards/details-netdb.json
```diff
@@ -0,0 +1,730 @@
...
+                        },
+                        {
+                            "selected": false,
+                            "text": "kovri_testnet_028",
+                            "value": "kovri_testnet_028"
+                        },
+                        {
+                            "selected": false,
+                            "text": "kovri_testnet_029",
+                            "value": "kovri_testnet_029"
```
anonimal: >kovri_testnet_029

Is there any way we can generate these entries dynamically? For the cases where instance number 029 doesn't exist? What will happen if that instance doesn't exist? What happens if there is 030?

MoroccanMalinois: > Is there any way we can generate these entries dynamically?

Yes. Basically, they are generated by a query :

    SHOW TAG VALUES WITH KEY = \"instance\"

So it should adapt to any number of instances, but i did indeed forgot to activate the auto refresh for `Instance Overview` (fixed now).

For the other dashboards, i disabled the auto refresh to keep a set of predefined filters (otherwise it auto selects the first element in the result, bypassing the predefined default filter and these views are more suited to compare multiple instances).

Issue #785, opened 2018-01-04
Remove crypto's pimpl idiom


anonimal:
We don't need it, I never wanted it, it's annoying to deal with and it's unneeded overhead. Crypto++ is a fine library of which I've worked side-by-side with and I don't foresee it leaving anytime soon. We also don't change our crypto enough to warrant a pimpl design so, simply treating our implementation file(s) as implementation files should suffice - without the need for pimpl.

Removing pimpl won't prevent us from changing libraries. If we end up changing any interfaces in the process of resolving this issue, the changes will be for the better.

---
**By submitting this issue, I confirm the following:**

- I have read and understood the contributor guide in [kovri-docs](https://github.com/monero-project/kovri-docs).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.
---

anonimal:
NOTICE: THIS ISSUE HAS BEEN MOVED TO GitLab. Please continue the discussion there. See https://github.com/monero-project/kovri/issues/1013 for details.

PR #788, opened 2018-01-10
Kovri: new base32/64 impl / radix interface / unit-test

### LINKED ISSUE #786: Remove i2pd's base32/64 implementation, re-implement using Crypto++
WIP.

---

---

anonimal:
- Removes i2pd's base32/64 impl, introduces new impl using crypto++
- Includes related rewrites, refactors, new documentation, and dead code removal

Resolves #786. Works into #785. 

---

---


anonimal:
Live tests and all new test-cases pass (only tested on Linux, awaiting buildbot).

src/core/router/info.cc
```diff
@@ -318,10 +319,23 @@ void RouterInfo::ParseRouterInfo(const std::string& router_info)
...
+                      key = core::Base64::Decode(value.c_str(), value.size());
+                    }
+                  catch (...)
+                    {
+                      m_Exception.Dispatch("RouterInfo: invalid intro key trait");
+                      is_valid_address = false;
+                    }
+
+                  //TODO(anonimal): let's try to avoid a memcpy
+                  std::memcpy(address.key, key.data(), sizeof(address.key));
```
coneiric: not easily searchable for a secure alternative. from (cppreference)[http://en.cppreference.com/w/cpp/string/byte/memcpy]:
- If either dest or src is a null pointer, the behavior is undefined, even if count is zero.
- If the objects are not TriviallyCopyable, the behavior of memcpy is not specified and may be undefined

For the first case, checking that buffers are non-null seems like an easy fix.

Maybe for the `TriviallyCopyable` case, could use:
```
try {
bool address_key_trivially_copyable = is_trivially_copyable<address.key>::value;
bool key_trivially_copyable ...

if !address_key_trivially_copyable || !key_trivially_copyable
    throw
}
catch (...)
```

Know this doesn't fully address the "secure memcpy" problem, but would it help make the current situation more secure?

src/core/crypto/impl/cryptopp/radix.cc
```diff
@@ -26,82 +26,89 @@
...
+{
+namespace core
+{
+template <typename T>
+std::string Radix<T>::m_Base32Alphabet("abcdefghijklmnopqrstuvwxyz234567");
+
+std::string Base32::Encode(const std::uint8_t* in, const std::uint64_t len)
+{
+  // Prepare encoder
+  CryptoPP::AlgorithmParameters static const params(CryptoPP::MakeParameters(
```
coneiric: would you be open to adding a documentation line pointing to the CryptoPP Manual page for AlgorithmParameters (https://www.cryptopp.com/docs/ref/class_algorithm_parameters.html)?

Behavior for MakeParameters and AlgorithmParameters was hard to understand from reading the code, until I read the manual page, and it became very obvious. 

Mainly the confusing bit was where AlgorithmParameters overrides the `()` operator to append a key-value pair, and return an AlgorithmParameters ref (allowing one to append an arbitrary number of key-value pairs).

src/core/router/info.cc
coneiric: not easily searchable for a secure alternative. from [cppreference](http://en.cppreference.com/w/cpp/string/byte/memcpy):
- If either dest or src is a null pointer, the behavior is undefined, even if count is zero.
- If the objects are not TriviallyCopyable, the behavior of memcpy is not specified and may be undefined

For the first case, checking that buffers are non-null seems like an easy fix.

Maybe for the `TriviallyCopyable` case, could use:
```
try {
bool address_key_trivially_copyable = is_trivially_copyable<address.key>::value;
bool key_trivially_copyable ...

if !address_key_trivially_copyable || !key_trivially_copyable
    throw
}
catch (...)
```

Know this doesn't fully address the "secure memcpy" problem, but would it help make the current situation more secure?

src/core/crypto/impl/cryptopp/radix.cc
anonimal: >would you be open to adding a documentation line pointing to the CryptoPP Manual page for AlgorithmParameters

This isn't necessary because this type of c++ / crypto++ question is outside the scope of kovri. Adding links to code is generally a bad idea too (links become outdated) and doing so would create a precedent for endless links. Crypto++'s manual merely produces doxygen output so you can read the *same docs* within the code (see `deps/cryptopp`) or visit their website (as you did).

src/core/router/info.cc
anonimal: This TODO was a note for me to implement `address.key` as a container and/or implement an overloaded operator for `Tag` and/or rewrite/throw-out `Tag`. See also #366 and #784 (note: a public address key in this context should not need #784).

The resolution of any of those aforementioned points should also resolve your concerns.

anonimal:
[FreeBSD's unit-test fails in test-case `ParseACL`](https://build.getmonero.org/builders/kovri-all-freebsd64/builds/752/steps/test/logs/stdio) though all other platforms pass. Hold on merging.

anonimal:
Here's a backtrace (note: the issue is reproducible with clang++39):

```
#0  0x000000080349339a in thr_kill () from /lib/libc.so.7
#1  0x0000000803493386 in raise () from /lib/libc.so.7
#2  0x0000000803493309 in abort () from /lib/libc.so.7
#3  0x000000000047a78f in __clang_call_terminate () at ios:507
#4  0x000000000059651e in CryptoPP::member_ptr<CryptoPP::AlgorithmParametersBase>::~member_ptr (this=0x7fffffffca80) at smartptr.h:71
#5  0x0000000000595481 in CryptoPP::AlgorithmParameters::~AlgorithmParameters (this=0x7fffffffca78) at algparam.h:433
#6  0x0000000000594e5d in kovri::core::Base32::Decode (in=0x805032a99 "", len=0) at /usr/home/anonimal/kovri/src/core/crypto/impl/cryptopp/radix.cc:76
#7  0x00000000007a2183 in kovri::core::Tag<32>::FromBase32 (this=0x7fffffffd1c0, encoded=@0x805032a98) at identity.h:124
#8  0x000000000092bf90 in kovri::client::ParseACL (list=
      {<std::__1::__basic_string_common<true>> = {<No data fields>}, __r_ = {<std::__1::__libcpp_compressed_pair_imp<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::__rep, std::__1::allocator<char>, 2>> = {<std::__1::allocator<char>> = {<No data fields>}, __first_ = {{__l = {__cap_ = 177, __size_ = 164, __data_ = 0x805045300 "7ooubi6turxmjmtrpx2sq2mfpmjaywr3aqzs6skii4lwiwipcmoa,,,,65m4tm6zwzn3wkj2c56abig7vjrxnucszbsdpwrog3cocpxru2hq,,,,3pz774osmoosan7opdjyzld2knlyizwnggfounbzmd6m5tz5brla"}, __s = {{__size_ = 177 '', __lx = -79 ''}, __data_ = 0x7fffffffd531 ""}, __r = {__words = 0x7fffffffd530}}}}, <No data fields>}, static npos = <optimized out>}) at /usr/home/anonimal/kovri/src/client/util/parse.cc:56
#9  0x00000000004de97c in ClientParsing::ParseACL::test_method (this=0x7fffffffddb0) at /usr/home/anonimal/kovri/tests/unit_tests/client/util/parse.cc:77
#10 0x00000000004ddf17 in ClientParsing::ParseACL_invoker () at /usr/home/anonimal/kovri/tests/unit_tests/client/util/parse.cc:45
```

At least two immediate working solutions:

1. I don't like this because there's no apparent reason why we should ever throw:

```diff
diff --git a/deps/cryptopp b/deps/cryptopp
--- a/deps/cryptopp
+++ b/deps/cryptopp
@@ -1 +1 @@
-Subproject commit f1a80e6a58e6cf72c969dda6825b9781eb300927
+Subproject commit f1a80e6a58e6cf72c969dda6825b9781eb300927-dirty
diff --git a/src/core/crypto/impl/cryptopp/radix.cc b/src/core/crypto/impl/cryptopp/radix.cc
index 4a37cdd..eac9f97 100644
--- a/src/core/crypto/impl/cryptopp/radix.cc
+++ b/src/core/crypto/impl/cryptopp/radix.cc
@@ -69,7 +69,11 @@ std::vector<std::uint8_t> Base32::Decode(
 
   CryptoPP::AlgorithmParameters const params(CryptoPP::MakeParameters(
       CryptoPP::Name::DecodingLookupArray(),
+#if defined(__FreeBSD__)  // See #788
+      reinterpret_cast<const int*>(lookup), false));
+#else
       reinterpret_cast<const int*>(lookup)));
+#endif
 
   // Decode
   return Radix::Decode<CryptoPP::Base32Decoder>(params, in, len);
```

2. I don't like this because it's more work for us:

```diff
diff --git a/src/core/router/identity.h b/src/core/router/identity.h
index e16f8975..c3716080 100644
--- a/src/core/router/identity.h
+++ b/src/core/router/identity.h
@@ -118,6 +118,9 @@ class Tag {
 
   void FromBase32(const std::string& encoded)
   {
+    if (encoded.empty())
+      throw std::length_error("Tag: null encoded length for b32");
+
     std::vector<std::uint8_t> const decoded =
         core::Base32::Decode(encoded.c_str(), encoded.length());

```

and then point 2 should be applied elsewhere. I'm personally vying for point 1.

After reviewing cryptopp's smartptr and algparam, I don't see what patches could be applied (at least ones that wouldn't be unnecessary or intrusive). Maybe freebsd clang needs to work out another bug? 

Pinging @noloader.

anonimal:
All tests now pass with the pre-processor patch applied. I'd like to wait for a resolution to https://github.com/weidai11/cryptopp/issues/562 until merging.

noloader:
@anonimal, @coneiric,

I'm crafting some self test for Crypto++ to test swapping an alphabet. Regarding this change, I think you may have been on the right track:

```
   CryptoPP::AlgorithmParameters const params(CryptoPP::MakeParameters(
       CryptoPP::Name::DecodingLookupArray(),
+#if defined(__FreeBSD__)  // See #788
+      reinterpret_cast<const int*>(lookup), false));
+#else
       reinterpret_cast<const int*>(lookup)));
+#endif
```

In fact, the library makes its parameters using `false`, too. Its due to a hidden `CombinedNameValuePairs`, which allows you two combine two sets of `NameValuePairs`:

```
void Base64Decoder::IsolatedInitialize(const NameValuePairs &parameters)
{
    BaseN_Decoder::IsolatedInitialize(CombinedNameValuePairs(
        parameters,
        MakeParameters(Name::DecodingLookupArray(), GetDecodingLookupArray(), false)(Name::Log2Base(), 6, true)));
}
```

So the quickest way to side step the issue and get on with your real work is probably:

```
    CryptoPP::AlgorithmParameters const params(CryptoPP::MakeParameters(
        CryptoPP::Name::DecodingLookupArray(),
            reinterpret_cast<const int*>(lookup), false));
```

anonimal:
>MakeParameters(Name::DecodingLookupArray(), GetDecodingLookupArray(), false)(Name::Log2Base(), 6, true)));

@noloader Ah, I see now. Thank you for clarifying!

Though I see that `false` is harmless, I'd like to keep the pre-processor cruft until https://github.com/weidai11/cryptopp/issues/562 is resolved (since I'll need to leave a comment anyway regarding the explicit use of `false`).

noloader:
@anonimal,

Forgive my ignorance... In this:

```
     std::vector<std::uint8_t> const decoded =
         core::Base32::Decode(encoded.c_str(), encoded.length());
```

What is `encoded`? Is it a `std::string` or something else?

If it is a `std::string` you should be OK. If it is another type, then be sure `c_str()` returns a pointer to something rather than `NULL`, `nullptr_t` or something else.

anonimal:
>What is encoded? Is it a std::string or something else?

Yes, `std::string`.

anonimal:
On an unrelated note, I've added and rebased a few small changes:

- added an alphabet test-case
- moved the alphabet getter into translation unit (will silence clang warnings)
- clarified the names of fixture data

On a related note, this PR should be safe to merge while @noloader and I continue debugging https://github.com/weidai11/cryptopp/issues/562.

PR #799, opened 2018-01-26
AddressBook: decode base32 substring from valid host name

Neozaru:
When decoding a base32 address, the full address was passed as argument to the decoder, instead of only the base32 subpart to decode.
Ex: Was trying to decode `uaknjcssainezmqcgfavwqsewu7nnuaakk2m7ywprcirra7m7tqa.b32.i2p` instead of just `uaknjcssainezmqcgfavwqsewu7nnuaakk2m7ywprcirra7m7tqa`

---

- I have [partially] read and understood the contributor guide in [kovri-docs](https://github.com/monero-project/kovri-docs).
---

### DIFF
--- a/src/client/address_book/impl.cc
+++ b/src/client/address_book/impl.cc
@@ -390,7 +390,7 @@ bool AddressBook::CheckAddressIdentHashFound(
     {
       try
         {
-          ident.FromBase32(address);
+          ident.FromBase32(address.substr(0, pos));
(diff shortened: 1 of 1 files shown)


Neozaru:
Yes, sorry. Changed commit message.

anonimal:
Note: this bug was introduced / overlooked in #788.

PR #802, opened 2018-01-29
Radix: remove FreeBSD-specific build block

anonimal:
Crypto++ 6.0 release has resolved this decoding issue (#794).

---
**By submitting this pull-request, I confirm the following:**

- I have read and understood the contributor guide in [kovri-docs](https://github.com/monero-project/kovri-docs).
- I have checked that another pull-request for this purpose does not exist.
- I have considered and confirmed that this submission will be valuable to others.
- I accept that this submission may not be used and that this pull-request may be closed by the will of the maintainer.
- I give this submission freely under the BSD 3-clause license.
---

PR #804, opened 2018-01-31
Unroll DecodingLookupArray for Base32 and Base64 decoders

noloader:
This is a time/space tradeoff, but it also provides one-time initialization while avoiding a race.
It should also clear https://github.com/weidai11/cryptopp/issues/562, https://github.com/monero-project/kovri/pull/788 and https://github.com/monero-project/kovri/pull/802 once and for all (famous last words).

I was not sure of the project's style regarding internal linkage and hiding names so things were added to an unnamed namespace. Feel free to modify it mercilessly.

---

---

### DIFF
--- a/src/core/crypto/impl/cryptopp/radix.cc
+++ b/src/core/crypto/impl/cryptopp/radix.cc
@@ -37,6 +37,49 @@
 #include <string>
 #include <vector>
 
+// Anonymous namespace
+namespace {
+    // Decoding table for m_Base32Alphabet
+    const int s_Base32Table[256] = {
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+        -1, -1, 26, 27, 28, 29, 30, 31, -1, -1, -1, -1, -1, -1, -1, -1,
+        -1,  0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14,
+        15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, -1, -1, -1, -1, -1,
+        -1,  0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14,
+        15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, -1, -1, -1, -1, -1,
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1
+    };
+
+    // Decoding table for m_Base64Alphabet
+    const int s_Base64Table[256] = {
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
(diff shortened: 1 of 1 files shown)


noloader:
Sorry again about the three commits. I tried to rebase, but Git is the only tool on my machine that cannot find the editor that just about everyone uses on WIndows. The worst part is, `notepad++.exe` is not hidden. It is in the same place Windows programs have resided for the last 2o years or so. That tool is so braindead.

```
$ git rebase -i HEAD~3
/mingw64/libexec/git-core/git-rebase--interactive: line 267: notepad++.exe: command not found
Could not execute editor
```

noloader:
> I do like the efficiency provided by the tradeoff, but must we unroll the decoding tables?

Nope, feel free to recompute it each time it is needed.

Computing the table on the fly should save you about 1KB per table in storage.

-----

> Outside of the problem that we're currently dealing with, do you know of any other use-case for a kovri-unrolled table?

None that I am aware, but I am not too familiar with Kovri.

-----

> Here, you suggest to simply pass false as a quick fix (confirmed fix) because that's what the library does internally:
> 
> ```
> CryptoPP::AlgorithmParameters const params(CryptoPP::MakeParameters(
>     CryptoPP::Name::DecodingLookupArray(),
>     reinterpret_cast<const int*>(lookup),
>     false));
> ```
> 
> Would you advise against us doing this?

No. I recommend you do it.

We have not been able to get things under a debugger to see what is causing the problem. Until we know the cause it is easier to sidestep the problem.

Also, if Wei did not want you to use the overload, then it would have been protected or private to the class. I suspect this is one of the use cases for public `throwIfUnused`.

src/core/crypto/impl/cryptopp/radix.cc
```diff
@@ -109,16 +140,8 @@ std::vector<std::uint8_t> Base64::Decode(
...
-      lookup,
-      reinterpret_cast<const CryptoPP::byte*>(m_Base64Alphabet.c_str()),
-      GetAlphabet().size(),
-      false);
-
-  CryptoPP::AlgorithmParameters const params(CryptoPP::MakeParameters(
-      CryptoPP::Name::DecodingLookupArray(),
-      reinterpret_cast<const int*>(lookup)));
+  CryptoPP::AlgorithmParameters static const params(CryptoPP::MakeParameters(
+      CryptoPP::Name::DecodingLookupArray(), static_cast<const int*>(s_Base64Table), false));
```
anonimal: >reinterpret_cast<const CryptoPP::byte*>(m_Base64Alphabet.c_str()),
>GetAlphabet().size(),

@noloader Live tests pass and unit-tests pass but what I don't understand is how they are able to pass without using our custom alphabet. I'm concerned about silent failures but our [testcase](https://github.com/monero-project/kovri/blob/master/tests/unit_tests/core/crypto/radix.cc#L133) shows successful decoding. How is this possible?!

noloader: > what I don't understand is how they are able to pass without using our custom alphabet.

It did use your custom alphabet:

```
p$ cat test.cxx
#include "cryptlib.h"
#include "secblock.h"
#include "osrng.h"
#include "base32.h"
#include "hex.h"

#include <iostream>
#include <string>

using namespace CryptoPP;

int main(int argc, char* argv[])
{
    AutoSeededRandomPool prng;

    // char alphabet[] = "abcdefghijklmnopqrstuvwxyz234567";  // and 32, true
    char alphabet[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-~";

    int lookup[256];
    Base32Decoder::InitializeDecodingLookupArray(
        lookup,
        (const byte*)alphabet, 64,
        false);

    unsigned int i=0;
    while(i<COUNTOF(lookup))
    {
        std::cout << lookup[i] << ", ";

        if (++i%16 == 0)
            std::cout << "\n";
    }
    std::cout << std::endl;

    // Test for Monero usage
    AlgorithmParameters static const params(MakeParameters(
        Name::DecodingLookupArray(), static_cast<const int*>(lookup), false));
    return 0;
```

anonimal: Sorry! I brain farted when comparing the RFC table. Thanks for clarifying.

noloader:
@anonimal,

> Here, you suggest to simply pass false as a quick fix (confirmed fix) because that's what the library does internally ...

One other thing I have not been able to rule out... The `const` on:

```
CryptoPP::AlgorithmParameters const params(CryptoPP::MakeParameters(
    CryptoPP::Name::DecodingLookupArray(), reinterpret_cast<const int*>(lookup));
```

Should prohibit consumers/called functions from changing the internal member `m_unsed` from `false` to `true`. However, ` m_used` is declared as `mutable bool m_used;` in [`AlgorithmParametersBase`](https://github.com/weidai11/cryptopp/blob/master/algparam.h#L300):

```
class CRYPTOPP_DLL AlgorithmParametersBase
{
   ...
protected:
    ...
   const char *m_name;
   bool m_throwIfNotUsed;
   mutable bool m_used;
   member_ptr<AlgorithmParametersBase> m_next;
};
```

I've got two theories on it. First, I think the optimizer may be getting a bit overzealous. In this case the optimizer deduces `m_used` is never written so it schedules the throw. Second, the `AlgorithmParameters const params` is being allocated in an unexpected way, like the BSS section with initialized and read-only memory.In this case `static` and `const` play a role in the problem.

Like I said a while back, I've only seen that one other time on OS X. But it is hard to duplicate the problem and untangle what is going on. I really need to see the problem locally so I can start poking and proding at things.

anonimal:
Oops, github is showing me https://github.com/monero-project/kovri/pull/804#issuecomment-362046184 *after* I approved the PR. I'll respond in https://github.com/weidai11/cryptopp/issues/562.

PR #808, opened 2018-02-08
Contrib: archiso profile for Kovri dev + testnet

coneiric:
Archiso profile and instructions for building a LiveUSB.
LiveUSB comes with everything a user needs to use, develop, and test Kovri.

---

---


contrib/archiso/airootfs/root/customize_airootfs.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# Clone latest Kovri repo if it doesn't exist
+if [[ ! -d /usr/src/kovri ]]; then
+  git clone --recursive https://github.com/monero-project/kovri.git /usr/src/kovri
+fi
+
+# Build and install Kovri
+cd /usr/src/kovri && KOVRI_DATA_PATH=/home/kovri/.kovri make -j$(nproc) release && make install
+chown -R kovri:kovri /home/kovri
+ln -sf /usr/src/kovri/build/{kovri,kovri-util} /usr/bin
```
anonimal: `make install` installs into `~/bin/`.

coneiric: Thanks for catching that. Leftover from previous iterations.

coneiric:
> Can we keep the archiso stuff elsewhere and keep the kovri things here?

I was thinking of starting a separate repo for this, and with your feedback this sounds like the right thing to do.

> I think an archiso is neat idea for when we have actual releases but having to rebuild for every system update until then?!

100% agreed, failed to realize that this would cause rebuilds for changes unrelated to Kovri.

I'll move the archiso files into their own repo, and maintain them as a stand-alone project.

anonimal:
>I was thinking of starting a separate repo for this, and with your feedback this sounds like the right thing to do.

I think that would be awesome. We can add a link somewhere in the docs or contrib.

Issue #812, opened 2018-02-17
HTTP Proxy Jump Service Features


coneiric:
---
**By submitting this issue, I confirm the following:**

- I have read and understood the contributor guide in [kovri-docs](https://github.com/monero-project/kovri-docs).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.
---

After looking through the TODOS [1](https://github.com/monero-project/kovri/blob/master/src/client/proxy/http.cc#L396) [2](https://github.com/monero-project/kovri/blob/master/src/client/proxy/http.cc#L421) for `HTTPMessage::HandleJumpService`, I came up with some ideas for a redesign.

__edit__: add prerequisite issues, reformat as task list

Below are what I understand to be the required features of the jump service handler.
Some are already implemented, others are new proposals:
- [ ] load entries from multiple host files (requires resolution of #305 / #337)
- [ ] add new host entry (requires resolution of #305 )
- check address not already in: (requires resolution of #305 / #337)
  - [ ] hosts.txt
  - [ ] private_hosts.txt 
  - [ ] subscription-specific hosts files
- [ ] prompt user to save new host entry
- [ ] display information of host entry being added
  - [ ] short hostname (host-name.i2p)
  - [ ] base32 address (long-b32-string.b32.i2p)
  - [ ] full base64-encoded destination information (base64 blob)
- [ ] provide information on save location options (requires resolution of #305 / #337)
  - private_hosts.txt, service_hosts.txt, etc.
- [ ] provide option to cancel saving new entry
- [ ] replace ByteStream instances with modular & secure interface, in pursuit of #811
- [x] convert `HTTPMessage::HandleJumpService()` to bool return for testing
- [ ] unit tests (pass existing tests, and write new tests)
- what can be implemented before #385 is resolved? (may be solved by resolution of #305 / #337)
  - treat storage implementation as blackbox
  - [ ] handle reading & writing with modular interfaces
  - [ ] use / write new interfaces independent of storage implementation (maybe?)

Related Kovri source files:
- HTTP Proxy [impl](https://github.com/monero-project/kovri/blob/master/src/client/proxy/http.cc) and [header](https://github.com/monero-project/kovri/blob/master/src/client/proxy/http.h)
[...]

Issue #721, opened 2017-09-20
Remove cpp-netlib as dependency, replace with Boost.Beast


anonimal:
Boost.Beast is in beta but will be released in Boost 1.66. Something like Boost.Beast is what we originally wanted when implementing cpp-netlib but Boost.Beast was not really around nor usable enough at the time (cpp-netlib went in the direction of Boost but, for various reasons, bailed and now appear to be a dead project).

Boost.Beast plans on implementing a URI parser. When that happens, we should drop cpp-netlib entirely. In the meantime, we can also consider a Boost.Beast refactor and simply keep the URI lib from cpp-netlib.

Referencing:

- https://github.com/cpp-netlib/cpp-netlib/issues/670
- https://github.com/boostorg/beast/issues/787
- https://github.com/monero-project/kovri/issues/155#issuecomment-330700570

---
**By submitting this issue, I confirm the following:**

- I have read and understood the contributor guide in [kovri-docs](https://github.com/monero-project/kovri-docs).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.
---

anonimal:
NOTICE: THIS ISSUE HAS BEEN MOVED TO GitLab. Please continue the discussion there. See https://github.com/monero-project/kovri/issues/1013 for details.

PR #813, opened 2018-02-20
HTTPMessage: refactor jump service parser + add test-cases

coneiric:
---

---

Part of #812, HTTPMessage::HandleJumpService now searches the HTTP URI for the `i2paddresshelper` parameter.

Unit test to ensure correct handling of valid jump service parameter.

Still requires unit tests for invalid parameters, and other failure cases.

### DIFF
--- a/src/client/proxy/http.cc
+++ b/src/client/proxy/http.cc
@@ -313,7 +313,39 @@ bool HTTPMessage::CreateHTTPRequest() {
     // m_ErrorResponse is set in ExtractIncomingRequest
     return false;
   }
-  HandleJumpService();
+
+  if (!IsJumpServiceRequest())
+    {
+      LOG(debug) << "HTTPProxyHandler: not a jump service request";
+    }
+  else  // Handle the jump service request
+    {
+      if (!HandleJumpService())
+        {
+          LOG(error) << "HTTPMessage: invalid jump service request";
+          m_ErrorResponse =
+              HTTPResponse(HTTPResponseCodes::status_t::bad_request);
+          return false;
+        }
+      else  // Requested address found, save to address book
+        {
+          // TODO(oneiric): this is very dangerous and broken
+          // we should prompt the user with an HTTP redirect to a save form
+          // save form should contain:
+          // - host info: short address, base32 address, base64 destination
+          // - save location options
+          // - continue without saving option
+          if (!SaveJumpServiceAddress())
+            {
+              LOG(error)
+                  << "HTTPProxyHandler: failed to save address to address book";
+              m_ErrorResponse = HTTPResponse(
+                  HTTPResponseCodes::status_t::internal_server_error);
+              return false;
+            }
+        }
+    }
+
   // Set method, path, and version
   m_Request = m_Method;
   m_Request.push_back(' ');
--- a/src/client/proxy/http.cc
+++ b/src/client/proxy/http.cc
@@ -390,41 +390,94 @@ bool HTTPMessage::ExtractIncomingRequest() {
   return true;
 }
 
-void HTTPMessage::HandleJumpService() {
-  // TODO(GUZZI) should have boolean return value; error
-  // response?; research this
+bool HTTPMessage::HandleJumpService()
+{
   // TODO(anonimal): add support for remaining services /
   // rewrite this function
-  std::size_t pos1 = m_Path.rfind(m_JumpService.at(0));
-  std::size_t pos2 = m_Path.rfind(m_JumpService.at(1));
-  std::size_t pos;
-  if (pos1 == std::string::npos) {
-    if (pos2 == std::string::npos)
-      return;  // Not a jump service
-    else
-      pos = pos2;
-  } else {
-    if (pos2 == std::string::npos)
-      pos = pos1;
-    else if (pos1 > pos2)
-      pos = pos1;
-    else
-      pos = pos2;
-  }
-  auto base64 = m_Path.substr(pos + m_JumpService.at(0).size());
-  // We must decode
-  base64 = boost::network::uri::decoded(base64);
-  // Insert into address book
-  LOG(debug)
-    << "HTTPProxyHandler: jump service for " << m_Address
-    << " found at " << base64 << ", inserting to address book";
-  // TODO(unassigned): this is very dangerous and broken.
-  // We should ask the user for confirmation before proceeding.
-  // Previous reference: http://pastethis.i2p/raw/pn5fL4YNJL7OSWj3Sc6N/
-  // We *could* redirect the user again to avoid dirtiness in the browser
-  kovri::client::context.GetAddressBook().InsertAddressIntoStorage(
-      m_Address, base64);
-  m_Path.erase(pos);
+
+  // Perform sanity check again to:
+  // - ensure valid jump service request
+  std::size_t const pos = IsJumpServiceRequest();
+  if (!pos)
+    {
+      LOG(error) << "HTTPProxyHandler: not a valid jump service request";
+      // Jump service parameter not found in URL, so just return
+      return false;
+    }
+
+  if (!ExtractBase64Destination(pos))
+    {
+      LOG(error) << "HTTPProxyHandler: unable to process base64 destination for "
+                 << m_Address;
+      m_URL.erase(pos);
+      return false;
+    }
+
+  LOG(debug) << "HTTPProxyHandler: jump service for " << m_Address
+             << " found at " << m_Base64Destination;
+  m_URL.erase(pos);
+  return true;
+}
+
+std::size_t HTTPMessage::IsJumpServiceRequest() const
+{
+  std::size_t pos = 0;
+  std::size_t const pos1 = m_URL.rfind(m_JumpService.at(0));
+  std::size_t const pos2 = m_URL.rfind(m_JumpService.at(1));
+  if (pos1 == std::string::npos)
+    {
+      if (pos2 == std::string::npos)
+        return 0;  // Not a jump service
+      else
+        pos = pos2;
+    }
+  else
+    {
+      if (pos2 == std::string::npos)
+        pos = pos1;
+      else if (pos1 > pos2)
+        pos = pos1;
+      else
+        pos = pos2;
+    }
+
+  // final sanity check
+  if (pos && pos != std::string::npos)
+    return pos;
+
+  return 0;
+}
+
+bool HTTPMessage::ExtractBase64Destination(std::size_t const pos)
+{
+  std::size_t const base64_size = pos + m_JumpService.at(0).size();
+  if (pos && base64_size < m_URL.size())
+    {
+      std::string const base64 = m_URL.substr(base64_size);
+      m_Base64Destination = boost::network::uri::decoded(base64);
+      return true;
+    }
+
+  return false;
+}
+
+bool HTTPMessage::SaveJumpServiceAddress()
+{
+  try
+    {
+      LOG(debug) << "HTTPProxyHandler: inserting " << m_Address
+                 << " into address book";
+      kovri::client::context.GetAddressBook().InsertAddressIntoStorage(
+          m_Address, m_Base64Destination);
+    }
+  catch (...)
+    {
+      core::Exception ex;
+      ex.Dispatch("HTTPProxyHandler: unable to insert address into storage");
+      return false;
+    }
+
+  return true;
 }
 
 /* All hope is lost beyond this point */
--- a/src/client/proxy/http.h
+++ b/src/client/proxy/http.h
@@ -1,5 +1,5 @@
 /**                                                                                           //
- * Copyright (c) 2013-2017, The Kovri I2P Router Project                                      //
+ * Copyright (c) 2013-2018, The Kovri I2P Router Project                                      //
  *                                                                                            //
  * All rights reserved.                                                                       //
  *                                                                                            //
@@ -195,7 +195,7 @@ class HTTPMessage : public std::enable_shared_from_this<HTTPMessage>{
     m_URL, m_Method, m_Version,  m_Path;
   std::vector<std::string> m_Headers;
   std::string m_Host, m_UserAgent;
-  std::string m_Address;
+  std::string m_Address, m_Base64Destination;
 
   boost::asio::streambuf m_Buffer;
   boost::asio::streambuf m_BodyBuffer;
@@ -226,9 +226,9 @@ class HTTPMessage : public std::enable_shared_from_this<HTTPMessage>{
   /// return bool
   bool HandleData(const std::string  &buf);
 
-  /// @brief Parses path for base64 address, inserts into address book
-  /// need to change to boolean return for testing
-  void HandleJumpService();
+  /// @brief Parses URI for base64 destination
+  /// @return true on success
+  bool HandleJumpService();
 
   /// @brief Performs regex, sets address/port/path, validates version
   ///   on request sent from user
@@ -242,6 +242,19 @@ class HTTPMessage : public std::enable_shared_from_this<HTTPMessage>{
 
   const unsigned int HEADERBODY_LEN = 2;
   const unsigned int REQUESTLINE_HEADERS_MIN = 1;
+ private:
+  /// @brief Checks if request is a valid jump service request
+  /// @return Index of jump service helper sub-string, 0 indicates failure
+  std::size_t IsJumpServiceRequest() const;
+
+  /// @brief Extracts & url-decodes base64 destination from URL
+  /// @param pos Index of jump service helper sub-string in URL
+  /// @return True on success
+  bool ExtractBase64Destination(std::size_t const pos);
+
+  /// @brief Saves found address in address book
+  /// @return True on success
+  bool SaveJumpServiceAddress();
 };
 /// @class HTTPProxyServer
 /// setup asio service
--- a/tests/unit_tests/client/proxy/http.cc
+++ b/tests/unit_tests/client/proxy/http.cc
@@ -1,5 +1,5 @@
 /**                                                                                           //
- * Copyright (c) 2013-2016, The Kovri I2P Router Project                                      //
+ * Copyright (c) 2013-2018, The Kovri I2P Router Project                                      //
  *                                                                                            //
  * All rights reserved.                                                                       //
  *                                                                                            //
@@ -33,6 +33,53 @@
 #include <boost/test/unit_test.hpp>
 #include "client/proxy/http.h"
 
+struct HTTPProxyRequestFixture
+{
+  enum struct FunctionName 
+  {
+    CreateHTTPRequest,
+    ExtractIncomingRequest,
+    HandleData,
+    HandleJumpService
+  };
+
+  std::string CreateProxyHeader(const std::string uri) const
+  {
+    std::string proxy_header("GET ");
+    proxy_header.append(uri);
+    proxy_header.append(" HTTP/1.1\r\n\r\n");
+    return proxy_header;
+  }
+
+  bool HandleProxyFunction(
+      const std::string uri,
+      FunctionName function_name) const
+  {
+    std::string const header = CreateProxyHeader(uri);
+    kovri::client::HTTPMessage proxy_request;
+    if (!proxy_request.HandleData(header))
+      return false;
+    switch (function_name)
+      {
+        case FunctionName::CreateHTTPRequest:
+          if (!proxy_request.CreateHTTPRequest())
+            return false;
+          break;
+        case FunctionName::ExtractIncomingRequest:
+          if (!proxy_request.ExtractIncomingRequest())
+            return false;
+          break;
+        case FunctionName::HandleJumpService:
+          if (!proxy_request.HandleJumpService())
+            return false;
+          break;
+        default:
+          throw std::invalid_argument("unknown proxy function");
+      }
+    return true;
+  }
+};
+
 BOOST_AUTO_TEST_SUITE(HTTPPProtocolTests)
 
 BOOST_AUTO_TEST_CASE(Short) {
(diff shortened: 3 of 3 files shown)


tests/unit_tests/client/proxy/http.cc
```diff
@@ -57,4 +57,16 @@ BOOST_AUTO_TEST_CASE(ok) {
   BOOST_CHECK(tmp.HandleData(tmpData));
 }
 
+BOOST_AUTO_TEST_CASE(jumpI2PAddressHelper) {
+  kovri::client::HTTPMessage tmp;
+  // Jump service GET parameter
+  std::string tmpData = "GET stats.i2p?i2paddresshelper=";
+  // Base64-encoded I2P address
+  tmpData+="WED~HHcaMK5GdfwrIOnDLj27Waso9~UUM4PBy~Dqn8-chmJwb1yg3l7WTsMRKkAkMs7vAk3pWOLvxZPRxGP3jP1ELNUg3mL2Dvj8euij8efF38fxzUcgNQ-DpFR40D44nn2si9COqIDjm1njixVBOy7ab4QWFSdE5DgEGOQxjkRiAU5K5vNnq6iFT~idukr1PASA3P5KWDu4s9g~S7k9FHev~HPhUArbuS6U75YlKjQXxhZhxDzdotZ~xUS2ZJcZlsH~73OO6zL3FgCs4JlDkklgbwarK57F9Auys6mRC1KkYgmavE~w6IA8kDp4IbGnhb2Rtcfb0SZoVtifVNkflvXX8bjkaNUAZtjmRKFt9I47pHSrdNfpgCV8a5kFl1QC5Li8OD3YmZg3kgX8Q0uc4IBtTzFvyT73hlIa~guoYYsstVJ58el01RPVYqJeEm1hlxloGSgm6WheFfY664IN7J92dhZvmL4DdZdIgJwXwjKM5QCcwpp2b~yYyY~XUjSCBQAEAAcAAA==";
+  tmpData+=" HTTP/1.1\r\n\r\n";
```
anonimal: Please, read the style guide. Our upstream style guide is clearly documented. The following changes apply to the entire test-case and in any new work:

1. Don't use camelCase
2. Use appropriate spacing
3. Use appropriate names
4. Run `clang-format`

tests/unit_tests/client/proxy/http.cc
```diff
@@ -57,4 +57,16 @@ BOOST_AUTO_TEST_CASE(ok) {
   BOOST_CHECK(tmp.HandleData(tmpData));
 }
 
+BOOST_AUTO_TEST_CASE(jumpI2PAddressHelper) {
+  kovri::client::HTTPMessage tmp;
+  // Jump service GET parameter
+  std::string tmpData = "GET stats.i2p?i2paddresshelper=";
+  // Base64-encoded I2P address
+  tmpData+="WED~HHcaMK5GdfwrIOnDLj27Waso9~UUM4PBy~Dqn8-chmJwb1yg3l7WTsMRKkAkMs7vAk3pWOLvxZPRxGP3jP1ELNUg3mL2Dvj8euij8efF38fxzUcgNQ-DpFR40D44nn2si9COqIDjm1njixVBOy7ab4QWFSdE5DgEGOQxjkRiAU5K5vNnq6iFT~idukr1PASA3P5KWDu4s9g~S7k9FHev~HPhUArbuS6U75YlKjQXxhZhxDzdotZ~xUS2ZJcZlsH~73OO6zL3FgCs4JlDkklgbwarK57F9Auys6mRC1KkYgmavE~w6IA8kDp4IbGnhb2Rtcfb0SZoVtifVNkflvXX8bjkaNUAZtjmRKFt9I47pHSrdNfpgCV8a5kFl1QC5Li8OD3YmZg3kgX8Q0uc4IBtTzFvyT73hlIa~guoYYsstVJ58el01RPVYqJeEm1hlxloGSgm6WheFfY664IN7J92dhZvmL4DdZdIgJwXwjKM5QCcwpp2b~yYyY~XUjSCBQAEAAcAAA==";
```
anonimal: If you want a newline, you need to add a newline.

coneiric: I didn't want a newline. 

Was separating for clarity, since those three assignments to `tmpData` should all be one HTTP header line.

tests/unit_tests/client/proxy/http.cc
coneiric: Apologies for the misstep, failed to reread the style guide and went off surrounding style.

I'll make the changes, and keep these formatting changes in their own commit.

coneiric:
> The test-case fails on all systems except OSX 10.10

After looking at the build bot log, I remember coming across a similar issue with NetDb tests.
From the [log](https://build.getmonero.org/builders/kovri-all-openbsd-amd64/builds/430/steps/test/logs/stdio):
```
inserting to address book
unknown location(0): fatal error: in "HTTPPProtocolTests/jumpI2PAddressHelper": std::runtime_error: Filesystem: failed to create /home/buildbot/.kovri/client/address_book/addresses: No such file or directory
```
Because I "fixed" this issue by manually creating the necessary folder during unrelated work, the error didn't come up when I ran the tests related to this PR.

I'll fix the code in [HandleJumpService](https://github.com/monero-project/kovri/blob/master/src/client/proxy/http.cc#L425) to call [AddressStorage::AddAddress](https://github.com/monero-project/kovri/blob/master/src/client/address_book/storage.cc#L68), and properly handle errors related to file creation/access.

> If you want a bool return, it should have some use e.g., don't create the request if applicable

Absolutely, I'll keep other use-cases in mind during refactoring, and add a sanity check where you suggested.

> Please rebase into a single commit and use the commit body to log details

Would you like me to keep the formatting-specific changes in their own commit, or literally everything in one commit?

> Please test work before PR'ing.

Apologies for the spam, still trying to not suck at GH. I did test the code multiple times, and ensure passing before pushing + PR'ing.

Unfortunately, due to a manual "fix" discussed above, I did not catch the error causing the build failures.

anonimal:
>unknown location(0): fatal error: in "HTTPPProtocolTests/jumpI2PAddressHelper": std::runtime_error: Filesystem: failed to create /home/buildbot/.kovri/client/address_book/addresses: No such file or directory

I suggest reading-up on boost.test. Boost unit-tests are not designed for use with filesystem I/O because you are meant to be testing the data and not where the data is housed (if not in memory). This also applies to socket programming. If you need to test data, you'll have to ensure you know exactly what the callee's are doing - otherwise the test is ineffective.

>Would you like me to keep the formatting-specific changes in their own commit, or literally everything in one commit?

Since this is new code, there's no formatting changes to be made (so long as it's done right the first time). This should be detailed in the contributing guide. If anything isn't clear, we should update the guide.

tests/unit_tests/client/proxy/http.cc
anonimal: On a side-note, you'll want to be mindful of websafe encoding. Example browser request:

```
GET http://kovri.i2p/?i2paddresshelper=0UVPqAA4xUSfPYPBca24h8fdokhwcJZ4zWvELv-5OsBYTHKtnLzvK7byXtXT~fOV2pExi8vrkgarGTNDfJbB2KCsdVS3V7qwtTvoCGYyklcDBlJsWMj7H763hEz5rt9SzLkcpwhO3t0Zwe6jXL1UB-QW8Kx[2/5549]
Pc6OiJ1QpnE6Bo5OUm6jPurQGXdWCAPio5Z-YnRL46n0IHWOQPYYSStJMYPlPS-S75rMIKbZbEMDraRvSzYAphUaHfvtWr2rCSPkKh3EbrOiBYiAP2oWvAQCsjouPgVF2qwQRnBbiAezHedM2gXzkgIyCV2kGOOcHhiihd~7fWwJOloH-gO78QkmCuY-3kp3633v3MBw7pmABr-XNK
WnATZOuf2syWVBZbTnOXsWf41tu6a33HOuNsMxAOUrwbu7QRmT4X8X-EITwNlKN6r1t3uoQ~yZm4RKsJUsBGfVtKl8PBMak3flQAg95oV0OBDGuizIQ9vREOWvPGlQCAXZzEg~cUNbfBQAEAAcAAA%3d%3d HTTP/1.1
Host: kovri.i2p
User-Agent: Mozilla/5.0 (Windows NT 6.1; rv:52.0) Gecko/20100101 Firefox/52.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate
DNT: 1
Connection: keep-alive
Upgrade-Insecure-Requests: 1
```

Notice the encoded `==` sent from the browser. Though this doesn't necessarily apply to the test-case at the moment, just an FYI.

coneiric: Great catch, I wasn't even thinking about URL encoding.

It doesn't look like the HTTP proxy is doing any kind of encoding/decoding at the moment.

Should I try to implement URI processing for the HTTP proxy in another PR, using Boost.Beast or cpp-netlib? If so, I can add it to #812 for tracking.

coneiric:
> I suggest reading-up on boost.test. Boost unit-tests are not designed for use with filesystem I/O because you are meant to be testing the data and not where the data is housed (if not in memory).

Doing that now. Definitely new to this framework.

Would you suggest handling the filesystem I/O in the Boost fixture, or writing the test under a different paradigm/framework (e.g. functional test, google test)?

> If you need to test data, you'll have to ensure you know exactly what the callee's are doing - otherwise the test is ineffective.

What's the best way to unit test HandleJumpService? Currently the "save-to-disk" feature is tangled up into the processing functionality. Should I refactor the code, or write the test to mock-out the filesystem I/O?

anonimal:
>Would you suggest handling the filesystem I/O in the Boost fixture, or writing the test under a different paradigm/framework (e.g. functional test, google test)?

No, the library handles I/O and library testing is incumbent upon the library - not the implementer.

>What's the best way to unit test HandleJumpService? Currently the "save-to-disk" feature is tangled up into the processing functionality. Should I refactor the code, or write the test to mock-out the filesystem I/O?

Refactor.

src/client/proxy/http.cc
```diff
@@ -313,7 +313,11 @@ bool HTTPMessage::CreateHTTPRequest() {
     // m_ErrorResponse is set in ExtractIncomingRequest
     return false;
   }
-  HandleJumpService();
+
+  // Valid HTTP request, but not a valid jump service request
+  if (!HandleJumpService())
+    LOG(debug) << "HTTPMessage: not a valid jump service request";
```
anonimal: If the jump request is invalid, simply proceed to the address?! But we don't have the address in our addressbook (assuming at this point we've confirmed that it does not exist in the addressbook)!  That is another issue to look out for: confirm that we don't already have the address in the addressbook before inserting. See https://github.com/monero-project/kovri/issues/814.

src/client/proxy/http.cc
```diff
@@ -422,9 +426,19 @@ void HTTPMessage::HandleJumpService() {
   // We should ask the user for confirmation before proceeding.
   // Previous reference: http://pastethis.i2p/raw/pn5fL4YNJL7OSWj3Sc6N/
   // We *could* redirect the user again to avoid dirtiness in the browser
-  kovri::client::context.GetAddressBook().InsertAddressIntoStorage(
-      m_Address, base64);
-  m_Path.erase(pos);
+  try
+  {
+    kovri::client::context.GetAddressBook().InsertAddressIntoStorage(
+        m_Address, base64);
```
anonimal: Also take a look inside this function. If the base64 can't be decoded, the function is simply returned. We want to either not catch the throw inside the function when base64 decoding fails - and assume caller responsibility (bad idea) or catch the whole function internally and re-throw (better idea).

tests/unit_tests/client/proxy/http.cc
```diff
@@ -57,4 +57,31 @@ BOOST_AUTO_TEST_CASE(ok) {
...
 }
 
+BOOST_AUTO_TEST_CASE(JumpServiceI2PAddressHelper) {
+  kovri::client::HTTPMessage tmp;
+  std::string tmp_data = "GET stats.i2p";
+  tmp_data +=
+      // Jump service GET parameter
+      "?i2paddresshelper="
+      // Base64- & URL-encoded I2P address
+ 			"0UVPqAA4xUSfPYPBca24h8fdokhwcJZ4zWvELv-5OsBYTHKtnLzvK7byXtXT~"
```
anonimal: Do you need help setting up clang-format?

src/client/proxy/http.cc
coneiric: Currently, the false return means that the request is a valid HTTP request, but not a valid jump service request. 

There isn't a way to distinguish whether it was:
- intended to be a jump service request (should return a bad_request response)
- a non-jump service request (should continue processing)

So far, I was thinking of moving `HandleJumpService`'s internal regex check into a separate helper function, and performing the check in `CreateHTTPRequest` instead/as well. 

The way it's implemented now, the refactor might be a bit hacky. 

I was thinking short-to-mid-term about entirely rewriting HTTP proxy with `Boost.Beast` or `cpp-netlib`, instead of the custom HTTP parsing code. Do you have advice on doing so, or if it is even a good idea?

tests/unit_tests/client/proxy/http.cc
coneiric: No, lol. Some derpy literal tab character slid its way in somehow.
Fixing it now.

src/client/proxy/http.cc
anonimal: >Currently, the false return means that the request is a valid HTTP request, but not a valid jump service request. At least, that's how I was thinking about it.

But you need to follow the flow. See #814. If what you want is not yet implemented, be sure to implement *before* you PR.

>So far, I was thinking of moving HandleJumpService's internal regex check into a separate helper function, and performing the check in CreateHTTPRequest instead/as well.

The entire proxy needs a rewrite. Do whatever you need to do.

>Do you have advice on doing so, or if it is even a good idea?

1. https://github.com/monero-project/kovri/issues/721
2. Beast doesn't have proxy support out of the box so you're looking at a chunk of work regardless of lib usage
3. Boost.beast is now in 1.66 but default ubuntu boost is *not* 1.66. As such, we can't depend on a lib which has yet to be readily available. This not only applies to linux but all of our platforms on build.getmonero.org

coneiric: > But you need to follow the flow. See #814. If what you want is not yet implemented, be sure to implement before you PR.

This was an oversight on my part, patched now with a fix (at least where parsing is concerned). I'll work on the "save-to-address-book" workflow in another PR. Updated TODO in latest patch.

Will also be working on a PR for #814 next.

Regarding Boost.beast, I'll refactor using cpp-netlib where necessary for now.
Will work on a PR for #721 once Boost.beast is supported on all buildbot platforms.

tests/unit_tests/client/proxy/http.cc
coneiric: Was the first place I looked after your comment, configuring now. :smile:

coneiric:
Coverall coverage decreased, hold on merging until I write unit tests to cover `IsJumpServiceAddress` and `ExtractBase64Address`.

src/client/proxy/http.cc
coneiric: > Also take a look inside this function. If the base64 can't be decoded, the function is simply returned.

I separated base64 parsing + decoding into its own function, and will work on a PR to address the storage issue.

src/client/proxy/http.cc
```diff
@@ -395,50 +424,103 @@ bool HTTPMessage::ExtractIncomingRequest() {
...
   }
-  auto base64 = m_URL.substr(pos + m_JumpService.at(0).size());
-  // We must decode
-  base64 = boost::network::uri::decoded(base64);
-  // Insert into address book
+
+  // final sanity check
+  // TODO(oneiric): may be unnecessary, consider removing
+	if (pos != std::string::npos && pos != 0)
+		return true;
```
coneiric: Another missed formatting error, got through before clang-format configuration.

src/client/proxy/http.cc
anonimal: >I'll refactor using cpp-netlib where necessary for now.

Do NOT. I repeat, DO NOT. Any further cpp-netlib work will be NACK'd. #721 means NOT adding MORE cpp-netlib.

coneiric:
> NACK. Do NOT make formatting changes to work that you have not changed! This is clearly stated in the contributing guide!

The formatting changes are fixed now, as well as the editor config that initially led to the problem.

tests/unit_tests/client/proxy/http.cc
```diff
@@ -57,4 +57,166 @@ BOOST_AUTO_TEST_CASE(ok) {
...
+  tmp.m_Address = "kovri.i2p";
+  // Process non-jump service request
+  BOOST_CHECK(tmp.CreateHTTPRequest());
+}
+
+BOOST_AUTO_TEST_CASE(CreateHTTPRequestJumpServiceNoBase64)
+{
+  kovri::client::HTTPMessage tmp;
+  // Valid jump service key with no base64 destination
+  std::string tmp_data = "GET stats.i2p?i2paddresshelper= HTTP/1.1\r\n\r\n";
```
anonimal: >tmp
>tmp_data

From the [style guide](https://google.github.io/styleguide/cppguide.html):

`In general the actual name of the variable should be descriptive enough to give a good idea of what the variable is used for.`

`tmp` is a "message" (if I wrote that class, it would've not be a called a "message" class because that's not what it is). `tmp_data` is the request. So, for the time being:

```cpp
kovri::client::HTTPMessage message;
std::string const request("GET stats.i2p");
```

for example.

tests/unit_tests/client/proxy/http.cc
```diff
@@ -57,4 +57,166 @@ BOOST_AUTO_TEST_CASE(ok) {
...
+  // Load variables into HTTPMessage
+  BOOST_CHECK(tmp.HandleData(tmp_data));
+  // Mock m_Address for test output
+  tmp.m_Address = "kovri.i2p";
+  // Process invalid jump service request
```
anonimal: >// Load variables into HTTPMessage
>// Mock m_Address for test output
>// Process invalid jump service request

Lots of redundant copypasta.

tests/unit_tests/client/proxy/http.cc
```diff
@@ -57,4 +57,166 @@ BOOST_AUTO_TEST_CASE(ok) {
...
+  BOOST_CHECK(!tmp.CreateHTTPRequest());
+}
```
anonimal: >kovri::client::HTTPMessage tmp;
>std::string tmp_data = "GET stats.i2p";
>tmp.m_Address = "kovri.i2p"; 

All this copypasta within the testcases points out a glaring issue: you need to use a test fixture! See other tests for an example and/or read up on boost.test.

src/client/proxy/http.h
```diff
@@ -242,6 +243,20 @@ class HTTPMessage : public std::enable_shared_from_this<HTTPMessage>{
 
   const unsigned int HEADERBODY_LEN = 2;
   const unsigned int REQUESTLINE_HEADERS_MIN = 1;
+ private:
+  /// @brief Checks if request is a valid jump service request
+  /// @param pos Index of m_JumpService sub-string in m_URL
```
anonimal: >m_JumpService
>m_URL

Using variable names is bad documentation style because A) it doesn't accurate describe the parameter in plain language (which is the point of documentation) and B) the names will need to be changed in any future refactoring (if the refactorer remembers to do so).

src/client/proxy/http.cc
```diff
@@ -390,41 +423,98 @@ bool HTTPMessage::ExtractIncomingRequest() {
...
+bool HTTPMessage::SaveJumpServiceAddress()
+{
+  // TODO(oneiric): move exception handling into callee
+  try
+    {
+      LOG(debug) << "HTTPProxyHandler: inserting " << m_Address
+                 << " into address book";
+      kovri::client::context.GetAddressBook().InsertAddressIntoStorage(
+          m_Address, m_Base64Destination);
+      return true;
```
anonimal: This should go at end of the function.

src/client/proxy/http.cc
```diff
@@ -390,41 +423,98 @@ bool HTTPMessage::ExtractIncomingRequest() {
...
+bool HTTPMessage::ExtractBase64Destination(std::size_t pos)
+{
+  std::size_t base64_size = pos + m_JumpService.at(0).size();
+  if (pos > 0 && base64_size < m_URL.size())
+    {
+      auto base64 = m_URL.substr(base64_size);
+      m_Base64Destination = boost::network::uri::decoded(base64);
+      return true;
+    }
+  else  // invalid index or base64_size too large
```
anonimal: Unnecessary `else`. Simply return false.

src/client/proxy/http.cc
```diff
@@ -390,41 +423,98 @@ bool HTTPMessage::ExtractIncomingRequest() {
...
+    return true;
+  else
+    return false;
+}
+
+// Extract & uri-decode base64-encoded I2P destination
+bool HTTPMessage::ExtractBase64Destination(std::size_t pos)
+{
+  std::size_t base64_size = pos + m_JumpService.at(0).size();
+  if (pos > 0 && base64_size < m_URL.size())
```
anonimal: `pos` is unsigned and will never be negative. Simply test if not null `if (pos && base64_size < M_URL.size())`

src/client/proxy/http.cc
```diff
@@ -390,41 +423,98 @@ bool HTTPMessage::ExtractIncomingRequest() {
...
+        pos = pos1;
+      else
+        pos = pos2;
+    }
+
+  // final sanity check
+  // TODO(oneiric): may be unnecessary, consider removing
+  if (pos != std::string::npos && pos != 0)
+    return true;
+  else
```
anonimal: Unnecessary else. Simply return false.

src/client/proxy/http.cc
```diff
@@ -390,41 +423,98 @@ bool HTTPMessage::ExtractIncomingRequest() {
...
-  // We should ask the user for confirmation before proceeding.
-  // Previous reference: http://pastethis.i2p/raw/pn5fL4YNJL7OSWj3Sc6N/
-  // We *could* redirect the user again to avoid dirtiness in the browser
-  kovri::client::context.GetAddressBook().InsertAddressIntoStorage(
-      m_Address, base64);
-  m_Path.erase(pos);
+
+  // Perform sanity check again to:
+  // - ensure valid jump service request
+  // - update pos with m_JumpService sub-string index (hacky)
```
anonimal: Using a variable name is bad documentation style because A) it doesn't accurate describe the function in plain language (which is the point of documentation) and B) the name will need to be changed in any future refactoring (if the refactorer remembers to do so).

src/client/proxy/http.cc
```diff
@@ -390,41 +423,98 @@ bool HTTPMessage::ExtractIncomingRequest() {
...
+    {
+      auto base64 = m_URL.substr(base64_size);
+      m_Base64Destination = boost::network::uri::decoded(base64);
+      return true;
+    }
+  else  // invalid index or base64_size too large
+    return false;
+}
+
+// Save found address in storage
```
anonimal: Unnecessary as this should be described in header file via Doxygen.

src/client/proxy/http.cc
```diff
@@ -390,41 +423,98 @@ bool HTTPMessage::ExtractIncomingRequest() {
...
+bool HTTPMessage::SaveJumpServiceAddress()
+{
+  // TODO(oneiric): move exception handling into callee
```
anonimal: >callee

I think you mean *caller*.

coneiric:
> This is not an atomic commit

I was unsure how to split the commit, since the changes are fairly small.
I'll break it up like you suggested.

> At first glance, the commit title makes no sense ... you can summarize your work in (at most) just a few lines.

I always have trouble here. Thanks for the advice and links, will definitely help. I get confused about whether the commit message should be a point-by-point summary. Clear now I should write higher-level, more concise summaries.

tests/unit_tests/client/proxy/http.cc
coneiric: That example is much cleaner. I was being a bit lazy here, and copied the surrounding style.
Will go back through, and make the suggested changes.

tests/unit_tests/client/proxy/http.cc
coneiric: Entirely, will fix with a fixture :)

tests/unit_tests/client/proxy/http.cc
coneiric: Entirely, will fix with a fixture :)

src/client/proxy/http.h
coneiric: This felt wrong when I was writing the docs as well.

Will change such instances to the high-level descriptor of the variable's use instead of it's name.

src/client/proxy/http.cc
coneiric: Ok, I'll move the return to the end. Didn't realize the return statement should go outside the try-catch block.

src/client/proxy/http.cc
coneiric: I do, thanks for catching (leftover from copy-pasta).

src/client/proxy/http.cc
coneiric: Great catch. Missed this when refactoring, will fix.

tests/unit_tests/client/proxy/http.cc
```diff
@@ -32,6 +32,12 @@
 #define BOOST_TEST_DYN_LINK
 #include <boost/test/unit_test.hpp>
 #include "client/proxy/http.h"
+#include "tests/unit_tests/client/proxy/http.h"
+
+struct HTTPProxyRequestFixture : public HTTPProxyHeaderFixture
```
anonimal: You don't need to objects when one will suffice.

tests/unit_tests/client/proxy/http.cc
```diff
@@ -32,6 +32,12 @@
 #define BOOST_TEST_DYN_LINK
 #include <boost/test/unit_test.hpp>
 #include "client/proxy/http.h"
+#include "tests/unit_tests/client/proxy/http.h"
```
anonimal: You don't need to create a header file; this is a simple, small unit-test.

tests/unit_tests/client/proxy/http.h
```diff
@@ -0,0 +1,60 @@
...
+
+  // Invalid jump service helper with preceding non-jump-service parameter
+  std::string const invalid_multiple_jump = "GET stats.i2p?some=key&i2paBBresshelper=0UVPqAA4xUSfPYPBca24h8fdokhwcJZ4zWvELv-5OsBYTHKtnLzvK7byXtXT~fOV2pExi8vrkgarGTNDfJbB2KCsdVS3V7qwtTvoCGYyklcDBlJsWMj7H763hEz5rt9SzLkcpwhO3t0Zwe6jXL1UB-QW8KxM30t-ZOfPc6OiJ1QpnE6Bo5OUm6jPurQGXdWCAPio5Z-YnRL46n0IHWOQPYYSStJMYPlPS-S75rMIKbZbEMDraRvSzYAphUaHfvtWr2rCSPkKh3EbrOiBYiAP2oWvAQCsjouPgVF2qwQRnBbiAezHedM2gXzkgIyCV2kGOOcHhiihd~7fWwJOloH-gO78QkmCuY-3kp3633v3MBw7pmABr-XNKWnATZOuf2syWVBZbTnOXsWf41tu6a33HOuNsMxAOUrwbu7QRmT4X8X-EITwNlKN6r1t3uoQ~yZm4RKsJUsBGfVtKl8PBMak3flQAg95oV0OBDGuizIQ9vREOWvPGlQCAXZzEg~cUNbfBQAEAAcAAA%3D%3D HTTP/1.1\r\n\r\n";
+
+  std::string const no_jump_helper = "GET stats.i2p HTTP/1.1\r\n\r\n";
+  std::string const no_jump_base64 = "GET stats.i2p?i2paddresshelper= HTTP/1.1\r\n\r\n";
+};
```
anonimal: Great, you've gotten a better handle on making a fixture. Though this approach technically works, it's overkill and opens up opportunities for developer error.

I'd highly recommend simplifying the approach by using a single string object that appends the test data at the test-case level. Not only will this be more efficient, but also far easier to review because you won't have to hop between the fixture and test-case to see what you are testing.

Example: create a function which inserts the test data in between the first part of the request and last part of the request - and then does the handling. "Insertion" is done by appending the test data to the first part of the request and then appends the end part of the request to the end of the test data. This function exists within the fixture and returns the bool outcome of the internal handlers.

coneiric: Excellent suggestion. I took your advice, and simplified the fixture interface.
Hopefully the fixture functions + enum are straightforward for new devs to understand.

Please let me know if I should refactor further.

coneiric:
Updated with latest recommended changes. Please let me know if I should refactor further.

coneiric:
> Too much time has been spent on this PR. Can you please get it right this time? Thanks.

Agreed, I'll keep working at getting better. Apologies for the list of commit updates, Github hid a number of your comments.

Everything should be fixed now.

src/client/proxy/http.cc
```diff
@@ -313,7 +313,39 @@ bool HTTPMessage::CreateHTTPRequest() {
     // m_ErrorResponse is set in ExtractIncomingRequest
     return false;
   }
-  HandleJumpService();
+
+  if (!IsJumpServiceRequest())
+    {
+      LOG(info) << "HTTPProxyHandler: not a jump service request";
```
anonimal: `info` is for end-user (least technical possible). This looks like a `debug` message.

src/client/proxy/http.cc
```diff
@@ -390,41 +422,94 @@ bool HTTPMessage::ExtractIncomingRequest() {
...
+  // final sanity check
+  if (pos && pos != std::string::npos)
+    return pos;
+
+  return 0;
+}
+
+bool HTTPMessage::ExtractBase64Destination(std::size_t const pos)
+{
+  auto const base64_size = pos + m_JumpService.at(0).size();
```
anonimal: https://github.com/monero-project/kovri/pull/813#discussion_r173045939

src/client/proxy/http.cc
```diff
@@ -390,41 +422,94 @@ bool HTTPMessage::ExtractIncomingRequest() {
...
+  if (pos && base64_size < m_URL.size())
+    {
+      auto base64 = m_URL.substr(base64_size);
```
anonimal: `std::string const base64`

src/client/proxy/http.cc
```diff
@@ -390,41 +422,94 @@ bool HTTPMessage::ExtractIncomingRequest() {
...
-  // We should ask the user for confirmation before proceeding.
-  // Previous reference: http://pastethis.i2p/raw/pn5fL4YNJL7OSWj3Sc6N/
-  // We *could* redirect the user again to avoid dirtiness in the browser
-  kovri::client::context.GetAddressBook().InsertAddressIntoStorage(
-      m_Address, base64);
-  m_Path.erase(pos);
+
+  // Perform sanity check again to:
+  // - ensure valid jump service request
+  auto const pos = IsJumpServiceRequest();
```
anonimal: >auto

`std::size_t`

coneiric:
> Last round of NACKs and then this should be ready to merge.

Thanks for your patience on this PR. I'm determined to get better, and take less of your time on review.

PR #816, opened 2018-02-24
AddressBook: add address to file I/O refactor + add exception handling

coneiric:
---

---

Referencing #305 and anonimal's [advice](https://github.com/monero-project/kovri/pull/813#discussion_r169812054) on exception handling.

Unsure if test case for filesystem error [is appropriate](https://github.com/monero-project/kovri/pull/813#issuecomment-367224433), can remove if desired.

### DIFF
--- a/src/client/address_book/impl.cc
+++ b/src/client/address_book/impl.cc
@@ -444,24 +444,26 @@ std::unique_ptr<const kovri::core::IdentHash> AddressBook::GetLoadedAddressIdent
 // Used only by HTTP Proxy
 void AddressBook::InsertAddressIntoStorage(
     const std::string& address,
-    const std::string& base64) {
-  kovri::core::IdentityEx ident;
+    const std::string& base64)
+{
   try
     {
+      kovri::core::IdentityEx ident;
       ident.FromBase64(base64);
+      if (!m_Storage)
+        m_Storage = GetNewStorageInstance();
+      m_Storage->AddAddress(ident);
+      const auto& ident_hash = ident.GetIdentHash();
+      m_Addresses[address] = ident_hash;
+      LOG(info) << "AddressBook: " << address << "->"
+                << kovri::core::GetB32Address(ident_hash)
+                << " added";
     }
   catch (...)
     {
       m_Exception.Dispatch(__func__);
-      return;
+      throw;
     }
-  if (!m_Storage)
-    m_Storage = GetNewStorageInstance();
-  m_Storage->AddAddress(ident);
-  m_Addresses[address] = ident.GetIdentHash();
-  LOG(info)
-    << "AddressBook: " << address << "->"
-    << GetB32AddressFromIdentHash(ident.GetIdentHash()) << " added";
 }
 
 void AddressBook::Stop() {
--- a/src/client/address_book/storage.cc
+++ b/src/client/address_book/storage.cc
@@ -1,5 +1,5 @@
 /**                                                                                           //
- * Copyright (c) 2013-2017, The Kovri I2P Router Project                                      //
+ * Copyright (c) 2013-2018, The Kovri I2P Router Project                                      //
  *                                                                                            //
  * All rights reserved.                                                                       //
  *                                                                                            //
@@ -65,18 +65,18 @@ bool AddressBookStorage::GetAddress(
   return true;
 }
 
-void AddressBookStorage::AddAddress(
-    const kovri::core::IdentityEx& address) {
-  auto filename = GetAddressesPath() / (address.GetIdentHash().ToBase32() + ".b32");
-  std::ofstream file(filename.string(), std::ofstream::binary);
-  if (!file)
-    LOG(error) << "AddressBookStorage: can't open file " << filename;
+void AddressBookStorage::AddAddress(const kovri::core::IdentityEx& address)
+{
+  auto filename =
+      GetAddressesPath() / (address.GetIdentHash().ToBase32() + ".b32");
+  kovri::core::OutputFileStream file(filename.string(), std::ofstream::binary);
+  if (!file.Good())
+    throw std::runtime_error("failed to open file for address writing");
   const std::size_t len = address.GetFullLen();
-  auto buf = std::make_unique<std::uint8_t[]>(len);
-  // For sanity, the validity of identity length is incumbent upon the parent caller.
-  // TODO(unassigned): triple check that this is the case
-  address.ToBuffer(buf.get(), len);
-  file.write(reinterpret_cast<char *>(buf.get()), len);
+  std::vector<std::uint8_t> buf(len);
+  address.ToBuffer(buf.data(), buf.size());
+  if (!file.Write(buf.data(), buf.size()))
+    throw std::runtime_error("failed to write address file");
 }
 
 /**


tests/unit_tests/client/address_book/impl.cc
```diff
@@ -240,6 +240,54 @@ BOOST_AUTO_TEST_CASE(PGPClearSign)
...
+          "KH78sRQmzWnv7dbhNodMbpO6xsf1vENf6bMRzqD5vgHEHZu2aSoNuPyYxDU1eM6--"
+          "61b2xp9mt1k3ud-5WvPVg89RaU9ugU5cxaHgR927lHMCAEU2Ax~"
+          "zUb3DbrvgQBOTHnJEx2Fp7pOK~"
+          "PnP6ylkYKQMfLROosLDXinxOoSKP0UYCh2WgIUPwE7WzJH3PiJVF0~WZ1dZ9mg00c~"
+          "gzLgmkOxe1NpFRNg6XzoARivNVB5NuWqNxr5WKWMLBGQ9YHvHO1OHhUJTowb9X90BhtH"
+          "nLK2"
+          "AHwO6fV-iHWxRJyDabhSMj1kuYpVUBQAEAAcAAA=="};
+  BOOST_CHECK_THROW(
+      book.InsertAddressIntoStorage(address, base64), std::runtime_error);
+}
```
anonimal: These tests are currently useless because:

  - You're intentionally creating the filesystem failure within the implementation by ignoring the fact that a file doesn't exist. This test-case will always be hopeless because it **should not rely on the filesystem** as I have [already said](https://github.com/monero-project/kovri/pull/813#issuecomment-367224433).
  - You're simply testing our decoder for which we already have tests.

Other issues:

- Using curly-braces to initialize a string-literal ctor?! The literal is not a list nor aggregate nor integral. Use `()` regardless of what the standard allows. Doing so is consistent with rest of this codebase and nearly every other c++ program out there.
- Use const correctness.
- Keep the base64 string on one line. No one wants to see a line split up like that and clang-format isn't bright when it comes to this issue. This is not your fault though because the guide clearly says to run clang-format. We should make an addendum to the guide.

src/client/address_book/storage.cc
```diff
@@ -71,17 +71,33 @@ bool AddressBookStorage::GetAddress(
 
 void AddressBookStorage::AddAddress(const kovri::core::IdentityEx& address)
 {
-  auto filename =
-      GetAddressesPath() / (address.GetIdentHash().ToBase32() + ".b32");
-  std::ofstream file(filename.string(), std::ofstream::binary);
-  if (!file)
```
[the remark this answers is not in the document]
coneiric: I wasn't aware of this utility library.

Am now attempting to use [OutputFileStream](https://github.com/monero-project/kovri/blob/master/src/core/util/filesystem.h#L307) here.

There is no direct replacement for the boolean check.

IIUC, an equivalent check using OutputFileStream would look like:
```
auto filename =
      GetAddressesPath() / (address.GetIdentHash().ToBase32() + ".b32");
  kovri::core::OutputFileStream file(filename.string(), std::ofstream::binary);
  if (file.Bad() || file.Fail())
```

Does this look reasonable to you, or should I approach it another way?

src/client/address_book/storage.cc
```diff
@@ -71,17 +71,33 @@ bool AddressBookStorage::GetAddress(
...
-  auto buf = std::make_unique<std::uint8_t[]>(len);
-  // For sanity, the validity of identity length is incumbent upon the parent caller.
-  // TODO(unassigned): triple check that this is the case
-  address.ToBuffer(buf.get(), len);
-  file.write(reinterpret_cast<char*>(buf.get()), len);
+  try
+    {
+      auto filename =
+          GetAddressesPath() / (address.GetIdentHash().ToBase32() + ".b32");
+      if (!boost::filesystem::exists(filename))
```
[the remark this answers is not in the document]
coneiric: I was mistaken in my attempt to use Boost.Filesystem here. This will be removed in my revisions.

src/client/address_book/storage.cc
```diff
@@ -71,17 +71,33 @@ bool AddressBookStorage::GetAddress(
...
+  catch (boost::filesystem::filesystem_error& ex)
+    {
+      LOG(error) << ex.what();
+      std::rethrow_exception(std::make_exception_ptr(ex));
+    }
+  catch (...)
+    {
+      LOG(error) << "AddressBookStorage: unknown error occured in " << __func__;
+      std::rethrow_exception(std::current_exception());
+    }
```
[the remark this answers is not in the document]
coneiric: I will remove the try-catch block from this function, and use the dispatcher for exception handling.

tests/unit_tests/client/address_book/impl.cc
coneiric: > These tests are currently useless because

I'll remove these tests, and try to come up with new cases.

> Keep the base64 string on one line. No one wants to see a line split up like that and clang-format isn't bright when it comes to this issue.

Noted. I've turned off auto-clang-formatting on save to hopefully cut-back on the formatting issues.

src/client/address_book/impl.cc
```diff
@@ -501,22 +501,27 @@ void AddressBook::InsertAddressIntoStorage(
...
+      LOG(error) << "AddressBook: " << ex.what();
+      std::rethrow_exception(std::make_exception_ptr(ex));
     }
-  catch (...)
+  catch (...)  // Unknown exception
     {
-      m_Exception.Dispatch(__func__);
-      return;
+      LOG(error) << "AddressBook: unknown error in " << __func__;
+      std::rethrow_exception(std::current_exception());
```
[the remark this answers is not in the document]
coneiric: I'll change the main catch block to use the exception disptacher + re-throw, and remove the specific catch blocks.

coneiric:
> First and foremost, PLEASE ensure your work builds and works BEFORE you PR. A vast majority of your PRs have been NACKed because they don't build or because they break functionality. This is extremely frustrating for any reviewer and my patience is wearing thin.

Honest apologies for straining your patience, this is not my intention. I built and re-built, ran and re-ran tests before PR'ing.

After looking at another developer's [buildbot setup](https://github.com/monero-project/meta/issues/184), I'll do the same to avoid these annoying build issues.

> DO NOT make completely unrelated formatting changes!

I understand your frustration here, and never intended any grief.

Perhaps there could be a note to avoid turning on auto-formatting for entire files, unless the entire file is new work? 

Both the [style guide](https://getkovri.org/docs.html) and [clang guide](https://clang.llvm.org/docs/ClangFormat.html#vim-integration) have instructions for running `clang-format` over the entire file. In the clang guide, there are also instructions for using `clang-format` selectively.

Without a warning in the Kovri style guide, using clang-format selectively seems error-prone. Seemingly, the best way to ensure all changes get formatted correctly, at least to me, was to turn on the feature automatically during save.

I've turned off this feature now, and will manually run `clang-format` over only the lines related to my changes.

> If you're going to contribute PRs you must do regular development with boost 1.58.

This is entirely an oversight on my part, and I will switch to building against boost 1.58. I'm new to cross-platform development, and appreciate your time + effort helping me along.

src/client/address_book/storage.cc
```diff
@@ -68,15 +69,20 @@ bool AddressBookStorage::GetAddress(
...
     LOG(error) << "AddressBookStorage: can't open file " << filename;
   const std::size_t len = address.GetFullLen();
   auto buf = std::make_unique<std::uint8_t[]>(len);
   // For sanity, the validity of identity length is incumbent upon the parent caller.
   // TODO(unassigned): triple check that this is the case
   address.ToBuffer(buf.get(), len);
-  file.write(reinterpret_cast<char *>(buf.get()), len);
+  if (!file.Write(buf.get(), len))
+    {
+      core::Exception ex;
```
anonimal: This should be a member to `AddressBookStorage`. The class acts on enough data to warrant it's own exception member.

src/client/address_book/storage.cc
```diff
@@ -68,15 +69,20 @@ bool AddressBookStorage::GetAddress(
 void AddressBookStorage::AddAddress(
     const kovri::core::IdentityEx& address) {
   auto filename = GetAddressesPath() / (address.GetIdentHash().ToBase32() + ".b32");
-  std::ofstream file(filename.string(), std::ofstream::binary);
-  if (!file)
+  kovri::core::OutputFileStream file(filename.string(), std::ofstream::binary);
+  if (file.Bad() || file.Fail())
```
anonimal: `if (!file.Good())`

src/client/address_book/storage.cc
```diff
@@ -68,15 +69,20 @@ bool AddressBookStorage::GetAddress(
...
     LOG(error) << "AddressBookStorage: can't open file " << filename;
```
anonimal: If we can't open a file for writing, then there's no point in trying to write below. This should either throw or return (ideally throw).

src/client/address_book/storage.cc
```diff
@@ -68,15 +69,20 @@ bool AddressBookStorage::GetAddress(
...
   const std::size_t len = address.GetFullLen();
   auto buf = std::make_unique<std::uint8_t[]>(len);
   // For sanity, the validity of identity length is incumbent upon the parent caller.
   // TODO(unassigned): triple check that this is the case
   address.ToBuffer(buf.get(), len);
-  file.write(reinterpret_cast<char *>(buf.get()), len);
+  if (!file.Write(buf.get(), len))
```
anonimal: >const std::size_t len = address.GetFullLen();
>auto buf = std::make_unique<std::uint8_t[]>(len);
>address.ToBuffer(buf.get(), len);
>if (!file.Write(buf.get(), len))

While you're here (though it's not mandatory for this PR), since this is a refactor PR, you can refactor the buffer into a vector so then you'll only need to use `buf.data()` and `buf.size()`.

src/client/address_book/impl.cc
```diff
@@ -444,24 +444,25 @@ std::unique_ptr<const kovri::core::IdentHash> AddressBook::GetLoadedAddressIdent
...
   try
     {
+      kovri::core::IdentityEx ident;
       ident.FromBase64(base64);
+      if (!m_Storage)
+        m_Storage = GetNewStorageInstance();
+      m_Storage->AddAddress(ident);
+      m_Addresses[address] = ident.GetIdentHash();
+      LOG(info) << "AddressBook: " << address << "->"
+                << ident.GetIdentHash().ToBase32() << " added";
```
anonimal: This is *not* the same as `GetB32AddressFromIdentHash(ident.GetIdentHash())`. Look carefully.

anonimal:
>Perhaps there could be a note to avoid turning on auto-formatting for entire files, unless the entire file is new work?

Yes, and we should also re-visit our clang-format configuration to see if we can fix the long string problem.

>Without a warning in the Kovri style guide, using clang-format selectively seems error-prone. Seemingly, the best way to ensure all changes get formatted correctly, at least to me, was to turn on the feature automatically during save.

Please feel free to PR.

>using clang-format selectively seems error-prone

I haven't had a problem running `ctrl+k` on selected text, but I use vim. I would like a more automated approach but I think that will only come with time - otherwise there would be a ton of formatting-only changes. That was ok when we first forked but now it's a bit unwarranted because the real incentive is to completely remove the old codebase.

>I'm new to cross-platform development, and appreciate your time + effort helping me along.

You're welcome, and I'm glad to see that you're diving right in and learning quickly.

src/client/address_book/storage.cc
```diff
@@ -65,18 +67,34 @@ bool AddressBookStorage::GetAddress(
...
-  if (!file)
-    LOG(error) << "AddressBookStorage: can't open file " << filename;
+void AddressBookStorage::AddAddress(const kovri::core::IdentityEx& address)
+{
+  auto filename =
+      GetAddressesPath() / (address.GetIdentHash().ToBase32() + ".b32");
+  kovri::core::OutputFileStream file(filename.string(), std::ofstream::binary);
+  if (!file.Good())
+    {
+      m_Exception.Dispatch(__func__);
```
anonimal: This is useless because nothing is thrown before it. Exception dispatchers are meant to be used when catching. If you do more than one check, catching would be appropriate. Otherwise, leave it to the caller.

src/client/address_book/storage.cc
```diff
@@ -65,18 +67,34 @@ bool AddressBookStorage::GetAddress(
...
+  if (len < kovri::core::DEFAULT_IDENTITY_SIZE
+      || len > kovri::core::MAX_IDENTITY_SIZE)
+    {
+      m_Exception.Dispatch(__func__);
+      throw std::length_error("invalid Identity length");
+    }
+  else
+    {
+      std::vector<std::uint8_t> buf(len);
+      address.ToBuffer(buf.data(), len);
```
anonimal: Like I said, `buf.size()`. If len is ever refactored, only 1 line will change instead of 3.

src/client/address_book/storage.cc
```diff
@@ -65,18 +67,34 @@ bool AddressBookStorage::GetAddress(
...
+      if (!file.Write(buf.data(), len))
```
anonimal: Like I said, `buf.size()`. If len is ever refactored, only 1 line will change instead of 3.

src/client/address_book/storage.cc
```diff
@@ -65,18 +67,34 @@ bool AddressBookStorage::GetAddress(
...
+{
+  auto filename =
+      GetAddressesPath() / (address.GetIdentHash().ToBase32() + ".b32");
+  kovri::core::OutputFileStream file(filename.string(), std::ofstream::binary);
+  if (!file.Good())
+    {
+      m_Exception.Dispatch(__func__);
+      throw std::runtime_error("failed to open file for writing");
+    }
+  // For sanity, triple-check the validity of the identity length
```
anonimal: This is entirely unnecessary, this check should be removed. Identity checks must be done within identity code.

src/client/address_book/storage.cc
```diff
@@ -65,18 +67,34 @@ bool AddressBookStorage::GetAddress(
...
   const std::size_t len = address.GetFullLen();
-  auto buf = std::make_unique<std::uint8_t[]>(len);
-  // For sanity, the validity of identity length is incumbent upon the parent caller.
-  // TODO(unassigned): triple check that this is the case
-  address.ToBuffer(buf.get(), len);
-  file.write(reinterpret_cast<char *>(buf.get()), len);
+  if (len < kovri::core::DEFAULT_IDENTITY_SIZE
+      || len > kovri::core::MAX_IDENTITY_SIZE)
```
anonimal: This is completely irrelevant to addressbook: `Offline only; never used in Key Certificates for Router Identities or Destinations`. The excess len is also computed within the identity implementation. Any further checks can be done during the indentity rewrite.

src/client/address_book/storage.cc
```diff
@@ -65,18 +67,34 @@ bool AddressBookStorage::GetAddress(
...
+      m_Exception.Dispatch(__func__);
+      throw std::length_error("invalid Identity length");
+    }
+  else
+    {
+      std::vector<std::uint8_t> buf(len);
+      address.ToBuffer(buf.data(), len);
+      if (!file.Write(buf.data(), len))
+        {
+          m_Exception.Dispatch(__func__);
```
anonimal: This is useless because nothing is thrown before it. Exception dispatchers are meant to be used when catching. If you do more than one check, catching would be appropriate. Otherwise, leave it to the caller.

src/client/address_book/storage.cc
```diff
@@ -65,18 +67,34 @@ bool AddressBookStorage::GetAddress(
...
-  // TODO(unassigned): triple check that this is the case
-  address.ToBuffer(buf.get(), len);
-  file.write(reinterpret_cast<char *>(buf.get()), len);
+  if (len < kovri::core::DEFAULT_IDENTITY_SIZE
+      || len > kovri::core::MAX_IDENTITY_SIZE)
+    {
+      m_Exception.Dispatch(__func__);
+      throw std::length_error("invalid Identity length");
+    }
+  else
```
anonimal: Remove this else statement, it's unnecessary.

src/client/address_book/storage.h
```diff
@@ -152,6 +152,8 @@ class AddressBookStorage : public AddressBookDefaults {
   boost::filesystem::path GetAddressesPath() const {
     return core::GetPath(core::Path::AddressBook) / "addresses";
   }
+
+  core::Exception m_Exception;
```
anonimal: If you end up not expanding its use-case, remove the member.

anonimal:
Once you've made the corrections, please reference this PR number in the bodies of the commit messages.

coneiric:
> Once you've made the corrections, please reference this PR number in the bodies of the commit messages.

All done :smile: Thanks for the notes on improving the PR.

Was thinking of refactoring  `AddressBook::SaveSubscription` and `AddressBookStorage::Save` in a similar way. Should those changes go into their own PR, or added as commits here?

src/client/address_book/storage.h
```diff
@@ -1,5 +1,5 @@
 /**                                                                                           //
- * Copyright (c) 2013-2017, The Kovri I2P Router Project                                      //
+ * Copyright (c) 2013-2018, The Kovri I2P Router Project                                      //
```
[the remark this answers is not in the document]
coneiric: Currently: nothing.

Leftover from previous edits, figured leaving the date updated would be a good thing.
I can revert to the previous date it if you want.

src/client/address_book/impl.cc
```diff
@@ -444,24 +444,26 @@ std::unique_ptr<const kovri::core::IdentHash> AddressBook::GetLoadedAddressIdent
...
+    const std::string& base64)
+{
   try
     {
+      kovri::core::IdentityEx ident;
       ident.FromBase64(base64);
+      if (!m_Storage)
+        m_Storage = GetNewStorageInstance();
+      m_Storage->AddAddress(ident);
+      auto const ident_hash = ident.GetIdentHash();
```
anonimal: Please use appropriate qualifier `auto& const`

coneiric: Not being a smartass, but the [C++ ISO](https://isocpp.org/wiki/faq/const-correctness#const-ref-nonsense) site calls this pattern nonsense:
```
 But that is redundant — references are always const ... “X& const x” is functionally equivalent to “X& x”
```

So would you like me to remove the const qualifier, and add the reference?

anonimal: >references are always const

Yes, but this is taken out of context when using `auto` as an lvalue. Case in point:

```cpp
#include <iostream>

class Test
{
 public:
  Test() : m_test("test") {}

  const std::string& func() const
  {
    return m_test;
  }

 private:
  std::string m_test;
};

int main()
{
  Test t;

  std::cout << t.func() << std::endl;
  //const auto a = t.func();  // Fortunately, won't compile
  //auto const a = t.func();  // Fortunately, won't compile
  //auto& a = t.func();  // Fortunately, won't compile
  //const auto& a = t.func();  // Fortunately, won't compile
  //auto& const a = t.func();  // Fortunately, won't compile
  auto a = t.func();  // Unfortunately, will compile

  a = "uh oh";
  std::cout << a << std::endl;

  return 0;
}
```

When one uses `auto`, one mustn't forget the modifiers when needed. As for the committee, they have a mostly valid point, but:

- For this review, ensuring that the modifier is in place was the biggest point
- If the API wasn't properly made (i.e., `GetIdentHash` did *not* return const reference), then `error: ‘const’ qualifiers cannot be applied to ‘auto&’` would fortunately arise when attempting to build with `auto& const` - thus quickly pointing out the API issue. The same couldn't be said with `auto const` or `const auto` (which is what you originally had) or `const auto&`.

But, of course, this is all arguable so, whatever works should suffice (just don't alter the member or make a copy).

PR #810, opened 2018-02-16
ByteStream security fix + new additions, SSU fix + refactoring

anonimal:
---
**By submitting this pull-request, I confirm the following:**

- I have read and understood the contributor guide in [kovri-docs](https://github.com/monero-project/kovri-docs).
- I have checked that another pull-request for this purpose does not exist.
- I have considered and confirmed that this submission will be valuable to others.
- I accept that this submission may not be used and that this pull-request may be closed by the will of the maintainer.
- I give this submission freely under the BSD 3-clause license.
---

Known issues to be resolved before merging:

- [x] ~1. Ubuntu (gcc 5.4.0-6) unit-tests pass but SIGSEGV because of a cryptopp issue that I will soon open. Cryptopp apparently doesn't like how we construct our internal-buffer for the "stream". I cannot reproduce this issue on Arch (gcc 7.3.0 / clang 5.0.1) but will open an issue upstream regardless.~ ~~**Edit**: Buildbot cannot reproduce so I can assume for now that the issue is specific to my personal build machine.~~ **Edit:** this appears to affect our ARMv7 machine [https://build.getmonero.org/builders/kovri-all-ubuntu-arm7/builds/613/steps/test/logs/stdio](https://build.getmonero.org/builders/kovri-all-ubuntu-arm7/builds/613/steps/test/logs/stdio).
- [x] 2. Internal-buffer "stream" read test-cases fail on OS X 10.12 `Apple LLVM version 8.0.0 (clang-800.0.42.1)`. I don't know how recently that machine has been updated though since many of our build machines are going out of date, I will open an issue for this too. **Edit**: opened https://github.com/monero-project/meta/issues/179.

In all, I'm becoming very irritated with the bytestream implementation / concept as a whole. While #140 has had the best of intentions, the bytestream is building on - and revolving around - previous spaghetti-code pointer concepts rather then a usable, modular, OOP design. The current bytestream design also prevents us from managing memory in a more guaranteed and interoperable way. This issue too must be opened in the issue tracker. **Edit**: opened #811.

PR #820, opened 2018-03-01
SSU: fixes, rewrites, refactors, docs

anonimal:
---

---

WIP for #140.

Live tests + unit-tests pass. I'd like to run the router for a while to observe new assertion behavior (they lead to some of the various fixes).

There are various existing issues and work to be done outside of the TODOs. I'll leave a note in #140.

### DIFF
--- a/src/core/router/transports/ssu/packet.cc
+++ b/src/core/router/transports/ssu/packet.cc
@@ -182,7 +182,11 @@ std::uint8_t const* SSUSessionRequestPacket::GetDhX() const {
 
 void SSUSessionRequestPacket::SetIPAddress(
     std::uint8_t* address,
-    std::size_t size) {
+    const std::uint8_t size)
+{
+  assert(size == 4 || size == 16);
+  if (size != 4 && size != 16)
+    throw std::length_error("invalid IP address size");
   m_IPAddress = address;
   m_IPAddressSize = size;
 }
@@ -191,7 +195,8 @@ std::uint8_t const* SSUSessionRequestPacket::GetIPAddress() const {
   return m_IPAddress;
 }
 
-std::size_t SSUSessionRequestPacket::GetIPAddressSize() const {
+std::uint8_t SSUSessionRequestPacket::GetIPAddressSize() const noexcept
+{
   return m_IPAddressSize;
 }
 
@@ -219,7 +224,11 @@ std::uint8_t const* SSUSessionCreatedPacket::GetDhY() const {
 
 void SSUSessionCreatedPacket::SetIPAddress(
     std::uint8_t* address,
-    std::size_t size) {
+    const std::uint8_t size)
+{
+  assert(size == 4 || size == 16);
+  if (size != 4 && size != 16)
+    throw std::length_error("invalid IP address size");
   m_IPAddress = address;
   m_AddressSize = size;
 }
@@ -228,7 +237,8 @@ std::uint8_t const* SSUSessionCreatedPacket::GetIPAddress() const {
   return m_IPAddress;
 }
 
-std::size_t SSUSessionCreatedPacket::GetIPAddressSize() const {
+std::uint8_t SSUSessionCreatedPacket::GetIPAddressSize() const noexcept
+{
   return m_AddressSize;
 }
 
@@ -342,7 +352,11 @@ std::uint32_t SSURelayRequestPacket::GetRelayTag() const {
 
 void SSURelayRequestPacket::SetIPAddress(
     std::uint8_t* address,
-    std::size_t size) {
+    const std::uint8_t size)
+{
+  assert(!size || size == 4);  // See spec for details
+  if (size && size != 4)
+    throw std::length_error("invalid IP address size");
   m_IPAddress = address;
   m_IPAddressSize = size;
 }
@@ -418,7 +432,11 @@ std::uint32_t SSURelayResponsePacket::GetNonce() const {
 
 void SSURelayResponsePacket::SetIPAddressAlice(
     std::uint8_t* address,
-    std::size_t size) {
+    const std::uint8_t size)
+{
+  assert(size == 4 || size == 16);
+  if (size != 4 && size != 16)
+    throw std::length_error("invalid IP address size");
   m_IPAddressAlice = address;
   m_IPAddressAliceSize = size;
 }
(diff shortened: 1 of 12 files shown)


anonimal:
Assertion fires when parsing Data fragment from certain router(s) (not applicable for all routers, only fires when reseeding from TBD hosts):

```
#7  0x00000000011063bd in kovri::core::SSUPacketParser::ParseFragment (this=0x7fffeb7fd4e0) at /home/anonimal/kovri/src/core/router/transports/ssu/packet.cc:694
        fragment = {m_MessageID = 208 '\320', m_Number = 210 '\322', m_IsLast = 127, m_Size = 140737144411424, m_Data = 0x7fffeb7fd2e0 "\260\323\177\353\377\177"}
        info_buf = {_M_elems = "P\322", <incomplete sequence \353>}
        fragment_info = 0
```

That size looks nasty. I'll start looking at this either Friday or starting Monday.

src/core/router/transports/ssu/session.cc
```diff
@@ -620,56 +621,71 @@ void SSUSession::ProcessRelayRequest(
...
-    FillHeaderAndEncrypt(
-        relay_request,
-        buf.data(),
-        96,
-        introducer_key,
-        iv.data(),
-        introducer_key);
-  }
+    const std::uint32_t introducer_tag,
+    const std::uint8_t* introducer_key)
```
coneiric: Noob question about const-correctness: is it a good idea to make `introducer_key` `const uint8_t* const`?

When is it necessary to also make the pointer const?

src/core/router/transports/ssu/session.cc
```diff
@@ -620,56 +621,71 @@ void SSUSession::ProcessRelayRequest(
...
+  // TODO(unassigned): Endianness is not spec-defined, assuming BE
+
+  // Skip header (written later)
+  message.SkipBytes(SSUSize::HeaderMin);
+
+  // Intro tag
+  message.Write<std::uint32_t>(introducer_tag);
+
+  // Address, port, and challenge (see SSU spec)
+  message.SkipBytes(4);
```
coneiric: IIUC, in the Java spec, the ip-size byte is followed by that many bytes of the actual IP.
Does this need to be accounted for here, or am I missing something?

src/core/router/transports/ssu/session.cc
```diff
@@ -716,73 +716,85 @@ void SSUSession::ProcessRelayResponse(SSUPacket* pkt) {
...
-        from);
-  }
+    const boost::asio::ip::udp::endpoint& to)
+{
+  // Charlie's address must always be IPv4
+  if (!to.address().is_v4())
+    {
+      LOG(error) << "SSUSession:" << GetFormattedSessionInfo() << __func__
+                 << ": Charlie's must use IPv4";
+      // TODO(anonimal): don't throw?...
```
coneiric: FWIW, one obvious benefit to throwing is the caller knowing the SSU session failed. What are the downsides to throwing here?

src/core/router/transports/ssu/session.cc
```diff
@@ -361,89 +364,128 @@ void SSUSession::SendSessionRequest() {
...
+                     << "SessionCreated signature verification failed";
+          // TODO(anonimal): review if Java routers resend the message on failure.
+          //   Instead of immediately resetting session key, we can explore ways
+          //   to observe and mitigate potential attacks. Another possible case
+          //   for failure:
+          //     "If Bob's NAT/firewall has mapped his internal port to a
+          //     different external port, and Bob is unaware of it, the
+          //     verification by Alice will fail."
+          m_IsSessionKey = false;
+          return;  // TODO(anonimal): throw/assert?
```
coneiric: Same question about throwing as for SendRelayResponse.
If retrying on failure, would an exception help signal to re-attempt the session?

src/core/router/transports/ssu/session.cc
```diff
@@ -488,88 +488,146 @@ void SSUSession::ProcessSessionCreated(const SSUPacket* packet)
...
+              relay_tag = 1;
+            }
+          m_Server.AddRelay(relay_tag, m_RemoteEndpoint);
+        }
+      message.SetRelayTag(relay_tag);
+      data.Write<std::uint32_t>(relay_tag);
+
+      // Our (Bob's) signed-on time
+      message.SetSignedOnTime(core::GetSecondsSinceEpoch());
+      data.Write<std::uint32_t>(message.GetSignedOnTime());
```
coneiric: Do we need to be concerned about thread safety here? Is there any danger of `GetSignedOnTime` returning/being called before `SetSignedOnTime` returns?

src/core/router/transports/ssu/packet.cc
```diff
@@ -104,7 +104,7 @@ void SSUHeader::SetRekey(
 
 void SSUHeader::SetExtendedOptionsData(
     std::uint8_t* data,
-    std::size_t size) {
+    std::uint8_t size) {
```
coneiric: Would it be useful for `data` to be `const std::uint8_t* const` and `size` to be `const`?

src/core/router/transports/ssu/session.cc
anonimal: >is it a good idea to make introducer_key const uint8_t* const

That's not even correct syntax - and won't even build. Even if it could, it's not necessary. 

>When is it necessary to also make the pointer const?

lmgtfy.

src/core/router/transports/ssu/session.cc
anonimal: >am I missing something?

Yes, look more carefully.

- `96 bytes (no Alice IP included)`
- `In the current implementation, the IP length is always 0 and the port is always 0`.
- `Challenge is unimplemented, challenge size is always zero`

src/core/router/transports/ssu/session.cc
anonimal: This will be addressed once a new message generator/dispatcher is implemented.

src/core/router/transports/ssu/session.cc
anonimal: This will be addressed once a new message generator/dispatcher is implemented.

src/core/router/transports/ssu/session.cc
anonimal: ...

No. This object is created locally, within the function, within the thread. If the 2nd line was executed before the 1st, threading would be the last of our concerns since the compiler would be broken beyond recognition.

Look at the transports impl for answers to future threading questions.

src/core/router/transports/ssu/packet.cc
anonimal: https://github.com/monero-project/kovri/issues/140#issuecomment-369775064.

Also, the const integral can be dealt with then too as it's not worth a rebase and re-trigger of the entire build system (we don't manipulate the data within the function).

anonimal:
From https://github.com/monero-project/kovri/pull/820#issuecomment-369778881:

>That size looks nasty

`m_Size` is merely an unused, uninitialized member that should've been removed in #810. It has no bearing on the assertion.

As for the assertion, it's very odd and also difficult to reproduce. The issue looks unrelated to this PR and appears to have been a long-standing, pre-existing issue that was exposed with #810. I will open a new ticket with details.

This PR can be merged.

src/core/router/transports/ssu/session.cc
coneiric: Knew I had to be missing something, thanks for explaining.

src/core/router/transports/ssu/packet.cc
coneiric: Understood, this seemed like a minor point to me as well (not worth the rebuild). Thanks for the clarification.

Issue #777, opened 2017-12-24
Expand --host to take both ipv4 and ipv6 addresses


anonimal:
---
**By submitting this issue, I confirm the following:**

- I have read and understood the contributor guide in [kovri-docs](https://github.com/monero-project/kovri-docs).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.
---

`--host` is currently limited to taking a single address for RI. Let's expand the option (comma separated?) to take both ipv4 + ipv6 addresses and implement the RI accordingly.

anonimal:
Resolved in #831.

Issue #627, opened 2017-04-21
Core: router context + RI rewrite/refactor


anonimal:
WIP as noted in #617 #621 #623 #626. Ticketing for housekeeping.

WIP in pursuit of isolating #187 (and because RI impl was in bad shape / unmaintainable).

---
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.
---

anonimal:
Added router context because it's in just as bad shape.

coneiric:
Small cheat-sheet for RI sizes: [router_info_size.md](https://gist.github.com/coneiric/a2bc40172cd7a5465ba8e2b527b1c0c5)

Some of the maximums are the max allowed by the size byte(s), when the specification doesn't otherwise limit. Should we set a cap in those cases?

anonimal:
Cap as in capability or cap as in top/hat limit?

anonimal:
>router_info_size.md

I think this would be great in the moneropedia (see monero-site) or perhaps something in kovri-docs or kovri-site?

coneiric:
> Cap as in capability or cap as in top/hat limit?

Cap as in limit to total `RouterInfo` size we'll accept. Otherwise, that ~17MB max total size looks like a potential DoS vector.

> I think this would be great in the moneropedia (see monero-site) or perhaps something in kovri-docs or kovri-site?

Sure, I can make a PR to monero-site and kovri-docs. Is kovri-site still pulling from kovri-docs? If not, I can make a PR there as well.

anonimal:
>Is kovri-site still pulling from kovri-docs?

Yes, it should be. This is still open though https://github.com/monero-project/kovri-site/issues/73.

anonimal:
https://github.com/coneiric/kovri/pull/2

anonimal:
https://github.com/coneiric/kovri/pull/4

anonimal:
NOTICE: THIS ISSUE HAS BEEN MOVED TO GitLab. Please continue the discussion there. See https://github.com/monero-project/kovri/issues/1013 for details.

PR #831, opened 2018-03-09
RouterInfo: Expand --host to take both ipv4 and ipv6 addresses.

brbzull0:
---

---

Add to the existing --host parameter the capabilities to handle both ipv4 and ipv6 address
using a "," separator.

See #777

### DIFF
--- a/src/core/router/info.cc
+++ b/src/core/router/info.cc
@@ -669,14 +669,12 @@ void RouterInfo::DisableV6()
   m_SupportedTransports &= ~SupportedTransport::SSUv6;
 
   // Remove addresses in question
-  for (std::size_t i = 0; i < m_Addresses.size(); i++)
-    {
-      if (m_Addresses[i].host.is_v6())
-        {
-          LOG(debug) << "RouterInfo: " << __func__ << ": removing address";
-          m_Addresses.erase(m_Addresses.begin() + i);
-        }
-    }
+  m_Addresses.erase(
+      std::remove_if(
+          std::begin(m_Addresses),
+          std::end(m_Addresses),
+          [](const auto& address) { return address.host.is_v6(); }),
+      std::end(m_Addresses));
 }
 
 void RouterInfo::Update(const std::uint8_t* buf, std::uint16_t len)
--- a/src/core/router/info.cc
+++ b/src/core/router/info.cc
@@ -60,7 +60,7 @@ RouterInfo::RouterInfo() : m_Exception(__func__), m_Buffer(nullptr)  // TODO(ano
 
 RouterInfo::RouterInfo(
     const core::PrivateKeys& keys,
-    const std::pair<std::string, std::uint16_t>& point,
+    const std::vector<std::pair<std::string, std::uint16_t>>& points,
     const std::pair<bool, bool>& has_transport,
     const std::uint8_t caps)
     : m_Exception(__func__), m_RouterIdentity(keys.GetPublic())
@@ -76,14 +76,19 @@ RouterInfo::RouterInfo(
   // Set default caps
   SetCaps(caps);
 
-  // Set default transports
-  if (has_transport.first)
-    AddAddress(std::make_tuple(Transport::NTCP, point.first, point.second));
+  for (const auto& point : points)
+    {
+      // Set default transports
+      if (has_transport.first)
+        AddAddress(std::make_tuple(Transport::NTCP, point.first, point.second));
+
+      if (has_transport.second)
+        AddAddress(
+            std::make_tuple(Transport::SSU, point.first, point.second), hash);
+    }
 
   if (has_transport.second)
     {
-      AddAddress(
-          std::make_tuple(Transport::SSU, point.first, point.second), hash);
       SetCaps(
           m_Caps | core::RouterInfo::Cap::SSUTesting
           | core::RouterInfo::Cap::SSUIntroducer);
--- a/src/core/router/info.h
+++ b/src/core/router/info.h
@@ -435,13 +435,13 @@ class RouterInfo : public RouterInfoTraits, public RoutingDestination
   ~RouterInfo();
 
   /// @brief Create RI with standard defaults
-  /// @param point Local hostname/ip address + port
-  /// @param has_transport Supports NTCP, SSU
   /// @param keys Privkeys which generate identity
+  /// @param points Local hostname/ip address + port list
+  /// @param has_transport Supports NTCP, SSU
   /// @param caps RI capabilities
   RouterInfo(
       const core::PrivateKeys& keys,
-      const std::pair<std::string, std::uint16_t>& point,
+      const std::vector<std::pair<std::string, std::uint16_t>>& points,
       const std::pair<bool, bool>& has_transport,  // TODO(anonimal): refactor as bitwise SupportedTransport?
       const std::uint8_t caps = core::RouterInfo::Cap::Reachable);
 
--- a/src/util/routerinfo.cc
+++ b/src/util/routerinfo.cc
@@ -39,6 +39,7 @@
 #include "core/util/exception.h"
 #include "core/util/filesystem.h"
 #include "core/util/log.h"
+#include "core/util/config.h"
 #include "version.h"  // NOLINT(build/include)
 
 namespace bpo = boost::program_options;
@@ -53,7 +54,9 @@ RouterInfoCommand::RouterInfoCommand()
   bpo::options_description create_options("Create options");
   create_options.add_options()(
       "create,c", bpo::bool_switch()->default_value(false))(
-      "host", bpo::value<std::string>()->default_value("127.0.0.1"))(
+      "host",
+      bpo::value<core::Configuration::ListParameter<std::string, 2>>()->default_value(
+          core::Configuration::ListParameter<std::string, 2>("127.0.0.1")))(
       "port", bpo::value<int>()->default_value(0))(
       "floodfill,f",
       bpo::value<bool>()->default_value(false)->value_name("bool"))(
@@ -74,8 +77,10 @@ void RouterInfoCommand::PrintUsage(const std::string& name) const
 {
   LOG(info) << "Syntax: " << name << m_Options;
   LOG(info) << "Example: " << name << " routerInfo-(...).dat";
-  LOG(info) << "or: " << name << " --create --host 192.168.1.1 --port 10100 "
-                                 "--floodfill 1 --bandwidth P";
+  LOG(info) << "or: " << name
+            << " --create --host "
+               "192.168.1.1,2a01:e35:8b5c:b240:71a2:6750:8d4:47fa --port 10100 "
+               "--floodfill 1 --bandwidth P";
 }
 
 bool RouterInfoCommand::Impl(
@@ -116,7 +121,7 @@ bool RouterInfoCommand::Impl(
                             "for creation";
               return false;
             }
-          auto host = vm["host"].as<std::string>();
+          auto hosts = vm["host"].as<core::Configuration::ListParameter<std::string, 2>>();
           auto port = vm["port"].defaulted() ? core::RandInRange32(
                                                    core::RouterInfo::MinPort,
                                                    core::RouterInfo::MaxPort)
@@ -134,9 +139,12 @@ bool RouterInfoCommand::Impl(
           // Generate private key
           core::PrivateKeys keys = core::PrivateKeys::CreateRandomKeys(
               core::DEFAULT_ROUTER_SIGNING_KEY_TYPE);
-          // Create router info
+
+          // Read and fill the RI points
           std::vector<std::pair<std::string, std::uint16_t>> points;
-          points.emplace_back(host, port);
+          for (const auto& host : hosts.values)
+            points.emplace_back(host, port);
+          // Create router info
           core::RouterInfo routerInfo(
               keys, points, std::make_pair(has_ntcp, has_ssu), caps);
           // Set capabilities after creation to allow for disabling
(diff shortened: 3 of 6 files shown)


src/util/routerinfo.cc
```diff
@@ -118,16 +121,33 @@ bool RouterInfoCommand::Impl(
...
+
           std::uint8_t caps(core::RouterInfo::Cap::Reachable);
           // Generate private key
           core::PrivateKeys keys = core::PrivateKeys::CreateRandomKeys(
               core::DEFAULT_ROUTER_SIGNING_KEY_TYPE);
           // Create router info
-          core::RouterInfo routerInfo(
-              keys,
-              std::make_pair(host, port),
-              std::make_pair(true, true),
```
anonimal: This reminds me that there is a bug that I've been meaning to address. As found on testnet, when disabling transports, they are not appropriately disabled.

src/core/router/info.h
```diff
@@ -430,23 +431,33 @@ struct RouterInfoTraits
...
+  /// @brief  Interface Address object to be used by RouterInfo holders.
+  struct AddressInfo
+  {
+    std::string host;
+    std::uint16_t port;
+    // has_transport Supports NTCP, SSU
+    std::pair<bool, bool>
+        has_transport;  // TODO(anonimal): refactor as bitwise SupportedTransport?
+  };
+  using AddressInfoList = std::vector<AddressInfo>;
```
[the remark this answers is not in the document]
brbzull0: > When creating an RI, we should simply pass reference to an Address object or, if we're really picky about memory, something more akin to an AddressBase object (if redesign needs to go that far). An additional perk of using an Address object is that boost::asio::ip::address (as host) can guarantee a valid address syntax sooner than later

Totally agreed.

Now, to put clear expectations, do we want:
* Fully `--host` support + existing `Address` type as part of the RI interface.
* Fully  RI refactoring ( which should be traced through #627 )

being totally 2 different tasks right? I do not want to hold #777 because of #627.

I'll wait for your thoughts about this.
Thanks.

src/util/routerinfo.cc
```diff
@@ -118,16 +121,33 @@ bool RouterInfoCommand::Impl(
...
+          for (const auto& host_token : host_tokens)
+            {
+              core::RouterInfo::AddressInfo address;
+              address.host = host_token;
+              address.port = port;
+              address.has_transport = std::make_pair(true, true);
+              hosts.push_back(std::move(address));
+            }
+          // ipv4 and ipv6 allowed only.
+          if (hosts.size() > 2)
```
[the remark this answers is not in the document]
brbzull0: Grand, this is the conversation I was looking for. 
To help with some of the validations that should be done here we may need the `Address` change we mentioned above so we can get the transport/host validation quite early.

src/core/router/info.h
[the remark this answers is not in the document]
anonimal: >being totally 2 different tasks right?

Yes. Worst case scenario: we get this PR working to resolve #777, write in appropriate TODO's (TODO's regarding what we've discussed here and also point to this PR number), merge the PR, then continue working on #627.

>Now, to put clear expectations, do we want:
>Fully --host support + existing Address type as part of the RI interface.
>Fully RI refactoring ( which should be traced through #627 )

Yes, we want both but, for now, if *you* want: `Fully --host support + existing Address type as part of the RI interface.`. The `Address` class may not fit as perfectly we we'd like, so the outcome remains to be seen. If there is more stumbling, then more PR NACKs may pop-up but I'm open to collaboration if you are (I don't know your patience or interest level yet). If you *don't* want to spend too much time on this PR, then I'd say do the 'worst case scenario' option instead.

src/util/routerinfo.cc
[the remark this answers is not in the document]
anonimal: >To help with some of the validations that should be done here we may need the Address change we mentioned above so we can get the transport/host validation quite early.

We can probably do some nice checks during construction.

Another note: ensuring that we don't insert 2 ipv4's or 2 ipv6's into `RouterInfo` and/or multiple `Address`es.

Maivaza:
Testnet: unable to disable transports #832#

brbzull0:
It seems that there is also a dependency of the `--host` param in `RouterContext::Initialize` which uses the host for creating new transport address when they are enabled via `--enable-ssu/ntcp` (and not already present). Should this be also updated to deal with the option of having both ipv4 and ipv6 as incoming params?

anonimal:
>Should this be also updated to deal with the option of having both ipv4 and ipv6 as incoming params?

By the time we reach initialization, the host(s) should already be parsed. Then, assuming we'll be dealing with boost asio ip address objects, we could perhaps store the objects in a sequence container and then process the is_v4/6 (etc.) internally, within the RI ctor, instead of cluttering the interface. TBD, I'm curious to see how you'll proceed.

anonimal:
@brbzull0 I'm just now seeing your IRC message. Awesome, and what I mentioned above should still apply (if that's where the interface ends up going) because of type `T`.

I would suggest keeping that work with core util config file for now, but I really do appreciate your concern. What I'm foreseeing soon is a core util misc file for just this type of occasion since I wanted to do that for tag, timestamp, and queue. TBD.

src/core/router/info.h
```diff
@@ -492,14 +480,17 @@ class RouterInfo : public RouterInfoTraits, public RoutingDestination
...
-  /// @param key Our intoducer key
-  /// @param mtu Address MTU
-  void AddAddress(
-      const std::tuple<Transport, std::string, std::uint16_t>& point,
-      const std::uint8_t* key = nullptr,
-      const std::uint16_t mtu = 0);
+  /// @params addr rvalue (will be moved into our container).
+  /// @params add_key If true we will insert the key using the IdentityEx
+  ///         which was passed during construction.
+  void AddAddress(Address&& addr, const bool add_key = false);
```
brbzull0: This won't scale, I just noticed this.  If keep the interface, it should be a template function and take an universal reference to guarantee it will work either with rvalues or lvalues(and all the flavors)

brbzull0:
Hi @anonimal , base on the conversation we had on the IRC(#kovri-dev) _as a side effect_ I'm changing the validation for the host parameter. 
1)  Where would be the best place to put this checks(implementation)?, I've seen some comments in this regard here : https://github.com/monero-project/kovri/blob/8b435571eb374c96fd6a3b2070a3c42cb66dc6d9/src/core/util/config.cc#L188-L193 


Another question is in relation of the checks we want to perform on this parameter. From the top of my head I can think of:

a - If present, at least one address
b - Not more than 2 address
c - Should be a valid hosts (`boost::asio::address::from_string()` should do the job here)
d - If two hosts provided, check for duplication (`--host 192.168.1.1, 192.168.1.1`)
e - If two host provided, they should be different versions (`--host 192.168.1.1, 192.168.2.100` should fail) ( 2 x ipv4 or 2 x ipv6)

I read  [this](https://geti2p.net/spec/common-structures#netdb-routerinfo) and [this](https://geti2p.net/en/docs/how/network-database). 2) Is there any other checks you think should be done on this param?

Thanks.

anonimal:
>Where would be the best place to put this checks(implementation)?

Whatever you end up doing, you'll need to ensure that the changes also work with the config file.

For now, ignore the TODO. You can place any config code as protected members within new `class ConfigInternal` (or similar naming), and have class `Configuration` inherit `ConfigInternal`. class `Configuration` should then be marked final. Keep all code within the `core` namespace and remove the new namespace you added.

>Is there any other checks you think should be done on this param?

Ensure that addresses are not within a [private address space](https://en.wikipedia.org/wiki/Private_network) or localhost (this is a netdb option more than a bind option). Eventually, we'll want to apply --host for NTCP only / when SSU is disabled but that should go in another PR. Thanks.

brbzull0:
Hi @anonimal , some follow up here.
>Whatever you end up doing, you'll need to ensure that the changes also work with the config file.

`host` is marked as `deprecated` in the default kovri config file. I do can play with the rest of the flags in combination with a manual(command line) passed `--host` but not sure if this is what you are referring here?

Thanks.

anonimal:
>host is marked as deprecated in the default kovri config file. I do can play with the rest of the flags in combination with a manual(command line) passed --host but not sure if this is what you are referring here?

If one doesn't pass `--host` and instead passes the equivalent arguments within the config file, kovri needs to work.

>It make sense to me to have this in a separate task as this validation wasn't there before for a single host input, it also will make easy to review for you.

Sure, add the TODO.

>Just to be clear, there is no new config code in place
>ValidateHostParameter() does not need any instance as (so far) is a namespace function
>Are you referring to the config_internal namespace?

- Everything you're adding in `config_list_parameter.h` is config code.
- Everything in `config_internal` is config code (no matter how vague you want the templates to be, its only current and foreseeable use is within config code).
- This will *not* be used outside of config code.
- Config code is core lib functionality.
- Core lib functionality should *not* rely on client / app code or non-dependency namespaces.
- The core library should not rely on cli arguments but does require config file parsing. This is why we keep bpo within the core library to begin with.
- There is no need for the extra file and there is no need for the extra namespace.
- namespaces are primarily meant for avoiding naming collisions but do have a loose coupling with design. To create a namespace for this use-case only enforces spaghetti logic and is an OOP anti-pattern. I'm asking nicely: please, don't.

There are other issues that need to be fixed before merging. Apparently you've completely ignored our API discussion on IRC and have decided to *not* make the necessary changes? As for the rebase fixups: I meant fixup the appropriate commits, not squash/fixup *all* the major changes into one commit. It's a potpourri commit now with a mashup of things that *can* be merged and things that *cannot* be merged. The contributing guide clearly says to keep commits atomic. A perfect example is https://github.com/monero-project/kovri/pull/842. 

One note to keep in mind is that host validation will ultimately want to be applied during RI creation for when a core API is in place (for after router startup / after config runs its course) but this doesn't require diverting from our current path at the moment - at least not until all of the aforementioned issues are resolved. We don't want redundancy, but I also need to finish the unfinished RI work because this PR is not taking things in the right direction (as we discussed in IRC... but you haven't made those necessary changes).

Thanks for your patience and persistence.

brbzull0:
Hi @anonimal 

>Apparently you've completely ignored our API discussion on IRC and have decided to not make the necessary changes?

Not at all! 
PR updated.

Waiting for comments.

Thanks.

src/core/router/info.cc
```diff
@@ -664,14 +669,12 @@ void RouterInfo::DisableV6()
...
-          LOG(debug) << "RouterInfo: " << __func__ << ": removing address";
-          m_Addresses.erase(m_Addresses.begin() + i);
-        }
-    }
+  m_Addresses.erase(
+      std::remove_if(
+          std::begin(m_Addresses),
+          std::end(m_Addresses),
+          [](const auto& address) { return address.host.is_v6(); }),
+      std::end(m_Addresses));
```
anonimal: >std::begin
>std::end

`.begin()` `.end()`. I don't think the map is going anywhere any time soon.

brbzull0: Hi, will update the PR.

On the `.begin()/.end()`, what the reason behind the request? As per documentation at [cppreference](http://en.cppreference.com/w/cpp/iterator/begin) :

`Returns exactly c.begin(), which is typically an iterator to the beginning of the sequence represented by c. If C is a standard Container, this returns C::iterator when c is not const-qualified, and C::const_iterator otherwise.`
which:
```
template< class C > 
auto begin( C& c ) -> decltype(c.begin());
```
I know, it does add another layer of indirection, that's true but that's not bad at all, it will give us flexibility if needed.

in this case a `std::map` meets all the container requirements. 
I do know that this is not generic code but there is no harm imo of using the free function instead. `std::begin/std::end` is widely recommended around the community from guys like Herb Sutter, etc. It was even posted at the isocpp [blog](https://isocpp.org/blog/2014/10/stdbegin-stdend).

Please let me know what's the reason behind not using this free functions?

Thanks,
brbzull.

anonimal: Because, in this case, as you've even pointed out, there's no reason to use free functions.

>std::begin/std::end is widely recommended around the community

That's overreaching and no direct evidence for this claim. Herb Sutter != the entire community.

>It was even posted at the isocpp blog.

A blog post, with no elaborating argument, that only points to a stack overflow post which ultimately says "so I guess it comes down to preference".

KISS.

brbzull0: > ..., it will give us flexibility if needed.

That's the one I pointed out, or at least the one I tried ;)

>...Herb Sutter != the entire community.

Agreed, probably I should have extend my 'etc'  instead. 
Over many years I've heard many guys from the community including HS, but all this is my personal opinion base on many things, including recommendation from many well known guys, and not that well known.
I won't go and find all the talks were this was recommended at community places like cppcon, cppnow, etc.

Not knowing what's gonna happen with a piece of code like this one,  let us  have flexibility of writing 'generic' code like this.

tbh, I do not think this discussion(if any) worth the time, as it's quite a trivial things which can be implemented in different ways, I've just did the one in my opinion is the best.

BR

anonimal: >Not knowing what's gonna happen with a piece of code like this one

I know *exactly* what's going to happen with this piece of code; that's why I bothered commenting. Instead of simply making the change, you've chosen to argue and it's been a waste of time but that's ok because the [C4](https://github.com/monero-project/kovri-docs/blob/master/i18n/en/contributing.md#code-of-conduct-22c41) allows anyone to revert any changes.

Just make the [appropriate git-log references](https://github.com/monero-project/kovri-docs/blob/master/i18n/en/contributing.md#sending-your-work) and we can merge this PR.

brbzull0:
Updated Reference for each commit. Except the last one.

Thanks,
brbzull

PR #851, opened 2018-04-06
Get rid of VLA usage

btolfa:
---

---

We don't really need VLA usage. Moreover, we can use copy elision optimization by fixing methods signature.

### DIFF
--- a/src/core/util/filesystem.cc
+++ b/src/core/util/filesystem.cc
@@ -96,7 +96,7 @@ StringStream::ReadKeyPair()
 
   // TODO(anonimal): debug logging; include delimiter/terminator
 
-  return std::make_tuple(key, value, read_size);
+  return std::make_tuple(std::move(key), std::move(value), read_size);
 }
 
 std::string StringStream::ReadStringFromByte()
--- a/src/core/util/filesystem.cc
+++ b/src/core/util/filesystem.cc
@@ -71,7 +71,7 @@ StringStream::StringStream(
   m_Terminator = terminator;
 }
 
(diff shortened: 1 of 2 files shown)


src/core/util/filesystem.cc
```diff
@@ -96,21 +96,20 @@ StringStream::ReadKeyPair()
...
   // Get stated length amount
   std::uint8_t len;
   m_Stream.read(reinterpret_cast<char*>(&len), 1);
 
+  std::string string(static_cast<std::string::size_type>(len), '\0');
+
   // Read given amount
-  char buf[len];
-  m_Stream.read(reinterpret_cast<char*>(buf), len);
+  m_Stream.read(const_cast<char*>(string.data()), len);
```
anonimal: >len

`string.size()`

src/core/util/filesystem.cc
```diff
@@ -96,21 +96,20 @@ StringStream::ReadKeyPair()
...
 }
 
-const std::string StringStream::ReadStringFromByte()
+std::string StringStream::ReadStringFromByte()
 {
   // Get stated length amount
   std::uint8_t len;
   m_Stream.read(reinterpret_cast<char*>(&len), 1);
 
+  std::string string(static_cast<std::string::size_type>(len), '\0');
```
anonimal: >static_cast<<std::string::size_type>>()

This isn't necessary with gcc or clang. Is this another stupid MSVC caveat? If not, please don't give me a pedantic lecture about something that isn't necessary.

btolfa: You're right, this code isn't necessary.

Issue #848, opened 2018-04-03
AddressBook: new address book entry container


coneiric:
---
**By submitting this issue, I confirm the following:**

- I have read and understood the contributor guide in [kovri-docs](https://github.com/monero-project/kovri-docs).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.
---

Add a container class for individual address book entry (host-address) pairs.

Will create an interface for passing individual references or collections to/from `AddressBook`. New class will be similar to [AddressBookEntry](https://github.com/monero-project/kovri/blob/master/tests/unit_tests/client/address_book/impl.cc#L47), generalized to be useful in more contexts.

anonimal:
>similar to

If you're going to use `public` `private` class access modifiers and not treat the class as a POD type, then use as class and not a struct. https://github.com/monero-project/kovri/blob/master/tests/unit_tests/client/address_book/impl.cc#L49

coneiric:
> not treat the class as a POD type

I'm not opposed to treating it as a POD type. That would make porting easier, IIUC.
For simple classes/structs like this, is that a better general design?

anonimal:
>is that a better general design?

Basic C++. Lookup what POD means.

coneiric:
> what POD means

Right, I was more asking about if it was worth it to convert `AddressBook::Entry` to a POD type.

Short answer: no. 

Tried some unfruitful experiments, and it's just cleaner as a class. It just looked too heavy with the extra code to convert `kovri::core::IdentHash` + `std::string` to trivial types (and back).

anonimal:
Resolved in #852.

PR #852, opened 2018-04-09
AddressBook: Container for host-address entry

coneiric:
---

---

References #848

### DIFF
--- a/src/client/address_book/impl.cc
+++ b/src/client/address_book/impl.cc
@@ -529,6 +529,68 @@ void AddressBook::Stop() {
   m_Subscribers.clear();
 }
 
+BookEntry::BookEntry(
+    const std::string& host,
+    const kovri::core::IdentHash& address) try : m_Host(host),
+                                                 m_Address(address)
+  {
+    if (m_Host.empty())
+      throw std::invalid_argument("AddressBook: empty entry hostname");
+  }
+catch (...)
+  {
+    kovri::core::Exception ex(__func__);
+    ex.Dispatch();
+    throw;
+  }
+
+BookEntry::BookEntry(
+    const std::string& host,
+    const std::string& address) try : m_Host(host)
+  {
+    if (m_Host.empty())
+      throw std::invalid_argument("AddressBook: empty entry hostname");
+    core::IdentityEx ident;
+    ident.FromBase64(address);
+    m_Address = ident.GetIdentHash();
+  }
+catch (...)
+  {
+    kovri::core::Exception ex(__func__);
+    ex.Dispatch();
+    throw;
+  }
+
+BookEntry::BookEntry(const std::string& subscription_line) try
+  {
+    if (subscription_line.empty())
+      throw std::invalid_argument("AddressBook: empty subscription line");
+    std::size_t pos = subscription_line.find('=');
+    if (pos == std::string::npos)
+      throw std::runtime_error("AddressBook: invalid subscription line");
+    m_Host = subscription_line.substr(0, pos++);
+    if (m_Host.empty())
+      throw std::runtime_error("AddressBook: empty entry hostname");
+    core::IdentityEx ident;
+    ident.FromBase64(subscription_line.substr(pos));
+    m_Address = ident.GetIdentHash();
+  }
+catch (...)
+  {
+    kovri::core::Exception ex(__func__);
+    ex.Dispatch();
+    throw;
+  }
+
+const std::string& BookEntry::get_host() const noexcept
+{
+  return m_Host;
+}
+
+const core::IdentHash& BookEntry::get_address() const noexcept
+{
+  return m_Address;
+}
 /*
 // TODO(unassigned): currently unused
 void AddressBook::InsertAddress(
--- a/src/client/address_book/impl.h
+++ b/src/client/address_book/impl.h
@@ -59,6 +59,42 @@ namespace kovri {
 namespace client {
 
 class AddressBookSubscriber;
+/// @class BookEntry
+/// @brief Container for Address Book entry
+class BookEntry
+{
+ public:
+  /// @brief Create address book entry from a hostname & identity hash
+  /// @param host Kovri hostname
+  /// @param address Kovri identity hash
+  /// @throw std::invalid_argument on empty hostname
+  BookEntry(const std::string& host, const kovri::core::IdentHash& address);
+
+  /// @brief Create address book entry from a host & base64-encoded address
+  /// @param host Kovri hostname
+  /// @param address Base64-encoded Kovri address
+  /// @throw std::invalid_argument on empty hostname
+  /// @throw std::runtime_error on malformed or empty address
+  BookEntry(const std::string& host, const std::string& address);
+
+  /// @brief Create address book entry from a subscription line
+  /// @param subscription_line Line from a subscription file (host=address format)
+  /// @throw std::invalid_argument on empty subscription line
+  /// @throw std::runtime_error on empty hostname
+  /// @throw std::runtime_error on malformed subscription line
+  /// @throw std::runtime_error on malformed address
+  explicit BookEntry(const std::string& subscription_line);
+
+  /// @brief Accessor for host data member
+  const std::string& get_host() const noexcept;
+
+  /// @brief Accessor for address data member
+  const core::IdentHash& get_address() const noexcept;
+
+ private:
+  std::string m_Host;  ///< Human-readable Kovri hostname
+  core::IdentHash m_Address;  ///< Kovri identity hash
+};
 /// @class AddressBook
 /// @brief Address book implementation
 class AddressBook : public AddressBookDefaults {
--- a/tests/unit_tests/client/address_book/impl.cc
+++ b/tests/unit_tests/client/address_book/impl.cc
@@ -206,6 +206,54 @@ BOOST_AUTO_TEST_CASE(PGPClearSign) {
   BOOST_CHECK(Validate());
 }
 
+BOOST_AUTO_TEST_CASE(ValidBookEntry) {
+  // Test constructing from a subscription line
+  BOOST_CHECK_NO_THROW(kovri::client::BookEntry entry(subscription.front()));
+  kovri::client::BookEntry entry(subscription.front());
+
+  // Test constructing from an identity hash
+  BOOST_CHECK_NO_THROW(
+      kovri::client::BookEntry entry("kovri.i2p", entry.get_address()));
+
+  // Test constructing from a base64-encoded address
+  std::string const valid_dest =
+      subscription.front().substr(subscription.front().find('=') + 1);
+  BOOST_CHECK_NO_THROW(kovri::client::BookEntry entry("kovri.i2p", valid_dest));
+}
+
+BOOST_AUTO_TEST_CASE(InvalidBookEntry) {
+  std::string const valid_dest =
+      subscription.front().substr(subscription.front().find('=') + 1);
+  kovri::core::IdentityEx ident;
+  BOOST_CHECK_NO_THROW(ident.FromBase64(valid_dest));
+  ident.FromBase64(valid_dest);
+
+  // Empty Kovri hostname w/ base64-encoded address
+  BOOST_CHECK_THROW(
+      kovri::client::BookEntry entry("", valid_dest), std::invalid_argument);
+  // Empty Kovri hostname w/ identity hash address
+  BOOST_CHECK_THROW(
+      kovri::client::BookEntry entry("", ident.GetIdentHash()),
+      std::invalid_argument);
+  // Garbage Kovri address
+  BOOST_CHECK_THROW(
+      kovri::client::BookEntry entry("kovri.i2p", "QnNcMfHF"),
+      std::runtime_error);
+  // Empty Kovri address
+  BOOST_CHECK_THROW(
+      kovri::client::BookEntry entry("kovri.i2p", ""), std::runtime_error);
+  // Subscription line with no hostname
+  BOOST_CHECK_THROW(
+      kovri::client::BookEntry entry("=" + valid_dest), std::runtime_error);
+  // Subscription line with no address
+  BOOST_CHECK_THROW(
+      kovri::client::BookEntry entry("kovri.i2p="), std::runtime_error);
+  // Garbage subscription line
+  BOOST_CHECK_THROW(
+      kovri::client::BookEntry entry("QnNcMfHF"), std::runtime_error);
+  // Empty subscription line
+  BOOST_CHECK_THROW(kovri::client::BookEntry entry(""), std::invalid_argument);
+}
 // TODO(unassigned): more cases?
 
 BOOST_AUTO_TEST_CASE(RejectDuplicateEntry)
--- a/tests/unit_tests/client/address_book/impl.cc
+++ b/tests/unit_tests/client/address_book/impl.cc
@@ -44,45 +44,6 @@
 
 /// @class SubscriptionFixture
 struct SubscriptionFixture {
-  struct AddressBookEntry
-  {
-   public:
-    explicit AddressBookEntry(const std::string& subscription_line)
-    {
-      try
-        {
-          std::size_t pos = subscription_line.find('=');
-          if (pos == std::string::npos)
-            throw std::runtime_error(
-                "AddressBookEntry: invalid subscription line");
-          m_Host = subscription_line.substr(0, pos++);
-          kovri::core::IdentityEx ident;
-          ident.FromBase64(subscription_line.substr(pos));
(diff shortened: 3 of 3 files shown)


src/client/address_book/impl.h
```diff
@@ -63,6 +63,99 @@ class AddressBookSubscriber;
...
+    /// @brief No default copy-constructor
+    Entry(Entry& rhs) = delete;
+    /// @brief No default move-constructor
+    Entry(Entry&& rhs) = delete;
+    /// @brief Create address book entry from a hostname & identity hash
+    /// @param host Kovri hostname
+    /// @param address Kovri identity hash
+    explicit Entry(
+        const std::string&& host,
+        const kovri::core::IdentHash&& address)
```
btolfa: Please, read more about move semantics and other modern C++ feature.
I'd recommend [Effective Modern C++](https://www.amazon.com/Effective-Modern-Specific-Ways-Improve/dp/1491903996)
- Are you sure with [explicit](http://en.cppreference.com/w/cpp/language/explicit) keyword here?
- `const T&&` type doesn't make sense here
- Even so, you're misusing the benefits of move semantics

I would rewrite this ctor the following way:
```cpp
    Entry(
        std::string host,
        kovri::core::IdentHash address)
      : m_Host(std::move(host))
      , m_Address(std::move(address))
    {
      if (host.empty())
      {
         kovri::core::Exception ex;
         ex.Dispatch(__func__);
         throw std::runtime_error("AddressBook: empty entry hostname");
      }
    }
```

src/client/address_book/impl.h
```diff
@@ -63,6 +63,99 @@ class AddressBookSubscriber;
...
+        {
+          std::size_t pos = subscription_line.find('=');
+          if (pos == std::string::npos)
+            throw std::runtime_error("AddressBook: invalid subscription line");
+          m_Host = std::move(subscription_line.substr(0, pos++));
+          if (m_Host.empty())
+            throw std::runtime_error("AddressBook: empty entry hostname");
+          kovri::core::IdentityEx ident;
+          ident.FromBase64(subscription_line.substr(pos));
+          m_Address = std::move(ident.GetIdentHash());
```
btolfa:
```cpp
m_Address = ident.GetIdentHash();
```
IdentytyEx is trivially-copyable https://clang.llvm.org/extra/clang-tidy/checks/performance-move-const-arg.html

src/client/address_book/impl.h
brbzull0: Just a note here:
if `std::move` is used on the passed `host` value, then `host` should not be used afterwards.

```
if (host.empty()) { ... }
```

Will for sure be true as `host` will not longer have the value you want.

btolfa: Thanks, I fixed the snippet.

coneiric: Thanks for the catch & rewrite. I'm pretty new to move semantics, so this helps a lot.

Had trouble in one of the unit-test cases with `IdentHash` copying, but that may have been from something else.

Will update with the recommended changes. Thanks @btolfa & @brbzull0 :+1:

src/client/address_book/impl.h
```diff
@@ -63,6 +63,99 @@ class AddressBookSubscriber;
...
+          kovri::core::Exception ex;
+          ex.Dispatch(__func__);
+          throw;
+        }
+    }
+
+    /// @brief Create address book entry from a host & base64-encoded address
+    /// @param host Kovri hostname
+    /// @param address Base64-encoded Kovri address
+    explicit Entry(const std::string&& host, const std::string& address)
```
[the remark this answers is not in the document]
coneiric: Will do, this is much cleaner.

src/client/address_book/impl.h
```diff
@@ -63,6 +63,99 @@ class AddressBookSubscriber;
...
+        : m_Host(host)
+    {
+      try
+        {
+          if (m_Host.empty())
+            throw std::runtime_error("AddressBook: empty entry hostname");
+          kovri::core::IdentityEx ident;
+          ident.FromBase64(address);
+          m_Address = std::move(ident.GetIdentHash());
```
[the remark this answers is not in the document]
coneiric: I wasn't sure about this when writing this class, will update with your recommendation.

Thanks for the link as well :)

coneiric:
Absolutely agree. I have a copy of the book, and will revisit the sections on move semantics.

I'm not sure about the explicit keyword usage. Thought it might help the compiler make the right decision on ctor choice, but now it seems I was mistaken. 

IIUC now, explicit would be useful for differentiating between overloaded ctors where the parameter types of one ctor could be implicitly converted to the parameter types of the other ctor?

Thank you much for the review! Will make the changes, and update the PR.

coneiric:
For some reason, coveralls is reporting a drop in coverage on `src/client/address_book/impl.h` (even though I have 100% coverage on added code).

Not sure if this is from testing a file that coveralls wasn't previously including in the statistics, or if something else is causing the coverage drop. Will investigate.

src/client/address_book/impl.h
```diff
@@ -63,6 +63,89 @@ class AddressBookSubscriber;
...
+          kovri::core::Exception ex;
+          ex.Dispatch(__func__);
+          throw;
+        }
+    }
+
+    /// @brief Create address book entry from a subscription line
+    /// @param subscription_line Line from a subscription file (host=address format)
+    /// @notes Throws on malformed subscription lines
+    Entry(const std::string& subscription_line)
```
anonimal: Single-parameter constructors should be marked explicit.

src/client/address_book/impl.h
```diff
@@ -63,6 +63,89 @@ class AddressBookSubscriber;
...
+    Entry(Entry& rhs) = delete;
+
+    /// @brief No default move-constructor
+    Entry(Entry&& rhs) = delete;
+
+    /// @brief Create address book entry from a hostname & identity hash
+    /// @param host Kovri hostname
+    /// @param address Kovri identity hash
+    Entry(std::string host, const kovri::core::IdentHash& address)
+        : m_Host(std::move(host)), m_Address(address)
```
anonimal: >Entry(std::string host,
>m_Host(std::move(host))

I see absolutely no reason to not pass const reference here. Why do you want move semantics here.

src/client/address_book/impl.h
```diff
@@ -63,6 +63,89 @@ class AddressBookSubscriber;
...
+    {
+      if (m_Host.empty())
+        throw std::runtime_error("AddressBook: empty entry hostname");
+    }
+
+    /// @brief Create address book entry from a host & base64-encoded address
+    /// @param host Kovri hostname
+    /// @param address Base64-encoded Kovri address
+    Entry(std::string host, const std::string& address)
+        : m_Host(std::move(host))
```
anonimal: >Entry(std::string host,
>m_Host(std::move(host))

I see absolutely no reason to not pass const reference here. Why do you want move semantics here.

anonimal:
>While this PR is in a failing state

It's not in a failing state, it builds fine.

src/client/address_book/impl.h
btolfa: > Item 41: Consider pass by value for copyable parameters that are cheap to move and always copied.

from [Effective Modern C++ by Scott Meyers](https://www.amazon.com/dp/1491903996)
https://www.safaribooksonline.com/library/view/effective-modern-c/9781491908419/ch08.html

passed by const reference
```cpp
class Entry {
public:
  Entry(std::string &host) : m_host(host) {}
private:
  std::string m_host;
};
```

will give the following result:

```cpp
Entry e1("hello.world");
// Create temporary std::string object - first allocation 
// Copy from temporary std::string to m_host - second allocation

auto lambda = []() -> std::string { return "hello.world" };
Entry e2(lambda());
// temporary std::string returned from lambda - first allocation somewhere inside lambda 
// copy from temporary std::string to m_host - second allocation

std::string v = "hello.world";
const std::string& cf = v;
Entry e3(cf);
// only one copy
```

where as, passed by value 

```cpp
class Entry {
public:
  Entry(std::string host) : m_host(std::move(host)) {}
private:
  std::string m_host;
};
```

will give the following result:

```cpp
Entry e1("hello.world");
// Create host argument from string literal - only allocation 
// move from host to m_host

auto lambda = []() -> std::string { return "hello.world" };
Entry e2(lambda());
// temporary std::string returned from lambda - allocation somewhere inside lambda
// move from temporary rvalue to host argument 
// move from host to m_host

std::string v = "hello.world";
const std::string& cf = v;
Entry e3(cf);
// copy from cf to host argument - allocation
// move from host to m_host
```

src/client/address_book/impl.h
```diff
@@ -63,6 +63,89 @@ class AddressBookSubscriber;
...
+    Entry(Entry&& rhs) = delete;
+
+    /// @brief Create address book entry from a hostname & identity hash
+    /// @param host Kovri hostname
+    /// @param address Kovri identity hash
+    Entry(std::string host, const kovri::core::IdentHash& address)
+        : m_Host(std::move(host)), m_Address(address)
+    {
+      if (m_Host.empty())
+        throw std::runtime_error("AddressBook: empty entry hostname");
```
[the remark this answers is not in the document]
coneiric: Will do.

tests/unit_tests/client/address_book/impl.cc
```diff
@@ -206,15 +167,65 @@ BOOST_AUTO_TEST_CASE(PGPClearSign) {
...
+      "kovri.i2p", subscription.front().substr(pos)));
+}
+
+BOOST_AUTO_TEST_CASE(InvalidBookEntry) {
+  std::size_t const pos = subscription.front().find('=') + 1;
+  std::string const valid_dest = subscription.front().substr(pos);
+  // Empty Kovri hostname, string-parameter move-constructor
+  BOOST_CHECK_THROW(
+      kovri::client::AddressBook::Entry entry("", valid_dest),
+      std::runtime_error);
```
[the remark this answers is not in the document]
coneiric: Will do.

tests/unit_tests/client/address_book/impl.cc
```diff
@@ -206,15 +167,65 @@ BOOST_AUTO_TEST_CASE(PGPClearSign) {
...
+  // Empty Kovri hostname, overloaded IdentHash-parameter move-constructor
+  kovri::core::IdentityEx ident;
+  ident.FromBase64(valid_dest);
+  BOOST_CHECK_THROW(
+      kovri::client::AddressBook::Entry entry("", ident.GetIdentHash()),
+      std::runtime_error);
```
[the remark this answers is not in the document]
coneiric: Will do.

src/client/address_book/impl.h
```diff
@@ -63,6 +63,89 @@ class AddressBookSubscriber;
 /// @brief Address book implementation
 class AddressBook : public AddressBookDefaults {
  public:
+  /// @brief Container for Address Book entry
+  class Entry
```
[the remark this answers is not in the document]
coneiric: I thought it made sense from a namespace perspective, and as a `AddressBook "has-a" `Entry`. Maybe it makes more sense un-nested as a `BookEntry` class?

src/client/address_book/impl.h
coneiric: If I keep this as a constructor, I'll mark it as explicit.

src/client/address_book/impl.h
coneiric: > Why do you want move semantics here.

It was an attempt at writing an optimized container, along the lines of @btolfa's examples above.

After reading more about move semantics, I was mistaken about how/when to use it.

In the contexts where I see `Entry` being beneficial, one would construct the entry using temporary variables. It seems too expensive to deal with the mental overhead of remembering to construct an `Entry` only on the last use of the temporary variables.

I'll look again at the contexts where `Entry` will be useful, but I think I'll remove explicit use of move semantics.

src/client/address_book/impl.h
coneiric: Same line of reasoning as the comment above.

coneiric:
> If you keep the moves (don't, unless you can prove a need), then include <utility>. Also remember to cpplint.

After reading more about move semantics + working more on the code, I think my uses here may be premature optimizations. Will make sure to run cpplint before submitting the next set of patches (& future PRs).

src/client/address_book/impl.h
anonimal: >passed by const reference
>Entry(std::string &host)

@btolfa **This is not const reference. The example you provided is not even accurate to our case nor is it needed for our case**.

@coneiric **KISS**

src/client/address_book/impl.h
```diff
@@ -59,6 +59,42 @@ namespace kovri {
...
+  /// @throw std::invalid_argument on empty hostname
+  /// @throw std::invalid_argument on malformed address
+  BookEntry(const std::string& host, const std::string& address);
+
+  /// @brief Create address book entry from a subscription line
+  /// @param subscription_line Line from a subscription file (host=address format)
+  /// @throw std::invalid_argument on empty hostname
+  /// @throw std::invalid_argument on empty subscription line
+  /// @throw std::invalid_argument on malformed address
+  /// @throw std::invalid_argument on malformed subscription line
```
anonimal: Malformed, in this case, is not considered std::invalid_argument because the argument is valid as it contains data - but the data itself isn't valid.

src/client/address_book/impl.h
```diff
@@ -59,6 +59,42 @@ namespace kovri {
...
+  /// @param host Kovri hostname
+  /// @param address Kovri identity hash
+  /// @throw std::invalid_argument on empty hostname
+  BookEntry(const std::string& host, const kovri::core::IdentHash& address);
+
+  /// @brief Create address book entry from a host & base64-encoded address
+  /// @param host Kovri hostname
+  /// @param address Base64-encoded Kovri address
+  /// @throw std::invalid_argument on empty hostname
+  /// @throw std::invalid_argument on malformed address
```
anonimal: Malformed, in this case, is not considered std::invalid_argument because the argument is valid as it contains data - but the data itself isn't valid.

coneiric:
> Malformed, in this case, is not considered std::invalid_argument because the argument is valid as it contains data

Updated with your latest recommended changes. Thanks for the review and your notes.

src/client/address_book/impl.cc
```diff
@@ -528,6 +528,68 @@ void AddressBook::Stop() {
...
+    m_Address = ident.GetIdentHash();
+  }
+catch (...)
+  {
+    kovri::core::Exception ex(__func__);
+    ex.Dispatch();
+    throw;
+  }
+
+const std::string& BookEntry::GetHost() const
```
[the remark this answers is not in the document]
coneiric: > Can (should) be noexcept

I thought this as well, but wasn't sure.

> accessors/mutators in lowercase/underscore

+1 this. Will make the changes.

src/client/address_book/impl.cc
```diff
@@ -528,6 +528,68 @@ void AddressBook::Stop() {
...
+{
+  return m_Host;
+}
+
+const core::IdentHash& BookEntry::GetAddress() const
```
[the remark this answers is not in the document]
coneiric: Agreed. See comment above.

src/client/address_book/impl.cc
```diff
@@ -528,6 +528,68 @@ void AddressBook::Stop() {
...
+    ex.Dispatch();
+    throw;
+  }
+
+BookEntry::BookEntry(
+    const std::string& host,
+    const std::string& address) try : m_Host(host)
+  {
+    if (m_Host.empty())
+      throw std::invalid_argument("AddressBook: empty entry hostname");
```
coneiric: Would you NACK adding a similar check for an empty address?
For example:

```
if (address.empty())
  throw std::invalid_argument("AddressBook: empty entry address");
```

IIUC now, this check is more efficient than the call to `ident.FromBase64`. Please correct me if I'm wrong.

src/client/address_book/impl.cc
```diff
@@ -528,6 +528,68 @@ void AddressBook::Stop() {
...
+    throw;
+  }
+
+BookEntry::BookEntry(const std::string& subscription_line) try
+  {
+    if (subscription_line.empty())
+      throw std::invalid_argument("AddressBook: empty subscription line");
+    std::size_t pos = subscription_line.find('=');
+    if (pos == std::string::npos)
+      throw std::invalid_argument("AddressBook: invalid subscription line");
```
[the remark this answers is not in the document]
coneiric: Will do.

src/client/address_book/impl.cc
anonimal: That depends on the case. I wouldn't concern myself with that right now for the remainder of this PR.

coneiric: Alright, will leave it out for now.

PR #867, opened 2018-04-25
Build: Use cotire for all supported platforms

btolfa:
---

---
https://github.com/sakra/cotire

Right now cotire is enabled only for MSVC builds. This PR enables it for other platforms.
On the local machine, it reduces the build time by 2 times.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,9 +59,10 @@ option(WITH_TESTS      "Build unit tests" OFF)
 option(WITH_FUZZ_TESTS "Build fuzz tests" OFF)
 option(WITH_UPNP       "Include support for UPnP client" OFF)
 option(WITH_COTIRE     "Enable cotire (compile time reducer) - precompiled header and single compilation unit builds" ${MSVC})
+# Verbosity enabled by default, but can be disabled from cli by adding -DCMAKE_VERBOSE_MAKEFILE=OFF
+# https://github.com/monero-project/kovri/pull/867#discussion_r184359361
+option(CMAKE_VERBOSE_MAKEFILE "Enable verbose build output" ON)
 
-
-set(CMAKE_VERBOSE_MAKEFILE true)
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake ${CMAKE_CURRENT_SOURCE_DIR}/cmake/cotire/CMake)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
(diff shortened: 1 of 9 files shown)


btolfa:
For comparison:
- build without cotire - [11 mins, 12 secs](https://build.getmonero.org/builders/kovri-all-freebsd64/builds/915)
- cotired build - [3 mins, 55 secs](https://build.getmonero.org/builders/kovri-all-freebsd64/builds/919)

The disadvantage is that cotired builds are more RAM-intensive.
[buildbot/kovri-all-dragonflybsd-amd64](https://build.getmonero.org/builders/kovri-all-dragonflybsd-amd64/builds/589) failed with internal compiler error:
```
CC: internal compiler error: Killed (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
See <http://bugs.dragonflybsd.org> for instructions.
```
However, it actually can be an internal compiler error and not a memory issue.

anonimal:
>However, it actually can be an internal compiler error and not a memory issue.

I doubt it, but if we can increase memory/swap on the build machines then great.

CMakeLists.txt
```diff
@@ -58,9 +58,9 @@ option(WITH_SUPERCOP   "Build Ed25519 using the ref10 implementation from SUPERC
 option(WITH_TESTS      "Build unit tests" OFF)
 option(WITH_FUZZ_TESTS "Build fuzz tests" OFF)
 option(WITH_UPNP       "Include support for UPnP client" OFF)
+option(WITH_COTIRE     "Enable cotire (compile time reducer) - precompiled header and single compilation unit builds" ${MSVC})
 
 
-set(CMAKE_VERBOSE_MAKEFILE true)
```
anonimal: No. This is useful for debugging on all our buildbot machines - and I literally just put this in this past week. I'm not going to ask Dan to maintain `VERBOSE=1`.

btolfa: From IRC
```
<btolfa> anonimal about https://github.com/monero-project/kovri/pull/867 - I don't like by default set(CMAKE_VERBOSE_MAKEFILE true), how about detecting buildbot environment from Makefile and use VERBOSE=1 only for buildbot's builds?
<anonimal> 1) that sounds like unnecessary work 2) we won't need to say "VERBOSE=1" every time someone submits a build log to debug
<anonimal> What's the big deal? It's more helpful than not.
<btolfa> It makes output a bit noisy. Ok, I will make it enabled by default, but overridable from cmake options - like option(CMAKE_VERBOSE_MAKEFILE "" ON) instead of set(CMAKE_VERBOSE_MAKEFILE TRUE)
<anonimal> Ok, an option() is fine. Please reference the PR # so we know why.
```

anonimal:
>I'd like to see if monero-project/meta#213 can be resolved before we merge.

Dan will get to this when he can. Merging.

PR #872, opened 2018-05-20
HTTP: don't catch / rethrow std::exception_ptr

coneiric:
---

---

[cpp-netlib/819](https://github.com/cpp-netlib/cpp-netlib/pull/819) fixes underlying issue that required the workaround in #491.

__notes__: a style issue remains (extra indentation level), but changing the indentation obscures the real changes for this PR. If desired, I can create a separate PR for the style changes.

### DIFF
--- a/src/client/util/http.cc
+++ b/src/client/util/http.cc
@@ -148,9 +148,6 @@ bool HTTP::DownloadViaClearnet() {
 
   // Create client with options
   Client client(options);
-  // TODO(unassigned): this top try block is specifically for Windows and ARMv8
-  // but is harmless for all platforms (see #453 and cpp-netlib/cpp-netlib#696)
-  try {
     try {
       // Create request
       Request request(uri.string());  // A fully-qualified, completed URI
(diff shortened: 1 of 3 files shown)


coneiric:
> 1. + 2.

I wasn't sure how to label the issue, and wasn't able to build/test on windows.

> 3.

There is a test in cpp-netlib that I more-or-less copy-pasta-ed. Thought this might be overkill, sure of it now.

> 4. Using a reseed server that we don't control is an unreliable way to do testing

Felt gross when I wrote the test. Short of setting up a reseed server, wasn't sure how else to test.

> 5. Network tests within unit-tests will leak the fact that someone is building/testing kovri without actually running kovri.

This is the best reason to close this PR. Along with dropping cpp-netlib support, I can't come up with a good reason to leave this open.

anonimal:
Thanks for the effort though, @coneiric. We should consider building up more network tests. I should've mentioned our python hooks with boost.python. We should consider building up tests with those / build up our python integration.

PR #877, opened 2018-05-28
kovri-install.sh: moved catch(). Fixes #876

### LINKED ISSUE #876: kovri-install.sh: line 117: catch: command not found
---

---

> make -j8 DESTDIR=/var/tmp/portage/net-vpn/kovri-9999/image/ install
./pkg/installers/**kovri-install.sh: line 117: catch: command not found**
The Kovri I2P Router Project (c) 2015-2017

The catch() function in pkg/installers/kovri-install.sh is defined after it's first used outside a function. This leads to the above error.

Wuodan:
---

---

### DIFF
--- a/src/app/daemon_linux.cc
+++ b/src/app/daemon_linux.cc
@@ -146,8 +146,11 @@ bool DaemonLinux::Initialize() {
     return false;
   }
   LOG(debug) << "DaemonLinux: pid file ready";
+
   // Signal handler
-  struct sigaction sa {{}};
+  struct sigaction sa;
+  memset(&sa, 0, sizeof(struct sigaction));  // C struct initialized in C way.
+
   sa.sa_handler = handle_signal;
   sigemptyset(&sa.sa_mask);
   sa.sa_flags = SA_RESTART;
--- a/src/client/api/streaming.cc
+++ b/src/client/api/streaming.cc
@@ -105,7 +105,7 @@ Stream::Stream(
       m_Status(eStreamStatusNew),
       m_IsAckSendScheduled(false),
       m_LocalDestination(local),
-      m_CurrentRemoteLease {{}},
+      m_CurrentRemoteLease {0, 0, 0},
       m_ReceiveTimer(m_Service),
       m_ResendTimer(m_Service),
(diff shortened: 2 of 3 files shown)


pkg/installers/kovri-install.sh
```diff
@@ -107,7 +117,7 @@ shift "$(($OPTIND -1))"
 
 # Set default resources if needed
 if [[ -z $resources ]]; then
-  resources="pkg/client pkg/config contrib/utils/kovri-bash.sh build/${bins[0]} build/${bins[1]}"
+  resources="pkg/client pkg/config contrib/utils/kovri-bash.sh build/${bins[0]}"
```
anonimal: This isn't optional though as it's part of our release build.

anonimal: Could you instead apply a workaround for non-release installs?

Wuodan: Sorry, I didn't mean for that second commit to be in this PR, only the mini issue with the catch() function.

Will propose a patch for the Makefile for non-release installs to build kovri-util instead.

I'm using Gentoo Linux amd64 and arm, want to package kovri.
Installing in Gentoo by default means to build from source, so the package-management is not much more then compile and install scripts in a chroot. Shell is bash.

Wuodan: About kovri-util being optional or not:
In the Makefile, it's not used unless for release-builds.
The proposed change to the installer still installs kovri-util, it just does not check for its presence as required resource (which triggers the catch() function).

Should I create another PR where the Makefile builds kovri-util for non-release installs? Thus making kovri-util no longer 'optional'?

anonimal:
You didn't need to close this PR. You could've kept adding to it instead of opening #880.

anonimal:
JFTR: the catch issue is not present in bash 4.3.48 (Ubuntu) but is present with 4.4.19 (Arch). I imagine gentoo is using the 4.4 series.

Wuodan:
bash, version 4.4.12, correct. Sorry for closing the PR, it automatically with rebase, force push and new branch.

PR #879, opened 2018-05-29
Tests: Configuration::ListParameter unit-test

coneiric:
---

---

### DIFF
--- a/tests/unit_tests/CMakeLists.txt
+++ b/tests/unit_tests/CMakeLists.txt
@@ -20,6 +20,7 @@ add_executable(kovri-tests
   core/router/identity.cc
   core/router/transports/ssu/packet.cc
   core/util/byte_stream.cc
+  core/util/config.cc
   main.cc
   core/router/identity.h)
 
--- a/tests/unit_tests/core/util/config.cc
+++ b/tests/unit_tests/core/util/config.cc
@@ -1,0 +1,103 @@
+/**                                                                                           //
+ * Copyright (c) 2015-2018, The Kovri I2P Router Project                                      //
+ *                                                                                            //
+ * All rights reserved.                                                                       //
+ *                                                                                            //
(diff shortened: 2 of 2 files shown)


coneiric:
Ok, I'll add some test cases for substitution failure (maybe turn SFINAE to SFIAE :smile: jk), and think about some other failure cases.

anonimal:
>Ok, I'll add some test cases for substitution failure (maybe turn SFINAE to SFIAE smile jk)

Proper implementation of SFINAE requires adequate testing of said implementation. Assuming that an implementation fully implements SFINAE within the context of said implementation, because the standard library says "do this for SFINAE", does not equate to a proper implementation of SFINAE.

More importantly, because we have no policy for SFINAE, "expected exceptions and/or expected failures" should be tested, regardless, so as to allow malleable implementations. This is the core concept of a unit-test: to cover any and all pertinent aspects of a unit.

coneiric:
> More importantly, because we have no policy for SFINAE, "expected exceptions and/or expected failures" should be tested

I'll do my best to test the expected exception/failure cases. Currently having some trouble getting the tests to work properly, but that should be resolved fairly soon. Apparently, I forgot to include the new test in the unit-test CMake file, and these tests were never ran. I've included it now, and am working on fixing the errors in my tests.

coneiric:
>  working on fixing the errors in my tests

Apparently Boost macros hate templates with multiple arguments (the commas...). This gotcha set me back a few hours, but I used type aliases as a work-around. I tried `BOOST_PP_COMMA()` with no luck. If you have a better solution than my (somewhat) gross hack, I'm open to suggestions.

coneiric:
> After you fix the i686 build (remember, boost 1.58), let's consider this PR resolved for now.

Fixing it now, didn't realize `BOOST_TEST` wasn't defined in 1.58.

PR #881, opened 2018-05-30
Build: fix mingw warnings

coneiric:
---

---

### DIFF
--- a/src/app/win32_service.cc
+++ b/src/app/win32_service.cc
@@ -40,6 +40,7 @@
 
 #include "app/daemon.h"
 
+#include "core/util/config.h"
 #include "core/util/log.h"
 
 I2PService *I2PService::m_Service = NULL;
@@ -151,8 +152,11 @@ void I2PService::Start(
 
 void I2PService::OnStart(
     DWORD argc,
-    PSTR *argv) {
-  // TODO(unassigned): unused args
+    PSTR * argv) {
+  std::vector<std::string> args{};
+  for (DWORD i=0; i < argc; ++i)
+    args.push_back(argv[i]);
+  Daemon.Configure(args);
   LOG(info)
     << "I2PServiceWin32: Service in " << __func__ << EVENTLOG_INFORMATION_TYPE;
   Daemon.Start();
--- a/src/app/win32_service.cc
+++ b/src/app/win32_service.cc
@@ -1,5 +1,5 @@
 /**                                                                                           //
- * Copyright (c) 2013-2017, The Kovri I2P Router Project                                      //
+ * Copyright (c) 2013-2018, The Kovri I2P Router Project                                      //
  *                                                                                            //
  * All rights reserved.                                                                       //
  *                                                                                            //
@@ -48,7 +48,7 @@ BOOL I2PService::IsService() {
   BOOL is_service = FALSE;
   HWINSTA win_station = GetProcessWindowStation();
   if (win_station != NULL) {
-    USEROBJECTFLAGS uof = { 0 };
+    USEROBJECTFLAGS uof{};
     if (GetUserObjectInformation(win_station, UOI_FLAGS, &uof,
       sizeof(USEROBJECTFLAGS), NULL) && ((uof.dwFlags & WSF_VISIBLE) == 0)) {
         is_service = TRUE;
--- a/src/app/win32_service.cc
+++ b/src/app/win32_service.cc
@@ -61,8 +61,10 @@ BOOL I2PService::Run(
     I2PService &service) {
   m_Service = &service;
   SERVICE_TABLE_ENTRY service_table[] = {
-    { service.m_Name, ServiceMain },
-    { NULL, NULL }
+  // If service name is NULL, set to empty string to comply
+  //   with SERVICE_TABLE_ENTRY spec.
+    { service.m_Name == NULL ? PSTR("") : service.m_Name, ServiceMain },
+    { PSTR(""), NULL }
   };
   return StartServiceCtrlDispatcher(service_table);
 }
@@ -96,11 +98,9 @@ I2PService::I2PService(
     BOOL can_stop,
     BOOL can_shutdown,
     BOOL can_pause_continue) {
-  if (service_name == NULL) {
-    m_Name = "";  // TODO(unassigned): why?
-  } else {
-    m_Name = service_name;
-  }
+  // If service name is NULL, set to empty string to comply
+  //   with SERVICE_TABLE_ENTRY spec.
+  m_Name = service_name == NULL ? PSTR("") : service_name;
   m_StatusHandle = NULL;
   m_Status.dwServiceType = SERVICE_WIN32_OWN_PROCESS;
   m_Status.dwCurrentState = SERVICE_START_PENDING;
--- a/src/core/util/tag.h
+++ b/src/core/util/tag.h
@@ -58,6 +58,9 @@ class Tag
 
   Tag(const std::uint8_t* buf)
   {
+    assert(buf);
+    if (!buf)
+      throw std::invalid_argument("Null buffer not allowed");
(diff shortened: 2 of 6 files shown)


src/core/util/tag.h
```diff
@@ -56,8 +56,10 @@ class Tag
 
   Tag() : m_Buf{} {}
 
-  Tag(const std::uint8_t* buf)
+  Tag(const std::uint8_t* buf) : m_Buf{}
   {
+    if (!buf)
```
anonimal: Might as well add an assertion before this check.

src/core/util/tag.h
```diff
@@ -56,8 +56,10 @@ class Tag
 
   Tag() : m_Buf{} {}
 
-  Tag(const std::uint8_t* buf)
+  Tag(const std::uint8_t* buf) : m_Buf{}
```
anonimal: >Tag copy-constructor needs non-null buffers for well-defined memcpy.

...but this isn't a copy-constructor. What compiler are you using?

> : m_Buf{}

I intentionally avoided this unnecessary initialization in https://github.com/monero-project/kovri/commit/507b888f337161d52c4e28626180fd2deac3cc38 because, with memcpy, so long as the source is not null, you'll be copying over god-knows-what anyway; and the size of the buffer is templated so there's no uncertainty of the length copied.

anonimal:
On a sidenote: there is an app I2PService class and client I2PService class. In your git log, it may be helpful to clarify that fact in the commit titles.

src/app/win32_service.cc
```diff
@@ -96,11 +96,7 @@ I2PService::I2PService(
     BOOL can_stop,
     BOOL can_shutdown,
     BOOL can_pause_continue) {
-  if (service_name == NULL) {
-    m_Name = "";  // TODO(unassigned): why?
-  } else {
-    m_Name = service_name;
-  }
+  m_Name = service_name;
```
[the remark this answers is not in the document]
coneiric: > you could be setting a null pointer or deference a pointer to uninitialized memory

You're right, and I was incorrectly assuming that Windows would treat `NULL` the same as an empty string. After looking at the `SERVICE_TABLE_ENTRY` documentation, it is clear that service name cannot be `NULL`. Will correct the change, and add similar checks in `I2PService::Run`.

src/app/win32_service.cc
```diff
@@ -152,7 +148,8 @@ void I2PService::Start(
 void I2PService::OnStart(
     DWORD argc,
     PSTR *argv) {
-  // TODO(unassigned): unused args
+  if (argc || argv)
+    LOG(error) << "I2PService: Unknown startup arguments.";
```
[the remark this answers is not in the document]
coneiric: Another assumption that the Windows service did not use startup arguments. Instead of silently ignoring the non-use, I decided to log the error, hoping this would provide a hint to implement startup argument use.

I'll investigate the proper use, or non-use, of Windows service startup arguments.

src/core/util/tag.h
coneiric: > this isn't a copy-constructor

You're right, I was using the wrong term.

> I intentionally avoided this unnecessary initialization in 507b888 because, with memcpy, so long as the source is not null, you'll be copying over god-knows-what anyway; and the size of the buffer is templated so there's no uncertainty of the length copied.

I understand part of your reasoning, length is guaranteed safe. However, isn't writing over uninitialized memory undefined behavior (and potentially dangerous)? Sincerely asking, since I might be misunderstanding something.

src/core/util/tag.h
coneiric: Should the assertion replace the check, or should I use both?

coneiric:
> On a sidenote: there is an app I2PService class and client I2PService class. In your git log, it may be helpful to clarify that fact in the commit titles.

Absolutely, adding the clarification to the rebased commits.

src/app/win32_service.cc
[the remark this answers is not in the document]
anonimal: Thank you.

src/core/util/tag.h
anonimal: Both because we won't always define NDEBUG.

src/core/util/tag.h
anonimal: I think you're confusing idioms, and confusing allocation with initialization. The destination memory is allocated before the copy. As for undefined behavior, it is not undefined because the destination is not a nullptr and it is also TriviallyCopyable.

So, to clarify: upon construction, zero-initializing allocated memory with `Size` only to subsequently fill it with an unknown source of `Size`, would be considered an "unnecessary initialization". Most importantly, since we're not pulling anything from the buffer before the copy, there shouldn't be concern about pulling from uninitialized memory (which would definitely be a cause for concern). 

Does that make sense?

coneiric: Makes complete sense now, and I appreciate you explaining it. I didn't understand the part about the destination buffer being allocated before the copy. Was mistakenly thinking that the buffer would be pointing to some arbitrary location on the heap that could be occupied by something else. Now, I understand that it has its own heap location, and that location is filled with whatever garbage.

src/app/win32_service.cc
```diff
@@ -151,8 +152,11 @@ void I2PService::Start(
 
 void I2PService::OnStart(
     DWORD argc,
-    PSTR *argv) {
-  // TODO(unassigned): unused args
+    PSTR * argv) {
+  std::vector<std::string> args{};
+  for (DWORD i=0; i < argc; ++i)
+    args.push_back(argv[i]);
+  kovri::core::Configuration config(args);
```
anonimal: But now you've replaced unused args with an unused local :smile:

anonimal:
On an unrelated note, @coneiric could you please check your email? I can't get in touch with you otherwise (I don't see you on IRC).

src/app/win32_service.cc
coneiric: lol, derp. Thanks for the catch, I'll add the correct usage. Guess I rushed a bit :smile:

anonimal:
Deleted both of our comments because we both want privacy :sweat_smile:

src/core/util/tag.h
```diff
@@ -58,7 +58,13 @@ class Tag
 
   Tag(const std::uint8_t* buf)
   {
-    std::memcpy(m_Buf, buf, Size);
+    if (!buf)
+      {  // Zero-initialize Tag if null-buffer supplied
+        std::array<std::uint8_t, Size> zero_buf{};
+        std::memcpy(m_Buf, zero_buf.data(), zero_buf.size());
+      }
+    else
```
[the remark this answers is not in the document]
coneiric: > This block and condition make no sense. We should never have a nullptr (but if we did then we should throw here). Why did you remove the assertion and std::exception?

I was mistakenly following Google's style guide, but it is wrong when it comes to exceptions for Kovri. I'm reading Bjarne's papers on exception safety, and will re-add the assert + exception.

Still not quite sure how to handle the cases where `IdentHash`es are function parameters, since passing `nullptr` to those functions also calls this conversion constructor, even when they are const references.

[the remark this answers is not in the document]
anonimal: >I was mistakenly following Google's style guide

Can you point out where?

src/app/win32_service.cc
```diff
@@ -151,8 +152,11 @@ void I2PService::Start(
 
 void I2PService::OnStart(
     DWORD argc,
-    PSTR *argv) {
-  // TODO(unassigned): unused args
+    PSTR * argv) {
+  std::vector<std::string> args{};
+  for (DWORD i=0; i < argc; ++i)
+    args.push_back(argv[i]);
+  Daemon.Configure(args);
```
anonimal: I can't confirm this on windows run-time because of https://github.com/monero-project/meta/issues/238 but this looks logically sound.

coneiric:
You're welcome @anonimal

PR #882, opened 2018-05-30
Identity: refactor routing key creation

coneiric:
---

---

Resolves TODO replacing `sprintf` with calls to standard library functions, and fixes gcc8 format truncation warning.

### DIFF
--- a/src/core/router/identity.cc
+++ b/src/core/router/identity.cc
@@ -33,9 +33,7 @@
 #include "core/router/identity.h"
 
 #include <boost/endian/conversion.hpp>
-
-#include <stdio.h>
-#include <time.h>
+#include <boost/date_time/gregorian/gregorian_types.hpp>
 
 #include "core/crypto/hash.h"
 #include "core/crypto/rand.h"
@@ -727,42 +725,43 @@ Keys CreateRandomKeys() {
   return keys;
 }
 
-IdentHash CreateRoutingKey(
-    const IdentHash& ident) {
-  std::uint8_t buf[41];  // ident + yyyymmdd
-  memcpy(buf, (const std::uint8_t *)ident, 32);
-  time_t t = time(nullptr);
-  struct tm tm;
-  // TODO(unassigned): never use sprintf, use snprintf instead.
-#ifdef _WIN32
-  gmtime_s(&tm, &t);
-  sprintf_s(
-      reinterpret_cast<char *>((buf + 32)),
-      9,
-      "%04i%02i%02i",
-      tm.tm_year + 1900,
-      tm.tm_mon + 1,
-      tm.tm_mday);
-#else
-  gmtime_r(&t, &tm);
-  sprintf(
-      reinterpret_cast<char *>((buf + 32)),
-      "%04i%02i%02i",
-      tm.tm_year + 1900,
-      tm.tm_mon + 1,
-      tm.tm_mday);
-#endif
+// TODO(unassigned): should not be a free function
+IdentHash CreateRoutingKey(const IdentHash& ident)
+{
   IdentHash key;
-  // TODO(anonimal): this try block should be larger or handled entirely by caller
-  try {
-    kovri::core::SHA256().CalculateDigest((std::uint8_t *)key, buf, 40);
-  } catch (...) {
-    core::Exception ex;
-    ex.Dispatch(__func__);
-    // TODO(anonimal): review if we need to safely break control, ensure exception handling by callers
-    throw;
-  }
+  try
+    {
+      if (ident.IsZero())
+        throw std::invalid_argument("Identity: zero-initialized ident");
+
+      // Get spec-defined formatted date
+      std::string const date(GetFormattedDate());
+
+      // Create buffer for key
+      std::vector<std::uint8_t> buf(
+          ident(), ident() + (ident.size() + date.size()));
+
+      // Append the ISO date
+      buf.insert(buf.end(), date.begin(), date.end());
+
+      // Hash the key
+      assert(key.size() <= buf.size());
+      kovri::core::SHA256().CalculateDigest(key(), buf.data(), buf.size());
+    }
+  catch (...)
+    {
+      kovri::core::Exception ex(__func__);
+      ex.Dispatch();
+      throw;
+    }
   return key;
+}
+
+// TODO(unassigned): should not be a free function
+std::string GetFormattedDate()
+{
+  boost::gregorian::date date(boost::gregorian::day_clock::universal_day());
+  return boost::gregorian::to_iso_string(date);
 }
 
 XORMetric operator^(
--- a/src/core/router/identity.h
+++ b/src/core/router/identity.h
@@ -292,8 +292,16 @@ struct XORMetric {
   }
 };
 
-IdentHash CreateRoutingKey(
-    const IdentHash& ident);
+// TODO(unassigned): should not be a free function
+/// @brief Hash identity with appended mod date according to spec
+/// @param ident Identity hash used as the basis for the routing key
+/// @throw std::invalid_argument if supplied a zero-initialized ident
+IdentHash CreateRoutingKey(const IdentHash& ident);
+
+// TODO(unassigned): should not be a free function
+/// @return Formatted UTC date used in routing key
+/// @notes Must return yyyyMMdd
+std::string GetFormattedDate();
 
 XORMetric operator^(
     const IdentHash& key1,
--- a/tests/unit_tests/core/router/identity.cc
+++ b/tests/unit_tests/core/router/identity.cc
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2015-2017, The Kovri I2P Router Project
+ * Copyright (c) 2015-2018, The Kovri I2P Router Project
  *
  * All rights reserved.
  *
@@ -31,9 +31,11 @@
 #define BOOST_TEST_DYN_LINK
 
 #include <boost/test/unit_test.hpp>
+#include <boost/date_time/gregorian/gregorian_types.hpp>
 
 #include <array>
 #include <memory>
+#include <regex>
 
 #include "core/crypto/signature.h"
 #include "core/router/identity.h"
@@ -83,6 +85,26 @@ BOOST_AUTO_TEST_CASE(ParseIdentityFailure)
     BOOST_CHECK_EQUAL(
         identity.FromBuffer(m_AliceIdentity.data(), m_AliceIdentity.size() - i),
         0);
+}
+
+BOOST_AUTO_TEST_CASE(ValidRoutingKey)
+{
+  core::IdentityEx ident;
+  BOOST_CHECK(ident.FromBuffer(m_AliceIdentity.data(), m_AliceIdentity.size()));
+  BOOST_CHECK_NO_THROW(core::CreateRoutingKey(ident.GetIdentHash()));
+}
+
+BOOST_AUTO_TEST_CASE(InvalidRoutingKey)
+{
+  kovri::core::IdentHash hash;
+  BOOST_CHECK_THROW(core::CreateRoutingKey(nullptr), std::invalid_argument);
+  BOOST_CHECK_THROW(core::CreateRoutingKey(hash), std::invalid_argument);
+}
+
+BOOST_AUTO_TEST_CASE(ValidDateFormat)
+{
+  std::regex regex("(20\\d{2})(\\d{2})(\\d{2})");  // Valid for only this century
+  BOOST_CHECK(std::regex_search(core::GetFormattedDate(), regex));
 }
 
 BOOST_AUTO_TEST_SUITE_END()


coneiric:
> lol, I never want to read that sentence again.

:smile: felt gross writing it, too.

> Let's not use any C functions for routing key generation / in this function. I'd like to see a smarter / safer implementation of this function.

I'm using boost input/output facets for stream formatting in another branch I'm working on. Would this count as smarter/safer? For this PR, I was just following the TODO, and trying to minimally change the code. I'm willing to implement the facets, if you think this is a reasonable approach. If not, I'm open to suggestions.

> What warnings? I can't reproduce.

Using gcc8 on your `fix-build` branch with `make all-options`:
```
In file included from /kovri/build/src/core/cotire/kovri-core_CXX_unity.cxx:8:
/kovri/src/core/router/identity.cc: In function 'kovri::core::IdentHash kovri::core::CreateRoutingKey(const IdentHash&)':
/kovri/src/core/router/identity.cc:750:7: warning: '%04i' directive writing between 4 and 11 bytes into a region of size 9 [-Wformat-overflow=]
       "%04i%02i%02i",
       ^~~~~~~~~~~~~~
/kovri/src/core/router/identity.cc:750:7: note: directive argument in the range [-2147481748, 2147483647]
/kovri/src/core/router/identity.cc:750:7: note: directive argument in the range [-2147483647, 2147483647]
In file included from /usr/include/stdio.h:862,
                 from /usr/include/c++/8.1.0/cstdio:42,
                 from /usr/include/c++/8.1.0/ext/string_conversions.h:43,
                 from /usr/include/c++/8.1.0/bits/basic_string.h:6361,
                 from /usr/include/c++/8.1.0/string:52,
                 from /kovri/src/core/../core/instance.h:34,
                 from /kovri/src/core/instance.cc:33,
                 from /kovri/build/src/core/cotire/kovri-core_CXX_unity.cxx:4:
/usr/include/bits/stdio2.h:33:34: note: '__builtin___sprintf_chk' output between 9 and 34 bytes into a destination of size 9
   return __builtin___sprintf_chk (__s, __USE_FORTIFY_LEVEL - 1,
          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
       __bos (__s), __fmt, __va_arg_pack ());
```

anonimal:
>Using gcc8 on your fix-build branch with make all-options:

Ah, I see now. But the format is for unsigned and the type is signed:

```
In file included from /home/anonimal/kovri/build/src/core/cotire/kovri-core_CXX_unity.cxx:8:
/home/anonimal/kovri/src/core/router/identity.cc:751:7: warning: format specifies type 'unsigned short' but the argument has type 'int' [-Wformat]
      tm.tm_year + 1900,
      ^~~~~~~~~~~~~~~~~
/home/anonimal/kovri/src/core/router/identity.cc:752:7: warning: format specifies type 'unsigned short' but the argument has type 'int' [-Wformat]
      tm.tm_mon + 1,
      ^~~~~~~~~~~~~
/home/anonimal/kovri/src/core/router/identity.cc:753:7: warning: format specifies type 'unsigned short' but the argument has type 'int' [-Wformat]
      tm.tm_mday);
```

So, gcc says one thing and clang says another - but both are correct :man_shrugging:. I don't suspect overflow given the parameters but obviously this should be resolved during the refactor.

>I'm using boost input/output facets for stream formatting in another branch I'm working on. Would this count as smarter/safer?

Probably not smarter since the standard library would work fine with this function.

>For this PR, I was just following the TODO, and trying to minimally change the code.

Ok, then please just patch up this PR to work with windows, add a TODO to the function, and do the usual referencing of PR number in gitlog. Thanks.

coneiric:
> So, gcc says one thing and clang says another - but both are correct man_shrugging.

Right, I don't know either. Side-stepping the issue by using `std::put_time`.

> Probably not smarter since the standard library would work fine with this function.

Agreed, I found a better way. Thanks for the hint.

> Ok, then please just patch up this PR to work with windows, add a TODO to the function

I didn't mean that I wouldn't fully patch the issue, only explaining my reasoning for using `snprintf`. I'm still using the `gmtime` and `struct tm` from C, mostly because of the note about thread safety for `std::gmtime` in the reference material. If that's not a concern, I can replace `gmtime` and `struct tm` with the standard library counterparts.

anonimal:
Whatever you end up doing, a unit-test will be extremely helpful.  Also please consider rebasing against master so you can eliminate unrelated build warnings.

coneiric:
> Whatever you end up doing, a unit-test will be extremely helpful. Also please consider rebasing against master so you can eliminate unrelated build warnings.

Agreed, will add a unit-test, and rebase against master.

src/core/router/identity.cc
```diff
@@ -730,27 +730,23 @@ Keys CreateRandomKeys() {
...
-  sprintf(
-      reinterpret_cast<char *>((buf + 32)),
-      "%04i%02i%02i",
-      tm.tm_year + 1900,
-      tm.tm_mon + 1,
-      tm.tm_mday);
+  tm.tm_year += 1900;
+  tm.tm_mon += 1;
+  ss << std::put_time(&tm, "%Y%m%d");
+  std::memcpy(reinterpret_cast<char *>((buf + 32)), ss.str().data(), 9);
```
rakhimov: I don't get 1. why this is not done with [std::strftime](http://en.cppreference.com/w/cpp/chrono/c/strftime)
2. why platform specific ``gmtime_r/s`` ?

What is wrong with:
```
  std::time_t current_time = std::time(nullptr);
  char time_string[9] = {};
  auto ret = std::strftime(time_string, sizeof(time_string),
                           "%Y%m%d", std::gmtime(&current_time));
```

This should work for both platforms.

rakhimov: I see that ``std::gmtime`` may not be thread-safe.
``std::gmtime_s`` is C11 standard alternative for POSIX ``gmtime_r``.

coneiric: > why this is not done with std::strftime

Good question :). I wasn't familiar with the function, and must have missed it when looking at the reference material.

> I see that std::gmtime may not be thread-safe.

Originally, I had left those alternatives in for this reason, and will add them back. Thanks for the review @rakhimov.

coneiric: After looking at the documentation, `std::put_time` makes more sense for use with streams. If arrays and `memcpy` are better, then I'll use your suggestion.

Ideally, I would like to get rid of the `gmtime_*` calls as well. Boost.DateTime seems the only viable thread-safe alternative, but looks like a lot of overhead for such a simple use-case. Maybe it's worth it...

src/core/router/identity.cc
```diff
@@ -730,27 +730,23 @@ Keys CreateRandomKeys() {
 IdentHash CreateRoutingKey(
     const IdentHash& ident) {
   std::uint8_t buf[41];  // ident + yyyymmdd
-  memcpy(buf, (const std::uint8_t *)ident, 32);
+  std::memcpy(buf, (const std::uint8_t *)ident, 32);
   time_t t = time(nullptr);
   struct tm tm;
-  // TODO(unassigned): never use sprintf, use snprintf instead.
+  std::stringstream ss;
+  ss.imbue(std::locale(std::locale::classic()));
```
anonimal: >ss.imbue(std::locale(std::locale::classic()));

Why do you want this? I understand conceptually why but this doesn't appear necessary (and the router works fine without it in this branch).

src/core/router/identity.cc
```diff
@@ -730,27 +730,23 @@ Keys CreateRandomKeys() {
 IdentHash CreateRoutingKey(
     const IdentHash& ident) {
   std::uint8_t buf[41];  // ident + yyyymmdd
-  memcpy(buf, (const std::uint8_t *)ident, 32);
+  std::memcpy(buf, (const std::uint8_t *)ident, 32);
```
anonimal: This cast is unnecessary because of `Tags` functor.

src/core/router/identity.cc
```diff
@@ -730,27 +730,23 @@ Keys CreateRandomKeys() {
...
-      9,
-      "%04i%02i%02i",
-      tm.tm_year + 1900,
-      tm.tm_mon + 1,
-      tm.tm_mday);
+  gmtime_s(&t, &tm);
+  tm.tm_year += 1900;
+  tm.tm_mon += 1;
+  ss << std::put_time(&tm, "%Y%m%d");
+  std::memcpy(reinterpret_cast<char *>((buf + 32)), ss.str().data(), 9);
```
anonimal: If you're going to keep the memcpy's then you should setup the empty buffer with `ByteStream` and refactor accordingly.

As for keeping a stringstream for time, that's TBD depending on what else you plan on doing.

src/core/router/identity.cc
anonimal: Please consider doing whatever it takes to present a clean, streamlined, simple, efficient implementation of this function. A simple cross-platform thread-safe solution is welcome.

coneiric: Will do, I'll try Boost.DateTime, and see if it comes out cleaner.

src/core/router/identity.cc
coneiric: > this doesn't appear necessary

I realize now that this is the default, but wanted to ensure UTC  with no localization leaks before.

src/core/router/identity.cc
coneiric: Ok, I'll remove it in the next refactor, if I keep the `memcpy`s.

src/core/router/identity.cc
coneiric: > If you're going to keep the memcpy's then you should setup the empty buffer with ByteStream and refactor accordingly.

Will do, I'll look at setting this up with `ByteStream`. Will try both refactors (memcpy and Boost.DateTime), and see which comes out cleaner.

Thanks for the review, and the pointers on where to go next :+1:

coneiric:
The latest set of changes includes Boost DateTime refactors, and refactors to `CreateRoutingKey` callers. I thought handling the errors internally, and returning a zeroed key made more sense.

Left off the reference to the PR, if throwing (or something else) is preferred.

src/core/router/identity.cc
```diff
@@ -729,39 +727,37 @@ Keys CreateRandomKeys() {
...
+          key(),
+          reinterpret_cast<const std::uint8_t*>(ss.str().data()),
+          ss.str().size());
+    }
+  catch (...)
+    {
+      kovri::core::Exception ex(__func__);
+      ex.Dispatch();
+      IdentHash fail_key;
+      return fail_key;
```
anonimal: >IdentHash fail_key;
>return fail_key;

We don't want to create an object simply to satisfy a return type. We should rather re-throw and deal with consequences in any callers.

tests/unit_tests/core/router/identity.cc
```diff
@@ -85,4 +85,20 @@ BOOST_AUTO_TEST_CASE(ParseIdentityFailure)
...
+  BOOST_CHECK(!core::CreateRoutingKey(ident.GetIdentHash()).IsZero());
+}
+
+BOOST_AUTO_TEST_CASE(CreateInvalidRoutingKey)
+{
+  kovri::core::IdentHash zeroed_ident;
+  BOOST_CHECK_THROW(core::CreateRoutingKey(nullptr), std::invalid_argument);
+  BOOST_CHECK(core::CreateRoutingKey(zeroed_ident).IsZero());
+}
+
```
anonimal: This test-case needs to ensure proper / spec-defined date formatting.

src/core/router/identity.cc
```diff
@@ -729,39 +727,37 @@ Keys CreateRandomKeys() {
...
-    core::Exception ex;
-    ex.Dispatch(__func__);
-    // TODO(anonimal): review if we need to safely break control, ensure exception handling by callers
-    throw;
-  }
+  if (ident.IsZero())
+    {
+      LOG(error) << __func__
+                 << ": cannot create routing key from zero-initialized ident";
+      return key;
```
anonimal: We don't want to return an object simply to satisfy a return type. `IsZero()` was a hack to work with the union and should be eliminated / not built upon. We can keep IsZero for now but we should throw and deal with consequences in any callers.

src/core/router/identity.cc
```diff
@@ -729,39 +727,37 @@ Keys CreateRandomKeys() {
...
+    }
+  std::stringstream ss;
+  ss.exceptions(std::ios::failbit | std::ios::badbit);
```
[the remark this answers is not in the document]
coneiric: I was under the impression from the Boost examples this was enabling the stringstream to rethrow it's exceptions. I'll remove it.

src/core/router/identity.cc
coneiric: Ok, that makes sense. I'll make the changes.

tests/unit_tests/core/router/identity.cc
coneiric: How can we inspect the date formatting without re-implementing the function, or is that what you mean? I'm confused...

src/core/router/identity.cc
coneiric: Alright, that makes sense. I'll try to think of alternatives too (for another PR).

tests/unit_tests/core/router/identity.cc
anonimal: >How can we inspect the date formatting without re-implementing the function, or is that what you mean?

That would be expected if you wish to adequately test the data before it is hashed. Perhaps split into multiple functions (formatting versus key creation / hashing)?

src/core/router/identity.h
```diff
@@ -292,9 +292,21 @@ struct XORMetric {
   }
 };
 
+/// @brief Hash identity with appended mod date according to spec
+/// @param ident Identity hash used as the basis for the routing key
+/// @throw std::invalid_argument if supplied a zero-initialized ident
+/// @throw std::ios_base::failure on date formatting failure
+/// @throw CryptoPP::Exception on hashing failure
```
anonimal: >/// @throw std::ios_base::failure on date formatting failure
>/// @throw CryptoPP::Exception on hashing failure

We don't need this. We shouldn't keep track of every throw point that isn't directly related to the function's return or interface. Doing so only clutters the code and documentation, and is a pain to maintain should an implementation change.

src/core/router/identity.h
```diff
@@ -292,9 +292,21 @@ struct XORMetric {
...
 IdentHash CreateRoutingKey(
     const IdentHash& ident);
 
+/// @brief Format routing key with mod date according to spec
+/// @param ident Identity hash used as the basis for the routing key
+/// @throw std::invalid_argument if supplied a zero-initialized ident
+/// @throw std::ios_base::failure on date formatting failure
```
anonimal: >/// @throw std::ios_base::failure on date formatting failure

We don't need this. We shouldn't keep track of every throw point that isn't directly related to the function's return or interface. Doing so only clutters the code and documentation, and is a pain to maintain should an implementation change.

src/core/router/identity.h
```diff
@@ -292,9 +292,21 @@ struct XORMetric {
...
+std::string FormatRoutingKey(
+    const IdentHash& ident);
```
anonimal: Is this really greater than 80 chars (it's 53, clang-format shouldn't do this)?

tests/unit_tests/core/router/identity.cc
```diff
@@ -85,4 +86,47 @@ BOOST_AUTO_TEST_CASE(ParseIdentityFailure)
...
+BOOST_AUTO_TEST_CASE(ValidFormatRoutingKey)
+{
+  core::IdentityEx ident;
+  BOOST_CHECK(
+      ident.FromBuffer(m_AliceIdentity.data(), m_AliceIdentity.size()));
+
+  BOOST_CHECK_NO_THROW(core::FormatRoutingKey(ident.GetIdentHash()));
+  const std::string formatted_key = core::FormatRoutingKey(ident.GetIdentHash());
+  const core::IdentHash formatted_ident(reinterpret_cast<const std::uint8_t*>(
+      formatted_key.substr(0, ident.GetIdentHash().size()).data()));
```
anonimal: > const std::string formatted_key = core::FormatRoutingKey(ident.GetIdentHash());
>const core::IdentHash formatted_ident(reinterpret_cast<const std::uint8_t*>(
>    formatted_key.substr(0, ident.GetIdentHash().size()).data()));

This block is logically related to the comment below, not the check above. For sake of readability, you should put this block *below* the line *// Ensure formatted routing key contains the original ident hash*.

src/core/router/identity.cc
```diff
@@ -729,42 +727,47 @@ Keys CreateRandomKeys() {
...
-  // TODO(anonimal): this try block should be larger or handled entirely by caller
-  try {
-    kovri::core::SHA256().CalculateDigest((std::uint8_t *)key, buf, 40);
-  } catch (...) {
-    core::Exception ex;
-    ex.Dispatch(__func__);
-    // TODO(anonimal): review if we need to safely break control, ensure exception handling by callers
-    throw;
-  }
+  try
```
anonimal: Extend this try block to:

```diff
diff --git a/src/core/router/identity.cc b/src/core/router/identity.cc
index ccd6c862..ed793bc2 100644
--- a/src/core/router/identity.cc
+++ b/src/core/router/identity.cc
@@ -725,16 +725,17 @@ Keys CreateRandomKeys() {
   return keys;
 }
 
-IdentHash CreateRoutingKey(
-    const IdentHash& ident) {
-  if (ident.IsZero())
-    {
-      throw std::invalid_argument(
-          "Identity: cannot create routing key from zero-initialized ident");
-    }
+IdentHash CreateRoutingKey(const IdentHash& ident)
+{
   IdentHash key;
   try
     {
+      if (ident.IsZero())
+        {
+          throw std::invalid_argument(
+              "Identity: cannot create routing key from zero-initialized "
+              "ident");
+        }
       const std::string formatted_key = FormatRoutingKey(ident);
       kovri::core::SHA256().CalculateDigest(
           key(),
```

tests/unit_tests/core/router/identity.cc
```diff
@@ -85,4 +86,47 @@ BOOST_AUTO_TEST_CASE(ParseIdentityFailure)
...
+}
+
+BOOST_AUTO_TEST_CASE(ValidFormatRoutingKey)
+{
+  core::IdentityEx ident;
+  BOOST_CHECK(
+      ident.FromBuffer(m_AliceIdentity.data(), m_AliceIdentity.size()));
+
+  BOOST_CHECK_NO_THROW(core::FormatRoutingKey(ident.GetIdentHash()));
+  const std::string formatted_key = core::FormatRoutingKey(ident.GetIdentHash());
```
anonimal: This case does not test for a valid format. You will have to supply an extra argument to the formatting function (a ptime object), and then process accordingly (ensuring that time given is properly formatted).

src/core/router/identity.h
coneiric: Wasn't sure about the level of detail to include. I agree with you though, I'll remove it.

src/core/router/identity.h
coneiric: Wasn't sure about the level of detail to include. I agree with you though, I'll remove it.

src/core/router/identity.h
coneiric: No, I'm not sure why the line broke here, I'll fix it.

tests/unit_tests/core/router/identity.cc
coneiric: Ok, I must have missed that when making changes to the tests. I'll fix it.

src/core/router/identity.cc
```diff
@@ -729,42 +727,47 @@ Keys CreateRandomKeys() {
...
+  catch (...)
+    {  // Error formatting or hashing the key
+      kovri::core::Exception ex(__func__);
+      ex.Dispatch();
+      throw;
+    }
   return key;
 }
 
+std::string FormatRoutingKey(const IdentHash& ident)
```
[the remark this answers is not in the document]
coneiric: > our key is more akin to a vector of bytes and not a string

I understand, and thought that using stringstream was the most straight-forward way to use the Boost.DateTime functions. Will try to use a vector of bytes instead on this next set of refactors. Thanks for the tips.

tests/unit_tests/core/router/identity.cc
coneiric: I used `boost::gregorian::from_undelimited_string()` below, which I thought would throw when supplied an invalid ISO date string. If you prefer an explicit `ptime` argument, I can try to implement that in the next set of refactors.

coneiric:
@anonimal I made the latest recommended changes, and replaced uses of stringstream. If these changes are unwanted, I can revert back.

Left off the commit references to this PR, just in case you do want me to revert.

src/core/router/identity.cc
coneiric: I extended the try-block, but Github is still picking it up as unchanged.

src/core/router/identity.cc
```diff
@@ -729,42 +727,51 @@ Keys CreateRandomKeys() {
...
+  catch (...)
+    {  // Error formatting or hashing the key
+      kovri::core::Exception ex(__func__);
+      ex.Dispatch();
+      throw;
+    }
   return key;
 }
 
+std::vector<std::uint8_t> FormatRoutingKey(const IdentHash& ident)
```
coneiric: I did not add the `boost::gregorian::date` object as a second parameter, since I thought passing it by non-const reference may lead to thread-unsafe usage. Instead, I extract the date and ident from the result in tests.

src/core/router/identity.cc
```diff
@@ -729,42 +727,51 @@ Keys CreateRandomKeys() {
...
+  const std::string date_string = boost::gregorian::to_iso_string(date);
+  std::vector<std::uint8_t> formatted_buf;
+  formatted_buf.reserve(ident.size() + date_string.size());
+  // Add ident to the formatted buffer
+  for (std::size_t i = 0; i < ident.size(); ++i)
+    formatted_buf.push_back(ident()[i]);
+  // Add date string to the formatted buffer
+  for (const auto& c : date_string)
+    formatted_buf.emplace_back(static_cast<std::uint8_t>(c));
+  return formatted_buf;
```
anonimal: >If these changes away from stringstream use are unwanted, I can revert back.

This is incredibly inefficient, worse than before. Please revert (and put some newlines in here so other people can more easily read the code).

src/core/router/identity.cc
anonimal: See revert comment.

src/core/router/identity.cc
coneiric: Didn't intend to make it hard to read. Will think about other future refactors, and revert to stringstream for now.

Issue #823, opened 2018-03-06
ByteStream: assertion fires on over-read when parsing SSU Data fragment


vi:
Tried running Kovri on amrv7 board using downloaded pre-built version (kovri-24a5bb8-Linux-armv7l-2018.03.05) and got this after running for some short time:

```
[2018.03.06 20:07:55.577276] [0xb32ff460] [error]   NTCPServer: [SO4D] HandleConnect: 'Connection timed out'
kovri: /home/buildbot/slave/kovri-static-ubuntu-arm7/build/src/core/util/byte_stream.cc:87: void kovri::core::ByteStream::Advance(std::size_t): Assertion `len && len <= m_Length' failed.
Aborted
```

Shall I try to reproduce it more thoroughly, with a backtrace?

anonimal:
Hi @vi.

See https://github.com/monero-project/kovri/pull/820#issuecomment-370740695. I posted that about 12 hours ago (before this issue (#823) was opened) but got caught up in other things before opening the new ticket.

I will post details soon.

anonimal:
As noted in https://github.com/monero-project/kovri/pull/820#issuecomment-370740695:

>As for the assertion, it's very odd and also difficult to reproduce. The issue looks unrelated to this PR and appears to have been a long-standing, preexisting issue that was exposed with #810. I will open a new ticket with details.

Indeed it is. Let's look at the trace:

```
#11 0x000000000105708b in kovri::core::SSUSession::ProcessNextMessage (this=0x7fffdc0f0c50,
    buf=0x7fffdc064520 "\211\016\064m1\300\060\n\223E\025o\366\363\341c\002\266\030\341byg\352R\027\275W\306\024s8`Z\236c\363\300\002\275*F\210\302\310\060\364\003\222\237\216\377\002(\356\217\317\002\365shz\00
2", len=69, sender_endpoint=...) at /home/anonimal/kovri/src/core/router/transports/ssu/session.cc:241
        is_session = true
        key = 0x7fffdc0f10d0 "\221 \316RxV\270IѲ\037\223\352\223\342\276\341aMX\236\037y\032Z5\267\205\340\313\017\270\361c\236Z"
        __PRETTY_FUNCTION__ = "void kovri::core::SSUSession::ProcessNextMessage(uint8_t*, std::size_t, const endpoint&)"
        __func__ = "ProcessNextMessage"
```

which passes `len` size of 69 bytes to the handler:

```
[...]

PR #896, opened 2018-06-08
InputByteStream: checks for null source buffer when reading

coneiric:
---

---

~~~Partial fix for #823.~~~

Checks that buffers passed to `kovri::core::InputByteStream::Read` are non-null.


coneiric:
You're right, I was mistaken. Will continue to look for the source of the mysterious over-read bug.

PR #897, opened 2018-06-11
I2PControlDataTraits: Fix uncaught exception.

tmoravec:
Fix for Coverity Issue 175261.

I2PControlDataTraits::GetTrait throws std::domain_error on unknown
method. It gets called from I2PControlDataTraits::GetMethodFromString
which is marked as noexcept, and instead returns Method::Unknown.

Signed-off-by: Tadeas Moravec <ted@cryptosphere-systems.com>

---

---

### DIFF
--- a/src/client/api/i2p_control/data.cc
+++ b/src/client/api/i2p_control/data.cc
@@ -107,7 +107,7 @@ const std::string I2PControlDataTraits::GetTrait(Method method) const
 }
 
 I2PControlDataTraits::Method I2PControlDataTraits::GetMethodFromString(
-    const std::string& value) const noexcept
+    const std::string& value) const
 {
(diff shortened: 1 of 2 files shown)


tmoravec:
Oddly enough, Coverity found this problem on
I2PControlDataTraits::GetMethodFromString. But there are more
occurrences of the same issues around, namely all subclasses of
AbstractMethod suffer from the same.

Fixing all occurrences in the way you suggested - by removing noexcept.

anonimal:
For future reference https://github.com/monero-project/kovri-docs/blob/master/i18n/en/developer_guide.md#sending-your-work

tmoravec:
Thanks @anonimal . Done. Will keep it in mind in the future.

Issue #511, opened 2016-12-16
Tunnel/I2NP/Garlic rewrite/refactor


anonimal:
**By submitting this issue, I confirm the following:**

- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

*Place an X inside the bracket to confirm*
- [X] I confirm.

---

The Elephant in the Room, possibly as bad as SSU was before work was done in #140 and elsewhere, and as mentioned in #355 (see also #366), the entire tunnel/I2NP implementation needs to be rewritten/refactored.

Note: this condition of code exists nearly everywhere in the codebase. I've refrained from opening a flood of issues for each area (maybe I should do that) though, fortunately, we've been chipping away at the problems over time.

Ticketing for housekeeping.

anonimal:
Also, something to fix: 1fa369f971ebfca4c8583b6af58e31f5d4785c94

anonimal:
NOTICE: THIS ISSUE HAS BEEN MOVED TO GitLab. Please continue the discussion there. See https://github.com/monero-project/kovri/issues/1013 for details.

PR #912, opened 2018-06-15
Router: Fix lock in managing tunnels.

tmoravec:
There are two problems with the position of m_TransitTunnelsMutex:

1. Value of `it` might change between the condition
   `it->second->GetCreationTime() + TUNNEL_EXPIRATION_TIMEOUT` and
   erasing `it` from m_TransitTunnels

   Hence we need a lock around the whole condition.

2. But the content of m_TransitTunnels might change during the cycle.
   If ManageTransitTunnels runs twice in parallel, a tunnel can get
   deleted in one loop, but be accessed in the other one.

   Hence we need a lock around the whole cycle.

Coverity #151155

Signed-off-by: Tadeas Moravec <moravec.tadeas@gmail.com>

---

---

### DIFF
--- a/src/core/router/tunnel/impl.cc
+++ b/src/core/router/tunnel/impl.cc
@@ -667,14 +667,14 @@ void Tunnels::ManageInboundTunnels() {
 }
 
 void Tunnels::ManageTransitTunnels() {
-  std::uint64_t ts = kovri::core::GetSecondsSinceEpoch();
+  std::unique_lock<std::mutex> l(m_TransitTunnelsMutex);
+
+  const std::uint64_t ts = kovri::core::GetSecondsSinceEpoch();
   for (auto it = m_TransitTunnels.begin(); it != m_TransitTunnels.end();) {
     if (ts > it->second->GetCreationTime() + TUNNEL_EXPIRATION_TIMEOUT) {
       auto tmp = it->second;
-      LOG(debug) << "Tunnels: transit tunnel " << tmp->GetTunnelID() << " expired"; {
-        std::unique_lock<std::mutex> l(m_TransitTunnelsMutex);
-        it = m_TransitTunnels.erase(it);
-      }
+      LOG(debug) << "Tunnels: transit tunnel " << tmp->GetTunnelID() << " expired";
+      it = m_TransitTunnels.erase(it);
       delete tmp;
     } else {
       it++;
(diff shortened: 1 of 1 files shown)


src/core/router/tunnel/impl.cc
```diff
@@ -668,13 +668,13 @@ void Tunnels::ManageInboundTunnels() {
...
   for (auto it = m_TransitTunnels.begin(); it != m_TransitTunnels.end();) {
     if (ts > it->second->GetCreationTime() + TUNNEL_EXPIRATION_TIMEOUT) {
       auto tmp = it->second;
-      LOG(debug) << "Tunnels: transit tunnel " << tmp->GetTunnelID() << " expired"; {
-        std::unique_lock<std::mutex> l(m_TransitTunnelsMutex);
-        it = m_TransitTunnels.erase(it);
-      }
+      LOG(debug) << "Tunnels: transit tunnel " << tmp->GetTunnelID() << " expired";
+      it = m_TransitTunnels.erase(it);
       delete tmp;
```
coneiric: Unrelated to your changes, but since we're here, could you also remove the manual delete here? IIUC, deleting after invalidating the reference with `m_TransitTunnels.erase(it)` leads to a delete-after-free bug.

If I'm correct, you should also remove `auto tmp = it->second;`, since it would be unnecessary.

anonimal: >IIUC, deleting after invalidating the reference with m_TransitTunnels.erase(it) leads to a delete-after-free bug

Yes, invalid delete. Valgrind will confirm.

>If I'm correct, you should also remove auto tmp = it->second;, since it would be unnecessary.

Yes, but replace the copy with `it->second` so we can keep the log.

anonimal: Referencing #511.

tmoravec:
> Do our comments make sense? Will you be willing to make a 2nd commit for the changes we discussed (first commit will be for moving the lock)?

Sure, will do that in the next few days. Dtto #911 .

tmoravec:
Removed the invalid delete.

tmoravec:
> @tmoravec Could you please move the lock to the beginning of the function so we can have an ever-so-slightly more accurate timestamp?
This means spending ever-so-slightly longer in the critical section under the lock. Personally, I'd trade more accurate timestamp for the performance, but you are the expert on Kovri here. Updated the code.

Is it OK to have a new commit for this? Changing one that has already been superseded by the one removing the invalid delete is somewhat painful.

anonimal:
>critical section

I wouldn't call this critical, but I understand what you mean.

>Is it OK to have a new commit for this? Changing one that has already been superseded by the one removing the invalid delete is somewhat painful.

`$ git rebase -S -i master  # <-- then fixup or squash the last two into a single commit and reword the first commit if needed (be sure to include the PR number)`. If you want to make the timestamp const, you can add that as well while you're here.

tmoravec:
Thanks, @anonimal . Rebased, and added the `const`.

Issue #337, opened 2016-09-08
AddressBook: discuss/implement multiple subscription files


anonimal:
**By submitting this issue, I confirm the following:**
- I have read and understood the [contributor guide](https://github.com/monero-project/kovri/blob/master/doc/CONTRIBUTING.md).
- I have checked that the issue I am reporting can be replicated or that the feature I am suggesting is not present.
- I have checked opened or recently closed [pull requests](https://github.com/monero-project/kovri/pulls) for existing solutions/implementations to my issue/suggestion.

_Place an X inside the bracket to confirm_
- [X] I confirm.

---

First, see #336. Then, consider [java i2p's naming spec](https://geti2p.net/en/docs/naming).

For this ticket, we should discuss if we're to have separate subscription files because we currently only use `hosts.txt`. Also, if we hand edit the file, it will be overridden upon next fetch from a _any_ publisher.

With a little design work, we can easily implement other files that won't be overridden. There's also the question of whether we want to have separate files for separate publishers.

Referencing #305 and #296

anonimal:
In today's meeting we discussed an interesting solution:

1. `hosts.txt` will continue to be a static file updated upstream and clobbered every 12 hours. Users *do not* edit this file
2. `uri.host()_filename.txt` will be a unique subscription from unique publisher; re-named accordingly with the hostname of publisher **and** filename replaced with the filename provided by publisher (in case publisher publishes various subscriptions)
3. `private_hosts.txt` will be private subscriptions: never clobbered, never shared

This doesn't address *where* we save inserted addresses that are saved from an address helper since they currently are simply inserted into `./addressbook` though we could simply throw them into `private_hosts.txt`.

Issues that arise from this design:

- We lack an effective de-duplication impl
  - multiple subscriptions containing the same address(s) will simply take up space
  - If I load only one subscription and it doesn't have an address that another subscription has, after using a jumpservice, why would I save said address to `private_hosts.txt` when it already exists in another subscription?

Temporary solution:

- Until a database (#385) is implemented, we can scan and load *all* subscriptions on startup (but shouldn't we then fetch *all* subscriptions before scanning?). Any visited addresses can then be appended to `private_hosts.txt`.

Really all of this seems messy so I think we should push hard for #385.

anonimal:
[...]

PR #914, opened 2018-06-16
AddressBook: create multiple subscription map

coneiric:
---

---

First steps toward implementing #337. 

Default, user, and private subscription types will be used for their respective subscription files (`hosts.txt`, `user_hosts.txt`, and `private_hosts.txt`).

All publisher subscription files will save their unique addresses into the `user` subscription type, and the publisher's file/stream will be discarded.

### DIFF
--- a/src/client/address_book/storage.h
+++ b/src/client/address_book/storage.h
@@ -1,5 +1,5 @@
 /**                                                                                           //
- * Copyright (c) 2013-2017, The Kovri I2P Router Project                                      //
+ * Copyright (c) 2013-2018, The Kovri I2P Router Project                                      //
  *                                                                                            //
  * All rights reserved.                                                                       //
  *                                                                                            //
@@ -54,6 +54,23 @@ namespace client {
 /// @class AddressBookDefaults
 /// @brief Default string constants used throughout address book
 struct AddressBookDefaults {
+  /// @enum Subscription
+  /// @brief Subscription type for where to load/save subscription addresses
+  /// @notes Scoped to prevent namespace pollution (otherwise, purely stylistic)
+  enum struct SubscriptionType
+  {
+    Default,
+    User,
+    Private,
+  };
+
+  /// @alias AddressMap
+  /// @brief Maps human-readable hostname to an identity hash and subscription type
+  /// @details Intended for user-convenience, readability, mapping to/from database entries, and potential subscription feed support
+  /// @notes For subscription feed details, see I2P proposal 112
+  using AddressMap =
+      std::map<std::string, std::pair<core::IdentHash, SubscriptionType> >;
+
   /// @enum AddressBookSize
   enum AddressBookSize : std::uint16_t {
     /// @brief Line in subscription file
@@ -138,14 +155,12 @@ class AddressBookStorage : public AddressBookDefaults {
   /// @brief Loads subscriptions from file into memory
   /// @return Number of subscriptions loaded
   /// @param addresses Reference to map of human-readable addresses to hashes
-  std::size_t Load(
-      std::map<std::string, kovri::core::IdentHash>& addresses);
+  std::size_t Load(AddressMap& addresses);
 
   /// @brief Saves subscriptions to file in CSV format to verify addresses loaded
   /// @return Number of addresses saved
   /// @param addresses Const reference to map of human-readable address to b32 hashes of address
-  std::size_t Save(
-      const std::map<std::string, kovri::core::IdentHash>& addresses);
+  std::size_t Save(const AddressMap& addresses);
 
  private:
   /// @return Address book path with appended addresses location
--- a/tests/unit_tests/client/address_book/storage.cc
+++ b/tests/unit_tests/client/address_book/storage.cc
@@ -1,0 +1,188 @@
+/**                                                                                           //
+ * Copyright (c) 2015-2018, The Kovri I2P Router Project                                      //
+ *                                                                                            //
+ * All rights reserved.                                                                       //
+ *                                                                                            //
+ * Redistribution and use in source and binary forms, with or without modification, are       //
+ * permitted provided that the following conditions are met:                                  //
+ *                                                                                            //
+ * 1. Redistributions of source code must retain the above copyright notice, this list of     //
+ *    conditions and the following disclaimer.                                                //
+ *                                                                                            //
+ * 2. Redistributions in binary form must reproduce the above copyright notice, this list     //
+ *    of conditions and the following disclaimer in the documentation and/or other            //
+ *    materials provided with the distribution.                                               //
+ *                                                                                            //
+ * 3. Neither the name of the copyright holder nor the names of its contributors may be       //
+ *    used to endorse or promote products derived from this software without specific         //
+ *    prior written permission.                                                               //
+ *                                                                                            //
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY        //
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF    //
+ * MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL     //
+ * THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,       //
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,               //
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS    //
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,          //
+ * STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF    //
+ * THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.               //
+ */
+
+#define BOOST_TEST_DYN_LINK
+
+#include <boost/test/unit_test.hpp>
+
+#include <iostream>
+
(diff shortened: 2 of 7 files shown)


src/client/address_book/storage.h
```diff
@@ -54,6 +54,22 @@ namespace client {
...
+  {
+    Default,
+    User,
+    Private,
+  };
+
+  /// @alias AddressMap
+  /// @brief Maps human-readable hostname to an identity hash and subscription type
+  /// @notes Alias makes refactoring and use simpler
+  using AddressMap =
```
anonimal: https://google.github.io/styleguide/cppguide.html#Aliases

You'll need to decide how you intend to use this code in the future.

src/client/address_book/storage.h
```diff
@@ -54,6 +54,22 @@ namespace client {
...
+      std::map<std::string, std::pair<kovri::core::IdentHash, Subscription> >;
```
anonimal: I'm not a fan of pairing the hash with the subscription type but it does make sense that a hash will be unique to a subscription. You can replace the existing doxygen notes (or add a details entry) with something more descriptive.

src/client/address_book/storage.h
```diff
@@ -54,6 +54,22 @@ namespace client {
 /// @class AddressBookDefaults
 /// @brief Default string constants used throughout address book
 struct AddressBookDefaults {
+  /// @enum Subscription
+  /// @brief Subscription type for where to load/save subscription addresses
+  /// @notes Scoped to prevent namespace pollution (otherwise, purely stylistic)
+  enum struct Subscription
```
anonimal: I'd prefer `SubscriptionType` or something that makes this more obvious of not being an object.

tests/unit_tests/client/address_book/storage.cc
```diff
@@ -0,0 +1,196 @@
...
+
+#include "client/address_book/storage.h"
+
+#include "core/router/context.h"
+
+struct AddressBookStorageFixture
+{
+  /// @alias Subscription
+  /// @brief Subscription alias for readability and convenience
+  using Subscription = kovri::client::AddressBook::Subscription;
```
anonimal: https://github.com/monero-project/kovri/pull/914/files#r197602895

tests/unit_tests/client/address_book/storage.cc
```diff
@@ -0,0 +1,196 @@
...
+           ++dir_itr)
+        {
+          auto temp_file = dir_itr->path().filename();
+
+          // Only remove files created by AddressBookStorage
+          if (!boost::starts_with(temp_file.string(), test_prefix))
+            boost::filesystem::remove(temp_file);
+        }
+    }
+  catch (boost::filesystem::filesystem_error& er)
```
anonimal: We have an exception dispatcher. If you want to add filesystem specifics, go ahead.

tests/unit_tests/client/address_book/storage.cc
```diff
@@ -0,0 +1,196 @@
...
+    {
+      std::cerr << __func__ << ": " << er.what();
```
anonimal: LOG

tests/unit_tests/client/address_book/storage.cc
```diff
@@ -0,0 +1,196 @@
...
+  std::unique_ptr<kovri::client::AddressBookStorage> GetStorageInstance()
+  {
+    return std::make_unique<kovri::client::AddressBookStorage>();
+  }
+
+
+  std::string const test_prefix{"addr_test_"};  // used for temporary test files
+  kovri::client::AddressBook::AddressMap m_Addresses;
+  std::unique_ptr<kovri::client::AddressBookStorage> m_Storage;
+  boost::filesystem::path temp_path = boost::filesystem::temp_directory_path();
```
anonimal: A security concern but these are for tests so that shouldn't be an issue.

tests/unit_tests/client/address_book/storage.cc
```diff
@@ -0,0 +1,196 @@
...
+      "monero.i2p=3VzGaQQXwzN1iAwaPI17RK~gUqKqMH6fI2dkkGBwdayAPAdiZMyk1KGoTq~q1~HBraPZnz9mZJlzf6WVGCkUmUV3SBjBEbrdL9ud0fArq3P1~Ui9ViR9B7m5EG8smAnFvKZdqS-cnmHploUfIOefoQe0ecM7YYHErZsn3kL-WtvlfoDiSth-edIBpWxeHfmXSKoHSGSJ2snl5p9hxh30KmKj9AB0d4En-jcD83Ep3jsSvtPoQl7tSsh575~q0JJLsqGqm2sR9w4nZr7O58cg-21A2tlZeldM287uoTMb9eHWnYuozUGzzWOXvqg0UxPQSTfwh7YEhx0aRTXT2OFpr84XPoH2M6xIXfEMkFtJEJ-XlM-ILUZkg3kuBEFN7n4mBK~8L0Ht1QCq8L3~y7YnN61sxC0E9ZdyEOoC~nFJxndri9s9NzgZPo5eo6DsZXweOrTAIVQgKFUozL7WXKMlgqBZ5Nl3ijD6MGCIy0fWYHGLJ4jDBY7wrcfynVXFGm4EBQAEAAcAAA==",
+  }};
+};
+
+BOOST_FIXTURE_TEST_SUITE(AddressBookStorageTests, AddressBookStorageFixture)
+
+BOOST_AUTO_TEST_CASE(ValidSaveSubscription)
+{
+  std::size_t saved_addresses = 0;
```
anonimal: Ironically, since you love to use brace initialization in the oddest places, you didn't use it here :smile:

tests/unit_tests/client/address_book/storage.cc
```diff
@@ -0,0 +1,196 @@
...
+  // Ensure addresses written to storage
+  BOOST_CHECK_NO_THROW(saved_addresses = m_Storage->Save(m_Addresses));
+
+  // Ensure addresses are saved
+  BOOST_CHECK(saved_addresses && saved_addresses == m_Hosts.size());
+}
+
+BOOST_AUTO_TEST_CASE(InvalidSaveSubscription)
+{
+  std::size_t saved_addresses = 0;
```
anonimal: Rather than set this up as a temporary for every case, consider using within the fixture.

src/client/address_book/storage.h
coneiric: Ok, I'll add a `details` doxygen comment with intended use, and get rid of the vague `notes` comment.

src/client/address_book/storage.h
coneiric: My intent is to allow for the address map to reflect a database entry, and more easily support [subscription feeds](https://geti2p.net/spec/proposals/112-addressbook-subscription-feed-commands) (if we decide to implement). Will add that to the `details` doxygen comment.

src/client/address_book/storage.h
coneiric: Agreed, I'll update the name.

tests/unit_tests/client/address_book/storage.cc
coneiric: Right, will adding logging instead.

tests/unit_tests/client/address_book/storage.cc
coneiric: Security concern because of leaking potentially private data, or is there a filesystem integrity issue?

tests/unit_tests/client/address_book/storage.cc
coneiric: :smile: Still learning those C++ idioms, will brace initialize here.

tests/unit_tests/client/address_book/storage.cc
coneiric: Will do.

tests/unit_tests/client/address_book/storage.cc
coneiric: Right, I'll be more explicit about the intent as a user convenience.

tests/unit_tests/client/address_book/storage.cc
```diff
@@ -0,0 +1,192 @@
...
+      {
+        boost::filesystem::directory_iterator end_itr;
+        for (boost::filesystem::directory_iterator dir_itr(
+                 core::GetPath(core::Path::AddressBook));
+             dir_itr != end_itr;
+             ++dir_itr)
+          {
+            auto const temp_file = dir_itr->path().filename();
+
+            // Only remove files created by AddressBookStorage
```
anonimal: Remove everything. We shouldn't have lingering files after running a test.

tests/unit_tests/client/address_book/storage.cc
```diff
@@ -0,0 +1,192 @@
...
+
+    core::EnsurePath(core::GetPath(core::Path::AddressBook));
+
+    // Setup storage
+    m_Storage = GetStorageInstance();
+  }
+
+  ~AddressBookStorageFixture()
+  {
+    RemoveFiles();
```
anonimal:
```diff
diff --git a/tests/unit_tests/client/address_book/storage.cc b/tests/unit_tests/client/address_book/storage.cc
index b93b8d68..ee27557a 100644
--- a/tests/unit_tests/client/address_book/storage.cc
+++ b/tests/unit_tests/client/address_book/storage.cc
@@ -65,7 +65,7 @@ struct AddressBookStorageFixture
 
   ~AddressBookStorageFixture()
   {
-    RemoveFiles();
+    BOOST_CHECK_NO_THROW(RemoveFiles());
   }
 
   /// @brief Remove files created by AddressBookStorage
@@ -90,6 +90,7 @@ struct AddressBookStorageFixture
       {
         core::Exception ex(__func__);
         ex.Dispatch();
+        throw;
       }
   }
 
```

because we want to know when something goes wrong.

tests/unit_tests/client/address_book/storage.cc
coneiric: Ok, this was left over from creating random files with `boost::unique_path`. I'll remove all files.

tests/unit_tests/client/address_book/storage.cc
coneiric: Good point, I'll add the exception handling.

tests/unit_tests/client/address_book/storage.cc
anonimal: So, when I woke up and PRs were running through my head as I laid in bed, I realized that what I meant to say was *when the unit-test was finished*. Because you `RemoveFiles()` in the fixture dtor, it will remove X for every test-case. I don't want to get in your way if you want to play with files throughout the test, what I meant was that, when the test is finished, remove everything :grin:

PR #925, opened 2018-06-27
Testnet: refactor testnet script and Dockerfiles

coneiric:
---

---

Number of small refactors to make creating and using the testnet easier:

* Fixes issues with user permissions between Docker host and containers
* Enables dynamic builds for testnets based on the Arch Dockerfile
* Fixes Python syntax bug if Docker host is running Python3
* Enables monitoring by default
* Detects presence of monitoring containers

### DIFF
--- a/contrib/testnet/Dockerfiles/Dockerfile.arch
+++ b/contrib/testnet/Dockerfiles/Dockerfile.arch
@@ -28,12 +28,16 @@
 
 FROM greyltc/archlinux:latest
 
+ARG userid
+
+ENV userid=$userid
+
 RUN pacman -Syu --noconfirm && pacman -S --noconfirm \
       binutils boost cmake gcc clang gdb make \
       gnu-netcat iftop iproute2 openssh tcpdump \
       htop iotop \
       && \
-      useradd -m kovri -s /usr/bin/bash && \
+      useradd -m kovri -u $userid -s /usr/bin/bash && \
       chown -R kovri:kovri /home/kovri
 
 USER kovri
--- a/contrib/testnet/testnet.sh
+++ b/contrib/testnet/testnet.sh
@@ -76,6 +76,7 @@ web_host_octet=".2"
 
 pid=$(id -u)
 gid="docker" # Assumes user is in docker group
+docker_uid=$pid
 
 # TODO(unassigned): better sequencing impl
 #Note: sequence limit [2:254]
@@ -247,7 +248,7 @@ set_images()
     read_input "Change Dockerfile?: [KOVRI_DOCKERFILE=${KOVRI_DOCKERFILE}]" KOVRI_DOCKERFILE
   fi
   local _kovri_dockerfile_path="${KOVRI_REPO}/${docker_dir}/Dockerfiles/${KOVRI_DOCKERFILE}"
-  read_bool_input "Build Kovri Docker image? [$KOVRI_IMAGE]" KOVRI_BUILD_IMAGE "docker build -t $KOVRI_IMAGE -f $_kovri_dockerfile_path $KOVRI_REPO"
+  read_bool_input "Build Kovri Docker image? [$KOVRI_IMAGE]" KOVRI_BUILD_IMAGE "docker build --build-arg userid=$docker_uid -t $KOVRI_IMAGE -f $_kovri_dockerfile_path $KOVRI_REPO"
 
   # Select Kovri Webserver Dockerfile
   local _default_web_dockerfile="Dockerfile.apache"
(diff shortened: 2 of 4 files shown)


contrib/testnet/Dockerfiles/Dockerfile.arch
```diff
@@ -28,12 +28,16 @@
 
 FROM greyltc/archlinux:latest
 
+ARG userid
+
+ENV userid=$userid
```
anonimal:
```
Build Web Docker image? [httpd:2.4] [Y/n] y                                                                                                                                                                       
Sending build context to Docker daemon    741MB                                                                                                                                                     
Step 1/1 : FROM httpd:2.4                                                                                                                                                                                         
 ---> fb2f3851a971                                                                                                                                                                                                
[Warning] One or more build-args [userid] were not consumed
```

I'm imagining this is expected because the apache dockerfile wasn't updated (and it doesn't need to be)?

anonimal:
I don't think this PR was intended to fix the grafana portion of the testnet, but just FYI: I'm greeted with this screen after logging in to 127.0.0.1:3030 (i.e., it's still not working).

grafana

anonimal:
Or am I assuming that the plugins were to be installed via docker?

contrib/testnet/Dockerfiles/Dockerfile.arch
coneiric: No, I missed this when refactoring. Will remove the build-arg from the web Docker image.

coneiric:
> It's only for cpp-netlib. crypto++ is not reliant on openssl. Can you fix the message in that commit?

Absolutely, the compiler error message made me think it was for crypto++. Will update the message.

> I don't think this PR was intended to fix the grafana portion of the testnet

No, it wasn't, just to make actually building the testnet easier. I can keep working on the Grafana in another PR.

> Or am I assuming that the plugins were to be installed via docker?

If their API supports installing plugins, we can do it during Docker setup. Otherwise, maybe we just include instructions for manually installing the plugins?

anonimal:
>Otherwise, maybe we just include instructions for manually installing the plugins?

As automated as possible.

MoroccanMalinois:
> (i.e., it's still not working)
> Or am I assuming that the plugins were to be installed via docker?

I thinks it's working (had same screen after login), it's not required to add more users or apps&plugins. If you click on "Home", you should see a list of dashboards that you can click on (hopefully :sweat_smile:  )

coneiric:
> If you click on "Home", you should see a list of dashboards that you can click on

Can confirm, it's working, you just have to click on that "Dashboards" icon in the top left to get a list of dashboards.

I added a commit to install the InfluxDB admin panel, which you can add to any dashboard. It lets you view and kill InfluxDB queries. If this isn't useful/wanted, I can drop the commit.

anonimal:
>Can confirm, it's working.

Ok, but the actual monitoring isn't working for netdb.

not_working

anonimal:
Does anyone want to try fixing this now or should we merge?

MoroccanMalinois:
i'll take a look tonight

coneiric:
I can look as well.

coneiric:
> Since this enables monitoring by default, it should add
"--i2pcontroladdress 0.0.0.0 --i2pcontrolport 7650" to "KOVRI_BIN_ARGS" and "KOVRI_FW_BIN_ARGS"

Will do, maybe this is what's causing `kovri-util control stats` to fail as well. Will also add your recommendations for the loop in `collect.sh`.

contrib/testnet/testnet.sh
```diff
@@ -271,7 +272,14 @@ set_bins()
     mount_repo_bins="-v ${KOVRI_REPO}/build/kovri:/usr/bin/kovri \
       -v ${KOVRI_REPO}/build/kovri-util:/usr/bin/kovri-util"
 
-    read_bool_input "Build repo binaries from within the container?" KOVRI_BUILD_REPO_BINS "Exec make release-static"
+    build_cmd="make clean"
```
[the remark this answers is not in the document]
coneiric: I'm not sure this is necessary. `make release` and `make release-static` include rules for making `util`, but I can add this in case those recipes change.

coneiric:
I added your recommended changes @MoroccanMalinois, and will see if that solves the problem with `$stats` not being populated in `collect.sh`. If not, I will keep looking for a solution.

Thank you @anonimal and @MoroccanMalinois for your help on this PR.

coneiric:
> Got this last work-around

And with this, it looks like everything is working.

I removed `util` from `make clean util` from testnet creation, since it caused binaries to be rebuilt with the final build command being `make clean util release` or `make clean util release-static`.

If you had another reason for including `util` in the build command, I'll add it back.

coneiric:
> Thanks for working on this you two. @coneiric can you please reference this PR number (and optionally/preferably give credit to @MoroccanMalinois for the applicable changes)?

Absolutely, will do both, updating now.

PR #917, opened 2018-06-24
RouterInfo: refactor signature creation and verification

coneiric:
---
**By submitting this pull-request, I confirm the following:**

- I have read and understood the developer guide in [kovri-docs](https://github.com/monero-project/kovri-docs).
- I have checked that another pull-request for this purpose does not exist.
- I have considered and confirmed that this submission will be valuable to others.
- I accept that this submission may not be used and that this pull-request may be closed by the will of the maintainer.
- I give this submission freely under the BSD 3-clause license.
---

Moves `RouterInfo` signing into creation, before converting to a buffer. Adds signature verification and unit-test.

PR #926, opened 2018-06-28
Core: create buffer utility class + buffer unit-test + RI buffer refactor

anonimal:
---

---

This was getting in the way of me finishing my `bandcaps` branch. Otherwise, I'd have left the TODO.

RouterInfo is still very messy. Pulling it apart and a writing unit-test will be beneficial. Referencing #627.

### DIFF
--- a/src/core/router/info.cc
+++ b/src/core/router/info.cc
@@ -54,7 +54,7 @@ namespace kovri
 namespace core
 {
 
-RouterInfo::RouterInfo() : m_Exception(__func__), m_Buffer(nullptr)  // TODO(anonimal): buffer refactor
+RouterInfo::RouterInfo() : m_Exception(__func__)
 {
 }
 
@@ -103,8 +103,7 @@ RouterInfo::RouterInfo(
 
 RouterInfo::RouterInfo(const std::string& path)
     : m_Exception(__func__),
-      m_Path(path),
-      m_Buffer(std::make_unique<std::uint8_t[]>(Size::MaxBuffer))  // TODO(anonimal): buffer refactor
+      m_Path(path)
 {
   ReadFromFile();
   ReadFromBuffer(false);
@@ -112,14 +111,8 @@ RouterInfo::RouterInfo(const std::string& path)
 
 RouterInfo::RouterInfo(const std::uint8_t* buf, std::uint16_t len)
     : m_Exception(__func__),
-      m_Buffer(std::make_unique<std::uint8_t[]>(Size::MaxBuffer)),  // TODO(anonimal): buffer refactor
-      m_BufferLen(len)
+      m_Buffer(buf, len)
 {
-  if (!buf)
-    throw std::invalid_argument("RouterInfo: null buffer");
-  if (len < Size::MinBuffer || len > Size::MaxBuffer)
-    throw std::length_error("RouterInfo: invalid buffer length");
-  std::memcpy(m_Buffer.get(), buf, len);
   ReadFromBuffer(true);
   m_IsUpdated = true;
 }
@@ -138,18 +131,11 @@ void RouterInfo::ReadFromFile()
 
       // Get full length of stream
       stream.Seekg(0, std::ios::end);
-      m_BufferLen = stream.Tellg();
-      if (m_BufferLen < Size::MinBuffer || m_BufferLen > Size::MaxBuffer)
-        {
-          LOG(error) << "RouterInfo: buffer length = " << m_BufferLen;
-          throw std::runtime_error(m_Path + " is malformed");
-        }
+      m_Buffer(stream.Tellg());
 
       // Read in complete length of stream
       stream.Seekg(0, std::ios::beg);
-      if (!m_Buffer)
-        m_Buffer = std::make_unique<std::uint8_t[]>(Size::MaxBuffer);
-      stream.Read(m_Buffer.get(), m_BufferLen);
+      stream.Read(m_Buffer.data(), m_Buffer.size());
     }
   catch (...)
     {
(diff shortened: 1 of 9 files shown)


src/core/router/info.cc
```diff
@@ -715,23 +694,20 @@ void RouterInfo::CreateBuffer(const PrivateKeys& private_keys)
...
 
       // Signature
       // TODO(anonimal): signing should be done when creating RI, not after. Requires other refactoring.
       private_keys.Sign(
-          reinterpret_cast<std::uint8_t*>(m_Buffer.get()),
-          m_BufferLen,
-          reinterpret_cast<std::uint8_t*>(m_Buffer.get()) + m_BufferLen);
+          m_Buffer.data(),
+          m_Buffer.size(),
+          m_Buffer.data() + m_Buffer.size());
```
coneiric: These changes are definitely better that what's currently in place.

This is potentially dangerous if `router_info.Str().size() + private_keys.GetPublic().GetSignatureLen()` is greater than `Size::MaxBuffer`, since it would write past the end of the buffer, without throwing an error. With the changes in #917, this will be a non-issue here, but may come up in other places.

See my comment about a suggested test for proof of a `SIGABRT`.

tests/unit_tests/core/util/buffer.cc
```diff
@@ -0,0 +1,124 @@
...
+  BOOST_CHECK(buf.get() == data);
+
+  data.fill(1);
+
+  BOOST_CHECK_EQUAL(buf.size(), data.size());
+  BOOST_CHECK(buf.get() != data);
+
+  buf(data.data(), data.size());
+  BOOST_CHECK(buf.get() == data);
+}
```
coneiric: I'm not sure how to catch the stack smashing error with Boost.Test, but wanted to simulate the exceptional case of an overwrite during `RouterInfo` signing.

Here is the test case that produces the error:

```diff
diff --git a/tests/unit_tests/core/util/buffer.cc b/tests/unit_tests/core/util/buffer.cc
index fbed86ef..88fa4721 100644
--- a/tests/unit_tests/core/util/buffer.cc
+++ b/tests/unit_tests/core/util/buffer.cc
@@ -97,6 +97,15 @@ BOOST_AUTO_TEST_CASE(Data)
   BOOST_CHECK(buf.get() == data);
 }
 
+BOOST_AUTO_TEST_CASE(DataOverwrite)
+{
+  std::array<std::uint8_t, Max> data{{}};
+  std::array<std::uint8_t, 64> extra{{}};
+
+  BOOST_CHECK_NO_THROW(buf(data.data(), data.size()));
+  memcpy(buf.data() + buf.size(), extra.data(), extra.size());
+}
+
```
the output from the suite:
```
Running 166 test cases...
*** stack smashing detected ***: <unknown> terminated
unknown location(0): fatal error: in "Buffer/DataOverwrite": signal: SIGABRT (application abort requested)
/kovri/tests/unit_tests/core/util/buffer.cc(100): last checkpoint: "DataOverwrite" fixture dtor

*** 1 failure is detected in the test module "Master Test Suite"
```

The same error occurs using `PrivateKeys::Sign` like in the `RouterInfo` impl.

src/core/util/buffer.h
```diff
@@ -0,0 +1,154 @@
...
+#include <cstdint>
+#include <exception>
+#include <string>
+
+namespace kovri
+{
+namespace core
+{
+// TODO(anonimal): Boost.Buffer?
+// TODO(anonimal): SecBlock buffer
```
coneiric: FWIW, my vote is for SecBlock, then we could also use this class for `PrivateKeys`.

tests/unit_tests/core/util/buffer.cc
anonimal: I can't reproduce. What compiler are you using?

src/core/router/info.cc
coneiric: This bug pre-existed the buffer util changes, and buffer util makes the situation better. With the changes in the signature PR, on top of this PR, this bug will no longer exist.

tests/unit_tests/core/util/buffer.cc
anonimal: Note: none of the buildbot compilers (the ones that are online) can reproduce the issue. I can't reproduce with clang6. After testing with gcc8 though, I can reproduce.

>buf.data() + buf.size()

But what do you expect, you're telling the compiler to do a raw copy beyond the bounds of the buffer... You will get the same thing with this:

```cpp
std::array<std::uint8_t, Max> arr{{}};
std::array<std::uint8_t, Max> data{{}};
std::memcpy(arr.data() + arr.size(), data.data(), data.size());
```

Are you sure you weren't trying to do this?:

```diff
diff --git a/tests/unit_tests/core/util/buffer.cc b/tests/unit_tests/core/util/buffer.cc
index fbed86ef..911948f1 100644
--- a/tests/unit_tests/core/util/buffer.cc
+++ b/tests/unit_tests/core/util/buffer.cc
@@ -121,4 +121,11 @@ BOOST_AUTO_TEST_CASE(InvalidBuffer)
   BOOST_CHECK_THROW(core::Buffer<> buf(Max + 1), std::exception);
 }
 
+BOOST_AUTO_TEST_CASE(DataOverwrite)
+{
+  core::Buffer<0, 32> bad;
+  std::array<std::uint8_t, Max> data{{}};
+  BOOST_CHECK_THROW(bad(data.data(), data.size()), std::exception);
+}
+
 BOOST_AUTO_TEST_SUITE_END()
```

src/core/router/info.cc
anonimal: Yes, this is a dangerous and stupid way to do things, but this applies to any underlying pointer handling (and this is legacy code btw). I imagined you'd rebase #917 and that's good because RI needs further refactoring.

src/core/util/buffer.h
anonimal: Yes. The `SecBlock` family was a definitive yes (#784). As for the boost.buffer question, that's for this underlying non-SecBlock implementation.

tests/unit_tests/core/util/buffer.cc
coneiric: That case shows the expected failure much better. The one I posted was to show the danger of the previous impl, and why resolving the signature TODO is important.

Resolving the signature TODO will be much safer and straightforward with this buffer util.

src/core/router/info.cc
coneiric: > I imagined you'd rebase #917 and that's good because RI needs further refactoring.

This is exactly what I intend to do. Happy to be helping fix the legacy code.

src/core/util/buffer.h
coneiric: Ok, that makes sense.

PR #932, opened 2018-06-29
SSU: extend session lock for handling packets

### LINKED ISSUE #823: ByteStream: assertion fires on over-read when parsing SSU Data fragment

coneiric:
Extends sessions' lock to first access of internal sessions map. Prevents callers from different threads processing the same session packet multiple times.

Ran a number of live tests, and this patch appears to resolve the Bytestream overread issue. 

One router without this patch crashed from the overread error.

* recompiled with the patch
* restarted without getting another overread error
* used the same NetDb with both routers, so both should have the same view of the network
* client tunnels succeed with the patched router

Other live tests have been running with patched routers for several hours each, without crashing from the overread error.

Resolves #823

---

---

### DIFF
--- a/src/core/router/transports/ssu/server.cc
+++ b/src/core/router/transports/ssu/server.cc
@@ -247,6 +247,7 @@ void SSUServer::HandleReceivedPackets(
       if (!session || session->GetRemoteEndpoint() != pkt->from) {
         if (session)
           session->FlushData();
+        std::unique_lock<std::mutex> l(m_SessionsMutex);
         auto session_it = m_Sessions.find(pkt->from);
         if (session_it != m_Sessions.end())
           {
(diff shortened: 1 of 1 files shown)


coneiric:
> I won't NACK this change outright but I will NACK the references to resolving #823

I'll close this PR, the whole point of it was to resolve #823, and it doesn't do that. The bottleneck tradeoff you mentioned isn't worth it, since I was mistaken about this solving the issue.

Also, a router crashed from the overread assertion again today. So, I'll go through the trace you posted, and the logs from the crashed router. Will try to come up with testnet scenario to reproduce without relying on chance.

PR #935, opened 2018-06-30
Tests: NetDb: update parameters when adding RI

anonimal:
Was not rebased after the buffer refactor in #926

---
**By submitting this pull-request, I confirm the following:**

- I have read and understood the developer guide in [kovri-docs](https://github.com/monero-project/kovri-docs).
- I have checked that another pull-request for this purpose does not exist.
- I have considered and confirmed that this submission will be valuable to others.
- I accept that this submission may not be used and that this pull-request may be closed by the will of the maintainer.
- I give this submission freely under the BSD 3-clause license.
---

PR #938, opened 2018-07-02
HTTPProxy: improve HTTP error reporting

### LINKED ISSUE #936: Client: HTTPProxy returns wrong error message upon 503
---

---

`Please wait for the router to integrate` isn't accurate when a site is offline (the router is well integrated, the site is simply offline). This is generally confusing, especially for new users.

A trivial fix is to remove the message. A better fix is to improve error reporting.

coneiric:
Resolves #936.

---

---

### DIFF
--- a/src/client/proxy/http.cc
+++ b/src/client/proxy/http.cc
@@ -242,7 +242,7 @@ bool HTTPMessage::HandleData(const std::string& protocol_string) {
   std::vector<std::string> tokens_header_body;
   // get header info
   // initially set error response to bad_request
-  m_ErrorResponse = HTTPResponse(HTTPResponseCodes::status_t::bad_request);
+  m_ErrorResponse.set_response(HTTPResponse::status_t::bad_request);
   if (boost::algorithm::split_regex(
           header_body, protocol_string, boost::regex("\r\n\r\n")).size()
       != HEADERBODY_LEN)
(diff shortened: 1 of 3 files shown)


coneiric:
Build tests currently fail, but will pass after rebasing on a merged #935. Will add PR reference to commit logs with the rebase.

coneiric:
> Can you please split up the patch for #936 and the refactor work into two separate commits

Yes, I wasn't 100% on what you meant by "improve error reporting". Your comment makes that clearer, and I can re-add a more detailed explanation of possible reasons for an offline service.
> the error message is a valuable one so we would ultimately want to include the message into a full fledge response page

Should I include a warning about possible black-holing, where the user is cut-off from the service, or might that cause unnecessary fear in new users?

In my mind the reasons are:

* the service provider is offline
* the service provider is overloaded, and unable to respond
* the client is not in contact with any floodfills that have the server destination
* the client is being black-holed/sybil attacked (highly unlikely)

Recommended solution to the first three is simply try again later. The solution for the last is to generate new router keys, and reseed from a different source.

Does this sound reasonable?

anonimal:
>Does this sound reasonable?

No, not at all :smile: 

>The solution for the last is to generate new router keys, and reseed from a different source.

That isn't a solution. New router keys and reseeding from a different source won't prevent sybil (as in, these measures alone aren't appropriate to prevent sybil nor fix a sybil'd tunnel path or host).

>Recommended solution to the first three is simply try again later.

You can list all of these (in more plain english) into single response page. That would be good enough for now.

coneiric:
Lol, alright, glad I asked then.

I'll leave out mention of black-hole/sybil until I understand the solution better. Maybe discuss on IRC?

> You can list all of these (in more plain english) into single response page.

Will do, making the updates now.

anonimal:
>Maybe discuss on IRC?

Yes.

PR #944, opened 2018-07-05
Build: don't skip rpath for in-tree shared deps

anonimal:
---

---


ordtrogen:
I have tried building with these changes and it works fine, whereas before ./bin/kovri wouldn't start because it couldn't find a shared library (libcryptopp et al)

This PR seems to solve the problem

PR #950, opened 2018-07-06
SSU: Remove assert that's always true.

tmoravec:
Maybe there should be `&&` instead of `||`. But it doesn't really matter, because this situation is handled by the `default` branch of the following switch statement.

Coverity #183013

---

---


anonimal:
JFTR I think the TODO should stay but it's also apparently by looking at the design.

PR #955, opened 2018-07-09
SSU: use index for processing data packets

coneiric:
---

---

Instead of incrementing the packet buffer directly, increments an index used to access to packet buffer.


coneiric:
> You can resolve that TODO if you truly don't want to wait but there are more important issues to deal with right now.

I thought this would help resolve the bytestream overread issue. Since I was mistaken, I'll close this PR and focus on what's important for release.

PR #956, opened 2018-07-10
 Bytestream: data pointer refactor, add tellp + related SSU refactor

coneiric:
Use counter as index into the internal stream buffer, and use member functions for access.

References #823

---

---

### DIFF
--- a/tests/unit_tests/core/util/byte_stream.cc
+++ b/tests/unit_tests/core/util/byte_stream.cc
@@ -88,8 +88,8 @@ BOOST_AUTO_TEST_CASE(InputByteStream)
   BOOST_CHECK_THROW(input.SkipBytes(1), std::length_error);
 
   BOOST_CHECK_EQUAL_COLLECTIONS(
-      input.Data(),
-      input.Data() + input.Size(),
+      input.data(),
+      input.data() + input.size(),
       m_IPv4Array.begin(),
       m_IPv4Array.end());
 }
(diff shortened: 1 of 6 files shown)


anonimal:
The commit message must also change. This is not "safer access to stream buffer".

coneiric:
> Tellp() can stay but the rest is unnecessary and defeats the purpose of DataPtr and Advance()'ing the pointer.

That is the exact point of this PR, to not directly advance the pointer. Thought that the same packet buffer getting processed multiple times was contributing to #823.

Since the internal counter gets reset with each construction, I thought it would be safer to leave the pointer alone (it doesn't get reset, but remains advanced).

I'm still trying to trace down the root cause of the overread, since this PR doesn't entirely address the issue.

Still think the changes help the situation, but if you strongly disagree, I'll remove the changes.

coneiric:
Updated with the requested changes, and changes to unit-tests discussed on IRC. Live and unit-tests pass.

tests/unit_tests/core/util/byte_stream.cc
```diff
@@ -181,44 +181,44 @@ BOOST_AUTO_TEST_CASE(Bits64Test)
...
+      m_IPv4Array.begin(),
+      m_IPv4Array.end());
   // Reconstruct a new address and check with original
   boost::asio::ip::address_v4::bytes_type bytes;
   std::memcpy(bytes.data(), ip.data(), address.to_v4().to_bytes().size());
   BOOST_CHECK_EQUAL(boost::asio::ip::address_v4(bytes), address.to_v4());
 }
 
-BOOST_AUTO_TEST_CASE(AddressToByteVectorIPv6)
+BOOST_AUTO_TEST_CASE(address_to_bytesIPv6)
```
anonimal: Watchout for your global sed.

coneiric: Right, I missed this. Will update with the other fixups.

coneiric:
> Write and Read are not basic setters/getters. Do not change those.

I went a bit overboard here. Made the changes per our discussion in IRC, now the fixes should match the style guide.

anonimal:
>The style refactor can stay.

as well as the container iterators.

coneiric:
> On 2nd thought, NACK tellp(). We need const return types to protect the internal pointer. This is a functional refactor going in the wrong direction.

After discussing on IRC, and reading your comment, this makes sense to me too. I made the requested revisions, and kept everything else.

coneiric:
no problem :)

PR #977, opened 2018-07-18
[HOLD] SocksProxy: unit-tests, functional refactors for testing, and inlining refactors

coneiric:
---

---

### DIFF
--- a/src/client/proxy/socks.h
+++ b/src/client/proxy/socks.h
@@ -81,7 +81,7 @@ struct SOCKSDNSAddress {
 class SOCKSHandler
     : public kovri::client::I2PServiceHandler,
       public std::enable_shared_from_this<SOCKSHandler> {
- private:
+ protected:
    /// @enum State
    /// @brief Enumerators used for parsing data to
    ///  fetch the corresponding variable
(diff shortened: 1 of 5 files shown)


src/client/proxy/socks.h
```diff
@@ -334,9 +340,15 @@ class SOCKSHandler
       SOCKSServer* parent,
       std::shared_ptr<boost::asio::ip::tcp::socket> socket);
 
-  ~SOCKSHandler();
+  ~SOCKSHandler()
+  {
+    Terminate();
```
anonimal: No, move back. These are not small functions. This is noted in https://google.github.io/styleguide/cppguide.html#Inline_Functions.

src/client/proxy/socks.h
```diff
@@ -334,9 +340,15 @@ class SOCKSHandler
...
+  }
 
-  void Handle();
+  void Handle()
+  {
+    AsyncSocketRead();
```
anonimal: No, move back. These are not small functions. This is noted in https://google.github.io/styleguide/cppguide.html#Inline_Functions.

src/client/proxy/socks.h
coneiric: Alright, I misunderstood again. Didn't realize that the line count applied through the inner function call.

src/client/proxy/socks.h
coneiric: Alright, I misunderstood again. Didn't realize that the line count applied through the inner function call.

src/client/proxy/socks.h
anonimal: Always, because that is the code path.

src/client/proxy/socks.h
anonimal: Always, because that is the code path.

PR #982, opened 2018-07-23
Crypto: throw exception on invalid Ed25519 message signing and verification

coneiric:
---

---


anonimal:
Resolved in #983. We also already have a test-case for bad signature. Assertions aren't fired during tests because of https://github.com/monero-project/kovri/blob/master/CMakeLists.txt#L74.

PR #985, opened 2018-07-25
Kovri: fix log levels, log appropriate asio errors

anonimal:
This mostly affects transports. Keeps errors to *absolutely errors* and
not runtime issues that may be caused by an overloaded peer, for
example. An ultimate solution is to fine-tune error handling and fix
other known bugs.

Note: most of these warnings could, theoretically, be moved to debug
logging.

---

---


anonimal:
>Minor style refactor

It's not a style refactor: I intentionally didn't use 2 bytes because we only need 1 byte.

PR #998, opened 2018-07-30
SSUSession: get either IPv4/6 intro key from remote

### LINKED ISSUE #997: SSU: assert fires for null address of remote router
During a call to `SSUSession::GetIntroKey`, an assert fired when retrieving a remote router's address.
Here is a [backtrace](https://github.com/monero-project/kovri/files/2239556/ssu-session-assert.txt) of the assert firing.

Luckily, the assert was there to avoid dereferencing a null-pointer. I'll open a PR to throw exceptions for when/if asserts are removed in release builds.

---

---

anonimal:
---

---

Resolves #997 (should).

As for the need to throw for all null deference checks, I'll provide a better solution when I return so we don't have to litter the code with throws for every pointer.

### DIFF
--- a/src/core/router/transports/ssu/session.cc
+++ b/src/core/router/transports/ssu/session.cc
@@ -1655,7 +1655,7 @@ const std::uint8_t* SSUSession::GetIntroKey() const
     {
       LOG(debug) << "SSUSession: " << __func__ << ": using remote's key";
       auto* const address =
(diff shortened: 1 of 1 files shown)


anonimal:
>returns a nullptr, a null-dereference can still happen in release builds.

It's not possible because nullptr should never be returned because a transport address will be available (by that point).

>Looking forward to your solution for null pointers.

The solution is not for null pointers but for *how* we assert and throw.

PR #1009, opened 2018-09-05
Util: report router status

coneiric:
Uses kovri-util to report current router status.

Resolves implementation TODO.

---

---

### DIFF
--- a/src/client/api/i2p_control/session.cc
+++ b/src/client/api/i2p_control/session.cc
@@ -241,8 +241,9 @@ void I2PControlSession::HandleRouterInfo(
     {
       switch (pair.first)
         {
-          case RouterInfo::Status:  // TODO(unassigned): implement
-            response->SetParam(pair.first, std::string("???"));
+          case RouterInfo::Status:
+            response->SetParam(
+                pair.first, core::context.GetState(core::context.GetState()));
             break;
 
           case RouterInfo::Uptime:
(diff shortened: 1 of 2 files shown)


coneiric:
> This has already been resolved

Thanks for the link to the conversion function, I missed it when working on this patch. I'll use that method for enum-to-string conversion.

Still leaves getting the current status of the router. What is the difference between `RouterState` and `NetStatus`, are they interchangeable? If not, which should be reported by `kovri-util control status`?

I only saw the router's state being updated [here](https://github.com/monero-project/kovri/blob/master/src/core/router/transports/ssu/session.cc#L1062) (and the lines below). Cannot find where `NetStatus` is updated.

MoroccanMalinois:
> What is the difference between RouterState and NetStatus, are they interchangeable?

NetStatus is exclusively for [I2PControl](https://geti2p.net/en/docs/api/i2pcontrol) protocol. Kovri's implementation of I2PControl is almost complete.

But from the inside, Kovri implements only a subset of these possible state values, i.e.  `RouterState`.

I2PControl is very simple, I think `kovri-util control` would work with java i2p :stuck_out_tongue: 

> If not, which should be reported by kovri-util control status

My idea initially was to print a mini status report, like a summary of multiple values (version, uptime, netstatus ...). It doesn't need to be the result of a single parameter (like `control stats`). Individual items can be queried with option `-k`, like:

    kovri-util control -k i2p.router.net.status

Still leaves the question of what should be reported by `i2p.router.status`: is it just the string conversion of `i2p.router.net.status` or something else

